PR #158, opened 2023-06-08
Wait for linting tasks to start before waiting

charlieegan3:
This fails sometimes when the waiting goroutine completes before the linting tasks. In such cases an empty result is returned.


srenatus (approved):

anderseknert:
This is why I'm not touching concurrency 😄 Thanks for identifying the culprit @charlieegan3!

PR #159, opened 2023-06-12
Update to the opa-fmt url as it was returning 404s from the README

Parsifal-M:
Hey Team! 👋 

The project looks awesome and I can't wait to use it on some projects I have ideas for.

While reading the `README` and going through all the URLs I noticed the `opa-fmt` one was missing a `/` and would give you a `404`.

A bit of a silly `pr` I know in terms of scope but just thought I'd send a quick fix 😅

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ The following rules are currently available:
 | style    | [avoid-get-and-list-prefix](https://github.com/StyraInc/regal/blob/main/docs/rules/style/avoid-get-and-list-prefix.md)   | Avoid get_ and list_ prefix for rules and functions    |
 | style    | [prefer-snake-case](https://github.com/StyraInc/regal/blob/main/docs/rules/style/prefer-snake-case.md)                   | Prefer snake_case for names                            |
 | style    | [function-arg-return](https://github.com/StyraInc/regal/blob/main/docs/rules/style/function-arg-return.md)               | Function argument used for return value                |
-| style    | [opa-fmt](https:/github.com/StyraInc/regal/blob/main/docs/rules/style/opa-fmt.md)                                        | File should be formatted with `opa fmt`                |
+| style    | [opa-fmt](https://github.com/StyraInc/regal/blob/main/docs/rules/style/opa-fmt.md)                                       | File should be formatted with `opa fmt`                |
 | testing  | [file-missing-test-suffix](https://github.com/StyraInc/regal/blob/main/docs/rules/testing/file-missing-test-suffix.md)   | Files containing tests should have a _test.rego suffix |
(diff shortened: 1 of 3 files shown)


srenatus (approved):
Thanks

anderseknert:
Thanks! Good find. The table is generated from the Go rule metadata, so we ought to fix it there. I’m on the phone but can point out where when I’m at my laptop unless you find it first yourself:)

Parsifal-M:
> Thanks! Good find. The table is generated from the Go rule metadata, so we ought to fix it there. I’m on the phone but can point out where when I’m at my laptop unless you find it first yourself:)

Ohh I see! Will take another look 👍

anderseknert:
I must be missing something obvious 😅 

The entry is generated here:

https://github.com/StyraInc/regal/blob/562d430dae282d1651a7b430d54669ac1b1f10f5/cmd/table.go#L119

and the `Documentation` function simply joins the base URL with the path parameters:

https://github.com/StyraInc/regal/blob/562d430dae282d1651a7b430d54669ac1b1f10f5/pkg/rules/fmt.go#L88-L90

The base URL looks alright to me:

https://github.com/StyraInc/regal/blob/562d430dae282d1651a7b430d54669ac1b1f10f5/internal/docs/docs.go#LL3C14-L3C14

So where does that extra "github.com" come from 🤔

Parsifal-M:
Hey @anderseknert 👋 

I believe I might have found the culprit I think we need to do this:

```go
func (f *OpaFmtRule) Documentation() string {
	return docs.DocsBaseURL + "/" + category + "/" + title + ".md"
}
```
instead of using `path.Join` as it seems to play silly buggers with `/` in the `url` 🤔 

How are you generating the whole table? after I build it to test it and run `regal table --write-to-readme` I can provide the path like so `pkg/rules` but only seems to return for `opa-fmt` 😅 

Thanks! 👍

anderseknert:
Hmm, but the problem is not only with `/` but that github.com is repeated twice in the URL, no? But if it fixes the problem, I'm not complaining.

Here's how you can generate the table: `go run main.go table --write-to-readme bundle`

anderseknert:
We should add something about that to the development doc: https://github.com/StyraInc/regal/blob/main/docs/development.md

If you'll want to do that, that'd be great! If not I can do so later 🙂

Parsifal-M:
Hey @anderseknert 👋 

I think it's okay now! Happy to reword the little addition to the `development.md` section if it's not quite what you had in mind!

And sorry for the back and forth! Thought this was just a quick one 😅 

Thanks!

anderseknert (approved):
Thanks a lot!

anderseknert:
The back and forth was necessary, so no need to apologize! Most "quick fixes" I embark on tend to be anything but quick :)

PR #161, opened 2023-06-12
Some doc touchups

anderseknert:
- Add missing .md extension to some links
- Add talk from CNCF London
- Various small edits
- Use 4 spaces for indentation in embedded example


docs/assets/regal_cncf_london.png
srenatus, line 1: I guess a png of the first slide would be a little less noisy (visually). But it's OK, it screams "this is a recording of a meetup talk" at high volume.

srenatus (approved):

docs/assets/regal_cncf_london.png
anderseknert: Yeah, it does indeed. And I think that's OK :)

PR #170, opened 2023-06-21
refactor: one rule, one file

### LINKED ISSUE #144: Restructure rules directory — one rule, one file
Currently all rules in a category are lumped into a file named by the category.
Having worked with this for a while now I find it tedious to scroll around in large policy files where all the rules are named identically (and perhaps there are some linter rules to consider for this as well). I suggest we'd rather use one file per linter rule, and one file for its tests. Directory structure and package names should stay the same.

anderseknert:
Having all rules lumped into a single file only per category got to be a bit messy eventually. This change uses one file per rule, making it easier to both find and to work with rules in isolation — hopefully making it easier to contribute new ones too.

There is still a few things I would like to fix, like having unit tests actually be unit tests, and not as it currently stands that they test an input against *all* rules in the category... but this change is big enough as it is, so I'm deferring that to a future PR.

Fixes #144


srenatus (approved):
split and moved, ✔️

charlieegan3 (approved):

PR #171, opened 2023-06-21
Add idiomatic category

anderseknert:
And two new rules:
- custom-has-key-construct
- custom-in-construct

### DIFF
--- a/cmd/table.go
+++ b/cmd/table.go
@@ -62,7 +62,7 @@ func resolveDocsURL(url, category string) string {
 
 func createTable(args []string, params tableCommandParams) error {
 	result, err := loader.NewFileLoader().Filtered(args, func(abspath string, info fs.FileInfo, depth int) bool {
-		return strings.HasSuffix(abspath, "_test.rego")
+		return false
 	})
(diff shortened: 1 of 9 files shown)


cmd/table.go
```diff
@@ -62,7 +62,7 @@ func resolveDocsURL(url, category string) string {
 
 func createTable(args []string, params tableCommandParams) error {
 	result, err := loader.NewFileLoader().Filtered(args, func(abspath string, info fs.FileInfo, depth int) bool {
-		return strings.HasSuffix(abspath, "_test.rego")
+		return false
```
charlieegan3, line 65: Is this intended?

charlieegan3 (approved):

cmd/table.go
anderseknert: Yes. Since the refactoring we now have the files named by their rules, and the `todo-test` rule then by necessity lives in a file called "todo_test.rego", while its test will have to be "todo_test_test.rego". Since we're still filtering metadata here, there's no risk of having tests included in the table.

PR #174, opened 2023-06-26
Refactor rule metadata logic

anderseknert:
Some additional refactoring made possible now that each rule resides in its own file. Extend this further to have each rule in its own package, and move the metadata from the `report` rule in each package and to the package declaration. We may then use `rego.metadata.chain()` instead, which includes the path of the package containing both the category and the title, thus allowing us to skip having to redeclare that in annotations.

This has a number of benefits:

- Tests are now proper _unit_ tests, and no longer test the whole category.
- Checking if a rule is ignored can now be the responsibility of the main rule, and linter rules need no longer be concerned with that unless they have unique configuration options.
- Reduce the amount of clutter in each rule related to metadata, and with that the possibility of mistakes.


charlieegan3 (approved):
👍 Looks good to me. It's more magic, but I think it's still a good improvement that should speed up adding new rules. It also makes rules share less code, making wider updates easier to make too.

PR #177, opened 2023-06-26
test-outside-test-package: functions prefixed test_ aren't tests

### LINKED ISSUE #176: False positive for test-outside-test-package when function named test_*
Functions aren't evaluated as tests, so these should not count as such.

anderseknert:
Fixes #176

### DIFF
--- a/bundle/regal/ast.rego
+++ b/bundle/regal/ast.rego
@@ -20,6 +20,7 @@ package_name := concat(".", [path.value |
 
 tests := [rule |
 	some rule in input.rules
+	not rule.head.args
 	startswith(rule.head.name, "test_")
 ]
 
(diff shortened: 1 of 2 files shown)


bundle/regal/ast.rego
```diff
@@ -20,6 +20,7 @@ package_name := concat(".", [path.value |
 
 tests := [rule |
 	some rule in input.rules
+	not rule.head.args
```
charlieegan3, line 23: Does this warrant a comment to say that this is saying `if not function`?

charlieegan3 (approved):

bundle/regal/ast.rego
anderseknert: Yeah, I guess I've become a little blind to these things by now 😄 Perhaps even better to add a `is_function(rule)` helper to `ast.rego`? I can do so in a follow-up PR.

srenatus: 💭 Or even
```rego
is_function(rule) if not rule.head.args
```

charlieegan3: Yeah that works better!

PR #179, opened 2023-06-27
Fix unconditional-assignment with else clause

### LINKED ISSUE #178: False positive unconditional-assignment when `else` clause present
The following code is (arguably correctly) flagged as an unconditional assignment.

```rego
msg := x {
	x := input.foo
} else := input.bar
```

While that is true, skipping the assignment isn't possible, as this is currently [not valid](https://github.com/open-policy-agent/opa/issues/5777#issue-1632672461):

```rego
msg := input.foo else := input.bar
```

While one could instead do something like:
```rego
msg := input.foo {
	true
} else := input.bar
```

I don't think Regal should be flagging this as unconditional until the issue is fixed in OPA proper.

anderseknert:
To be revisited once the issue in OPA is resolved.

Fixes #178

### DIFF
--- a/bundle/regal/rules/style/unconditional_assignment.rego
+++ b/bundle/regal/rules/style/unconditional_assignment.rego
@@ -17,6 +17,11 @@ report contains violation if {
 	# a `print` call.. but let's keep it simple for now
 	count(rule.body) == 1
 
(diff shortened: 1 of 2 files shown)


charlieegan3 (approved):
Good catch!

PR #188, opened 2023-06-29
Comment and metadata rules

anderseknert:
- Add `no-whitespace-comment` rule in the `style` category, because comments should start with a whitespace!
- Add `invalid-metadata-attribute` rule in the `bugs` category. This one checks that all attributes of a metadata block are valid, i.e. not custom outside of the custom map.

### DIFF
--- a/bundle/regal/ast.rego
+++ b/bundle/regal/ast.rego
@@ -191,3 +191,38 @@ implicit_boolean_assignment(rule) if {
 all_functions := object.union(opa.builtins, function_decls(input.rules))
 
 all_function_names := object.keys(all_functions)
+
+comments_decoded := [decoded |
+	some comment in input.comments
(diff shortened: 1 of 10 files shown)


bundle/regal/ast.rego
```diff
@@ -191,3 +191,50 @@ implicit_boolean_assignment(rule) if {
...
+	"title",
+	"description",
+	"related_resources",
+	"authors",
+	"organizations",
+	"schemas",
+	"entrypoint",
+	"custom",
+}
+
```
srenatus, line 213: This could become a bit of a hassle if OPA introduces further attributes. But OK for now.

srenatus (approved):

bundle/regal/ast.rego
anderseknert: True. Not sure if there's a way to enumerate them from the Go API, but that would be preferable 🤔

PR #192, opened 2023-07-03
Add e2e test of all rules

### LINKED ISSUE #53: Add e2e tests
In addition to the unit tests, it would be good to add a few end-to-end tests testing the commands like `regal lint` with flags like `--timeout`, `--config-file`, etc.

anderseknert:
Which exposed flaws in a few rules:
- custom-in-construct would not work with multiple wildcards
- function-arg-return wrong input format for fetching config
- line-length wrong input format for fetching config

While we by no means are done with e2e tests, I think it's OK to say we now at least *have* e2e tests, and much of course thanks to @srenatus previous work on this.

Resolves #53

### DIFF
--- a/e2e/cli_test.go
+++ b/e2e/cli_test.go
@@ -6,12 +6,38 @@ import (
 	"bytes"
 	"encoding/json"
 	"io"
+	"log"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"gopkg.in/yaml.v3"
+
+	"github.com/styrainc/regal/pkg/config"
+	"github.com/styrainc/regal/pkg/report"
 )
+
+func readProvidedConfig(t *testing.T) config.Config {
+	t.Helper()
+
+	cwd := must(os.Getwd)
+
+	bs, err := os.ReadFile(filepath.Join(cwd, "..", "bundle", "regal", "config", "provided", "data.yaml"))
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	var cfg config.Config
+
+	err = yaml.Unmarshal(bs, &cfg)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	return cfg
+}
 
 func TestCLIUsage(t *testing.T) {
 	t.Parallel()
@@ -84,7 +110,7 @@ func TestLintEmptyDir(t *testing.T) {
 	}
 }
 
-func TestLintNonExistantDir(t *testing.T) {
+func TestLintNonExistentDir(t *testing.T) {
 	t.Parallel()
 
 	stdout := bytes.Buffer{}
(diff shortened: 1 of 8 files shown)


e2e/cli_test.go
```diff
@@ -6,13 +6,39 @@ import (
...
 
+func readProvidedConfig(t *testing.T) config.Config {
+	t.Helper()
+
+	cwd := must(os.Getwd)
+
+	bs, err := os.ReadFile(filepath.Join(cwd, "..", "bundle", "regal", "config", "provided", "data.yaml"))
+	if err != nil {
+		log.Fatal(err)
+	}
```
srenatus, line 30: I think I'd have gone with t.Fatalf, since it's already a helper function.

e2e/cli_test.go
```diff
@@ -151,3 +226,14 @@ func ExitStatus(err error) int {
 
 	panic("unreachable")
 }
+
+func must[R any](f func() (R, error)) R {
```
srenatus, line 230: Nice!

srenatus (approved):

e2e/testdata/violations/most_violations.rego
```diff
@@ -0,0 +1,110 @@
...
+# METADATA
+# invalid-metadata-attribute: true
+should := "fail"
+
+not_equals_in_loop {
+	"foo" != input.bar[_]
+}
+
+# rule-named-if
+# if { true }
```
charlieegan3, line 40: Should this be uncommented?

srenatus: https://github.com/StyraInc/regal/pull/192/files#diff-2fa353151f098139be7a7e83e4d4866fb7723b8d6833ce1b467cf69512be09f1R3-R5 isn't what it says here that it should be removed, as it needs its own file...? But I'm a bit confused myself 😅

anderseknert: Yes it should be removed as it won’t trigger in this file due to the keywords import. Probably just forgot.

charlieegan3: 👍 ahh I've removed it. I made another small fix in cce360e0cde331dadcb3d3f9ae13c0c6f27db472 after your comment.

e2e/cli_test.go
charlieegan3: See cce360e0cde331dadcb3d3f9ae13c0c6f27db472

e2e/testdata/violations/most_violations.rego
charlieegan3: Ahh, ok! I'll update.

charlieegan3: af0966c022d9b3a383804e7b748387a58bc0b7f3 look ok?

charlieegan3: Ooops, https://github.com/StyraInc/regal/pull/194

PR #197, opened 2023-07-11
Add Regal bundle to test cmd runner

### LINKED ISSUE #196: regal test reports "undefined function data.regal.result.fail"
# Description

I am following the guide [here](https://github.com/StyraInc/regal/blob/698c78ec2cd215921c9cfbe9ceb5099f8ecb47bc/docs/custom-rules.md) for writing custom Regal lint rules. However, when I try to write tests for the custom rules stored in `.regal/rules` as described in the guide, I get the following error messages when I run `regal test`

```
$ regal test .regal/rules
2 errors occurred:
.regal/rules/guide.rego:17: rego_type_error: undefined function data.regal.result.location
.regal/rules/guide.rego:17: rego_type_error: undefined function data.regal.result.fail
```

# Expected output

Based on what I understand from the guide referenced above,

> * `regal test` works like `opa test`, but aware of any custom Regal additions. Use this to test custom linter rules

I would expect this to work out of the box.

# Reproduce reported behaviour

1. Add the example custom rule from [here](https://github.com/StyraInc/regal/blob/698c78ec2cd215921c9cfbe9ceb5099f8ecb47bc/docs/custom-rules.md) to a file `.regal/rules/guide.rego`
2. Run `regal test .regal/rules`
3. Observe output.

# Environment
* Regal version:
  ```
  Version:    0.4.0
  Go Version: go1.20.5
  Platform:   darwin/arm64
  Commit:        unknown
  Timestamp:  2023-07-04T09:02:54Z
  ```

kristiansvalland:
Also add e2e test to validate that the bundle is added correctly.

Fixes #196

### DIFF
--- a/e2e/cli_test.go
+++ b/e2e/cli_test.go
@@ -202,6 +202,7 @@ func TestTestRegalBundledRules(t *testing.T) {
 	}
 
 	var res []tester.Result
+
 	err = json.Unmarshal(stdout.Bytes(), &res)
 	if err != nil {
 		t.Fatalf("expected JSON response, got %v", stdout.String())
--- a/e2e/cli_test.go
+++ b/e2e/cli_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"gopkg.in/yaml.v3"
 
+	"github.com/open-policy-agent/opa/tester"
 	"github.com/styrainc/regal/pkg/config"
 	"github.com/styrainc/regal/pkg/report"
 )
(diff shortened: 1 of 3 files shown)


srenatus (approved):
👏 What a nice first contribution. Thanks a lot!

e2e/cli_test.go
```diff
@@ -182,6 +183,31 @@ func TestLintAllViolations(t *testing.T) {
 	}
 }
 
+func TestTestRegalBundledRules(t *testing.T) {
+	t.Parallel()
+
+	stdout := bytes.Buffer{}
+	stderr := bytes.Buffer{}
```
srenatus, line 190: Not for this PR, just a random thought -- I wonder if something like this would be useful for testing:

```go
type panicWriter struct{}

func (*panicWriter) Write([]byte) (int, error) {
	panic("write unexpected")
}
```

It's like `io.Discard` just with a little more drama. https://play.golang.com/p/xCQlGZH2nUq

anderseknert:
This looks great! I’m on mobile but do I see it right that no checks are running for this PR? I wonder why that might be. I don’t see an option to “approve” of them either. @srenatus you know?

e2e/cli_test.go
srenatus: Yeah you've made a good point there. Perhaps something that takes a `*testing.T` and fails accordingly would strike a middle ground... maybe some day 🙃

srenatus:
@anderseknert The [build workflow's "on"](https://github.com/StyraInc/regal/blob/main/.github/workflows/build.yaml#L3-L7) is wonky, I think:
```yaml
on:
  push:
    branches:
      - '**'
  workflow_dispatch:
```

This only matches branches _on this repository_ and will thus not run for PRs from forks. OPA uses that
```yaml
on: [pull_request]
```

Also, we don't have protection rules defined, it seems, so the checks aren't required by github either. Hence the PR got green without any tests run.

Issue #187, opened 2023-06-27
Add rule specific ignore options to configuration


anderseknert:
Similar to #186, but rather than ignore files entirely, it should be possible to configure an ignore option on any rule.

One example might be that one may want to allow e.g. `print` calls in any package with a `_test` suffix, or file with a `_test.rego` suffix, but **not** anywhere else.

anderseknert:
This was fixed in #200. Closing.

PR #200, opened 2023-07-16
Add functionality for ignoring files at a global level and rule level.

### LINKED ISSUE #186: Add ignore option as CLI flag and to configuration
Some policy repositories encountered have the bad habit of including non-valid Rego files in their directories, often including something like template variables that are supposedly replaced in some build step. In order to work around this, we should allow ignoring files and directories matching either a path, a pattern or a filename.

It would be good if we supported this both via an `--ignore` flag in the `regal lint` command, as well as via configuration file.
I could imagine the latter provided a new top-level attribute called `ignore` or `ignore-options`, where one could provide a list of files/dirs to ignore.

kristiansvalland:
## Summary

Add ignore options for ignoring files for
- a specific test
- globally using either CLI flag or config `.ignore`

## Discussion

The specifics of how files are matched is something that should be discussed and considered carefully, as changing how patterns are understood will be hard after it has been established.

At the time of writing, I have tried to massage the OPA implementation of `glob.match()` (https://github.com/gobwas/glob AFAIK) into working the same way as `.gitignore` file patterns work, based on the [description](https://git-scm.com/docs/gitignore#_pattern_format). The motivation for doing this is the assumption that this is the patterns that most users will be used to. There are some differences in how patterns match directories and wildcards between the implementations.

I have tried to create some examples in [this Rego playground example](https://play.openpolicyagent.org/p/2CTwdisqwl) for how the pure `glob.match` would work, how the proposed matching works and what the differences and similarities between them are.

The cost of the current approach is that it will likely increase the maintenance of this repository. I think I have covered most of the obvious `.gitignore` patterns, but there will likely be some cases that I have missed and will constitute future bugs that needs to be fixed when they arise. The likely source of such bugs could be differences from how `.gitignore` interprets
> Other consecutive asterisks are considered regular asterisks and will match according to the previous rules.
The hypothesis is that these potential future bugs are more towards being edge-cases and will appear infrequent.

If we were to use pure `glob.match`/`gobwas/glob`, then we do not need to maintain the functionality of it -- which is great. On the other hand, this would likely require users to get acquainted with how the `glob.match` actually works.

In summary, I would really appreciate some feedback and thoughts on the above discussion and way forward from the maintainers! :) 

## To-do

I have yet to
- Write tests in go
- Finish the go function `excludeFile`. The body currently contains just a panic.

## Full list of changes

- bundle/: Add rules for ignoring files based on user config and CLI params.
- README.md: Add info on how to ignore files using the new functionality.
- .gitignore: Only ignore regal binary in root dir, not any folder or file named "regal"
- cmd/: Add "ignore" flag to CLI.
- pkg/: Provide ignore CLI flag and filter out files for Go lints.

Closes #186 and #187.


anderseknert:
Awesome! Thanks for looking into this. I'm travelling for a couple of days, but I hope to get some time to take a closer look this wek. Will get back to you as soon as I've done so!

README.md
```diff
@@ -187,6 +187,11 @@ more about it.
 
 **.regal/config.yaml**
 ```yaml
+# Files can be excluded from all lint 
+# rules according to glob-patterns
+ignore:
```
anderseknert, line 192: I wonder if we should do:
```yaml
ignore:
  files:
    - file1.rego
    - "*_tmp.rego"
```
I.e. add a `files` attribute under `ignore`, in case we want to add something similar for packages (or whatnot) later?

bundle/regal/config/exclusion.rego
```diff
@@ -0,0 +1,65 @@
...
+# exclude imitates Gits .gitignore pattern matching as best it can
+# Ref: https://git-scm.com/docs/gitignore#_pattern_format
+exclude(pattern, file) if {
+	patterns := pattern_compiler(pattern)
+	some p in patterns
+	glob.match(p, ["/"], file)
+} else := false
+
+# pattern_compiler transforms a glob pattern into a set of glob patterns to make the
+# groups behave as Gits .gitignore
```
anderseknert, line 34: This is really cool!

bundle/regal/config/exclusion_test.rego
```diff
@@ -0,0 +1,79 @@
...
+			exp != act
+			res1 := {"exp": exp, "act": act}
+		}
+		count(res) > 0
+	}
+
+	count(not_exp) == 0
+}
+
+rule_specific_config := {"ignore": ["p.rego"]}
```
anderseknert, line 51: Is this really rule specific? I don't see where it's attached to some specific rule.

README.md
kristiansvalland: Yes, I like that idea much better! Will implement :)

bundle/regal/config/exclusion_test.rego
anderseknert: I believe it's memoized given the same args, but honestly not sure ... we should probably add some docs on that in OPA 😅 @srenatus would know, I think.

What I meant by a rule specific ignore would be something like:

```yaml
rules:
  testing:
    print-or-trace-call:
      level: error
      ignore:
        files:
          - **/*_test.rego
```

anderseknert: Or perhaps I'm just too tired and this is already how things work 😅 All good if so! But a test case that made this more clear would be great.

kristiansvalland: As discussed on Slack, ignoring files is possible at the rule level, global config level and through the CLI flag `ignore-files`.

Also, the interface of the `exclusion.rego` file is made similar to that of the `config.for_rule`, so it now takes `metadata` as input and gets the config (presumably memoized, but likely not a big factor for performance anyway). 

I have added e2e tests, unit tests and more clear policy tests now. 

Can you have another look when you have time and let me know what you think.

README.md
anderseknert: Awesome!

bundle/regal/main_test.rego
```diff
@@ -105,7 +105,33 @@ test_main_force_exclude_file_eval_param {
...
+test_main_exclude_files_rule_config_integration {
+	policy := `package p
+
+	camelCase := "yes"
+	`
+	report := data.regal.main.report with input as regal.parse_module("/some/dir/p.rego", policy)
+		with data.regal_user_config as {"rules": {"style": {"prefer-snake-case": {
+			"level": "error",
+			"ignore": {"files": ["p.rego"]},
+		}}}}
```
anderseknert, line 134: This is perfect! Just what I was looking for 👍

e2e/cli_test.go
```diff
@@ -183,6 +183,69 @@ func TestLintAllViolations(t *testing.T) {
...
+	tc := "prefer-snake-case"
+
+	if _, ok := violationNames[tc]; ok {
+		t.Errorf("expected violation for rule todo-comments to be ignored")
+	}
+
+	// Should be ignored, so exclude from below check
+	delete(ruleNames, tc)
+
+	//if len(ruleNames) != len(violationNames) {
```
anderseknert, line 240: Let's clean this up before merge :)

anderseknert (approved):
This is starting to look great! Anything left to do here now? Excited to have this in the next release 😃

pkg/config/config.go
```diff
@@ -110,6 +117,17 @@ func (rule *Rule) UnmarshalJSON(data []byte) error {
...
+
+		err := rio.JSONRoundTrip(ignore, &dst)
+		if err != nil {
+			return fmt.Errorf("unmarshalling rule ignore failed: %w", err)
+		}
+
+		rule.Ignore = dst
+	}
+
 	delete(result, "enabled")
```
kristiansvalland, line 131: Wonder if we should delete "level" and "ignore" as well, or is there a particular reason this isn't already done in the case of "level"?

e2e/cli_test.go
kristiansvalland: Thanks! Removed the dead code.

kristiansvalland:
> This is starting to look great! Anything left to do here now? Excited to have this in the next release 😃

I think that the base functionality is implemented and should be good to go 😄 I did some minor changes since you last reviewed it, specifically in relation to leading "/" in patterns to make it work even more identical to .gitignore, but I do not see the need for any more changes, at least not any major ones.

pkg/config/config.go
anderseknert: "enabled" is what we used to call it (and it was a boolean value) before we changed to "level" to allow more granual settings, so this should be dead code. I'll have it removed in a followup PR.

anderseknert:
👏 👏 👏

PR #201, opened 2023-07-17
Enhanced type checking using Regal AST schema

### LINKED ISSUE #52: Provide a JSON schema for the AST + custom additons
For both our own use, as well as for authors of custom rules, it would be really useful if we had a JSON schema defining the AST JSON, including any custom additions we have added. This would help catch mistakes in Regal policy, and could be integrated as part of our build pipeline.

anderseknert:
Currently enabled for the `regal test` command, which is the most likely entrypoint for rule authors, whether for built-in Regal rules or custom ones.

While it would be useful to provide this also for `regal lint` — we'll probably want to do this together with the strictness rules, as we currently don't require compilation, and it's not clear that we'll want to unless a rule is configured that needs this, like most of the strictness rules would.

Additionally, the schema is provided when running the Rego tests via `go test ./...`, which should help catch errors we make ourselves.

It would be nice if custom rule authors didn't have to include the schema in the metadata annotation, but I've yet to figure out the best way of doing that, as we can't bundle a custom policy and also have them read from disk without the bundle "owning" that path. I suppose we could just parse a policy from a string, including a subpackage scoped custom.regal.rules package annotation... but I'll need to think more about this.

Fixes #52

### DIFF
--- a/main.go
+++ b/main.go
@@ -6,6 +6,7 @@ import (
 	"os"
 
 	"github.com/styrainc/regal/cmd"
+	"github.com/styrainc/regal/internal/embeds"
 )
 
 // Note: this will bundle the tests as well, but since that has negligible impact on the size of the binary,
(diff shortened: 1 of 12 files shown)


main.go
```diff
@@ -14,14 +15,19 @@ import (
...
 	log.SetFlags(0)
 
-	// The embedded FS can't point to parent directories, so while not pretty, we'll
-	// need to pass it from here to the next command
-	cmd.EmbedBundleFS = bundle
+	// Embedded files and directories can only traverse down in the tree, i.e. no parent paths,
+	// so while this isn't pretty, we'll use a separate package to carry state from here. If you
+	// know of a better way of doing this, don't hesitate to fix this.
+	embeds.EmbedBundleFS = bundle
+	embeds.ASTSchema = schema
```
charlieegan3, line 30: thought: hmm, yeah. I don't know of a better way to do this. I think it's best to have the bundle at the top level. I wonder if schema needs to be too though? Perhaps it could sit under `embeds`

charlieegan3 (approved):
This looks good to me! Tried it out too 🙂 

```
$ go run main.go test e2e/testdata/ast_type_failure/
1 error occurred: e2e/testdata/ast_type_failure/custom_type_fail.rego:13: rego_type_error: undefined ref: input.foo
        input.foo
              ^
              have: "foo"
              want (one of): ["annotations" "comments" "imports" "package" "regal" "rules"]
exit status 1
```

main.go
anderseknert: True. I'll leave it as is for now, but I might move the schema into embeds later 👍

Issue #185, opened 2023-06-27
Move `use-in-operator` to idiomatic category


anderseknert:
I think this should be considered idiomatic more than style advice, but that category did not exist before.

What needs to be done:

- Move the `use-in-operator` rule and its test file to the `idiomatic` directory
- Change the package name to reflect the new category
- Copy the old documentation to the idiomatic category under the docs directory, update any references in the doc
- Update the documentation in the style category to say the rule has moved to the idiomatic category, and provide a link
- Update the table in the README by running `go run main.go table --write-to-readme bundle`

PR #213, opened 2023-07-22
#185, moving the use-in-operator to idiomatic and its relevant files

Parsifal-M:
Hey @anderseknert and Team! 👋 

Thought I'd try to lend a hand!

In terms of the `README`, I was not 100% sure how you envisioned it e.g.:

- Leave the readme where it is and update it saying it moved.
- Move the readme to the new location under `idiomatic` and update it saying it was previously under `style`

Let me know what you think is best here and I will update it 😄 

Not sure if I missed anything else! Please LMK and I'll get back to it!

This Fixes https://github.com/StyraInc/regal/issues/185

Thanks 🚀

### DIFF
--- a/docs/rules/idiomatic/use-in-operator.md
+++ b/docs/rules/idiomatic/use-in-operator.md
@@ -2,7 +2,7 @@
 
 **Summary**: Use `in` to check for membership
 
-**Category**: Style
+**Category**: Idiomatic
 
 **Avoid**
 ```rego
@@ -39,12 +39,8 @@ This linter rule provides the following configuration options:
 
 ```yaml
 rules: 
-  style:
+  idiomatic:
     use-in-operator:
       # one of "error", "warning", "ignore"
       level: error
 ```
-
-## Related Resources
-
-- Rego Style Guide [Use `in` to check for membership](https://github.com/StyraInc/rego-style-guide#use-in-to-check-for-membership)
--- a/docs/rules/idiomatic/use-in-operator.md
+++ b/docs/rules/idiomatic/use-in-operator.md
@@ -44,3 +44,7 @@ rules:
       # one of "error", "warning", "ignore"
       level: error
 ```
+
+## Related Resources
+
+- Rego Style Guide [Use `in` to check for membership](https://github.com/StyraInc/rego-style-guide#use-in-to-check-for-membership)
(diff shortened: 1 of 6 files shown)


docs/rules/idiomatic/use-in-operator.md
```diff
@@ -39,12 +39,8 @@ This linter rule provides the following configuration options:
...
 ```yaml
 rules: 
-  style:
+  idiomatic:
     use-in-operator:
       # one of "error", "warning", "ignore"
       level: error
 ```
-
-## Related Resources
```
anderseknert, line 48: I don't think we should remove this. While it's a style guide, this is still a good reference.

docs/rules/idiomatic/use-in-operator.md
```diff
@@ -2,7 +2,7 @@
 
 **Summary**: Use `in` to check for membership
 
-**Category**: Style
```
anderseknert, line 5: Considering backwards compatibility for this one, and apologies for not having done so in the issue 😅 But I think it's probably good to keep the` docs/rules/style/use-in-operator.md` file around for a while but have it contain only a link to the doc in the idiomatic category. That way, users of older versions would still find the docs and be able to update their config if needed. Does that make sense?

anderseknert (commented):
Much appreciated, Peter! Just dropped a comment or two, but this is good 🙂

anderseknert:
Also, don't forget to add "Fixes #185" to the commit message / PR description, in order to link the PR to the issue 🙂

docs/rules/idiomatic/use-in-operator.md
Parsifal-M: Ah yeah that makes sense now I think so basically two readme's but the one in `docs/rules/style/use-in-operator.md` just points to the "new" one (of course nicely) in `docs/rules/idiomatic/use-in-operator.md` am I following you on this? 😅

anderseknert: Yep! We could just say something like "Rule moved to idiomatic category" and provide a link to its new location.

Parsifal-M: Hey! 👋 

LMK what you think, happy to change it a bit more if needed 👍 . I added back the style reference guide and added a note to the old docs pointing to the new ones.

docs/rules/idiomatic/use-in-operator.md
Parsifal-M: Added it back 👍

anderseknert (approved):
Thanks Peter! 👏

Issue #206, opened 2023-07-19
Scaffolding command for new built-in and custom rules


anderseknert:
Something like:
```
regal new rule --type custom --category naming --title lower-case
```
Which would:
1. Create a `.regal` directory if none exists.
2. Create a `.regal/rules` directory if none exists.
3. Create a `.regal/rules/category/title.rego` file pre-populated with the basic required policy structure.
4. Create a `.regal/rules/category/title_test.rego` file pre-populated with the basic required policy test structure.

And something similar to the above for creating new built-in rules as well, where the new files are added to the Regal bundle instead, and where there's also an entry added for documentation in the docs directory.

PR #214, opened 2023-07-25
VSCode snippets

ThomasKingoTM:
Added snippets for VSCode to pre-populate 3 types of files: 
1. Policy file
2. Test file
3. Example policy (e.g. for parsing)

On Mac, it is to be placed in Users/username/Library/Application_Support/Code/User/Snippets. 
On Windows, it is to be placed in Windows/Users/username/AppData/Roaming/Code/User/Snippets.

Otherwise you can do the following in VSCode:
Preferences -> Configure User Snippets -> Type "rego" -> Copy paste this file. 

The snippets are known by their prefix/keywords. 
1. Policy: "policy", "rule", "template"
2. Test: "regal_test", "testing", "template"
3. Example: "regal_example", "template"

### DIFF
--- a/rego.json
+++ b/rego.json
@@ -88,10 +88,10 @@
             "package $1",
             "",
             "",
-            "decision := {allow: true} {",
+            "decision := {\"allow\": true} {",
             "\tinput.foo",
             "}",
-            "decision := {allow: false} {",
+            "decision := {\"allow\": false} {",
             "\tnot input.foo",
             "}"
         ],
--- a/rego.json
+++ b/rego.json
@@ -1,0 +1,100 @@
+{
+    "policy template": {
+        "prefix": [
+            "policy",
+            "rule",
+            "template"
+        ],
+        "body": [
+            "# METADATA ",
+            "# description: $1",
+            "package custom.regal.rules.$2",
+            "",
+            "import future.keywords.in",
+            "import future.keywords.if",
+            "import future.keywords.contains",
+            "",
+            "import data.regal.result",
+            "",
+            "report contains violation if {",
+            "\t$3",
+            "\t",
+            "\t",
+            "\tviolation := result.fail(regal.metadata.chain(), result.position($4))",
+            "}"
+        ],
+        "description": "Regal policy template"
+    },
+    "policy test": {
+        "prefix": [
+            "regal_test",
+            "testing",
+            "template"
+        ],
+        "body": [
+            "package custom.regal.rules.$1",
+            "",
+            "import future.keywords.if",
+            "import future.keywords.in",
+            "import future.keywords.contains",
+            "",
+            "import data.regal.ast",
+            "import data.regal.config",
+            "import custom.regal.rules.$2 as rule",
+            "",
+            "test_fail_$3 if {",
+            "\tr := rule.report with input as ast.policy(",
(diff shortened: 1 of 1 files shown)


rego.json
```diff
@@ -0,0 +1,100 @@
+{
+    "policy template": {
+        "prefix": [
```
charlieegan3, line 3: suggestion: since the templates are Regal specific, rather than for any Rego policy, it might make sense to have that as the matching prefixes? We seem to have this in the other templates.

rego.json
```diff
@@ -0,0 +1,100 @@
...
+            "",
+            "",
+            "decision := {\"allow\": true} {",
+            "\tinput.foo",
+            "}",
+            "decision := {\"allow\": false} {",
+            "\tnot input.foo",
+            "}"
+        ],
+        "description": "Regal example template"
```
charlieegan3, line 98: suggestion: I feel like this example is more a Rego example than a Regal example? I wonder if it makes sense to have a complete regal example instead?

charlieegan3 (commented):
Hi Thomas 👋 this looks good. I tested it out and it seems to work.

Video of my testing: https://github.com/StyraInc/regal/assets/1774239/75bdf3be-6560-4284-b64f-98444054faad

I had some ideas for things we might want to do before we merge this in, feel free to take on as much or as little as you like:

### How can we teach users how to use this?

You've written a great little description of how to use the JSON in VSCode in the PR description. It's be good to get this into a place where users of the repo or Regal can see it too.

Stretch goal: is there a way we can automate the installation? Perhaps we could extend Regal with a command to install this file in the correct location depending on the user's operating system? it'd need to:
* make sure it created the nested dir if it didn't exist
* update the file if it already existed, merging the templates with any other rego templates the user had.
* Include the JSON file here inside the Regal binary and expose it via a command like like `regal tooling vscode install-templates`.

rego.json
ThomasKingoTM: Correct - the example is to create a rego file identical to what currently results from calling ast.policy. Thus, hopefully recreating a rego file identical to what is input to the tests.

rego.json
ThomasKingoTM: So simply include "regal" as prefix for their names?

ThomasKingoTM:
Hi Charlie :) 
Thanks for reviewing it! 

First off, I would definitely put the description of the templates inside the docs for custom rules and development. 
- possibly even with a link to your video so users can see what the snippets do? 

The stretch goal is slightly too big for what I can finish with the amount of hours I will be allocating in the near future - someone else might have to look into it.

bdjgs:
I work together with Thomas, so I, or another from our team will probably try and get this PR merged.

One thing I thought of is: Is this even the way to do this?

In this https://github.com/StyraInc/regal/issues/206 Anders suggests to implement some scaffolding as a regal command. That is probably not the way to go with this (but I dont know). 

But also, not everyone uses VSCode, so is it better to add a Makefile and associated helper bash scripts to create stuff like this? In that way all can use it and not just those who use VSCode.

bdjgs:
I can see that #206 has been implemented, so I think we can just close this PR. Currently, Thomas is not available, but I will tell him to close this PR when we gets back :)

Issue #152, opened 2023-06-08
Boolean rule without default condition


anderseknert:
It would be a good addition to have a rule that warns on use of boolean rules that are missing a default condition. Obviously this isn't wrong per se, so this falls into the style category (I think). It should also be possible to provide a list of rules for which does not apply.

srenatus:
Hm I'm on the fence here. I'd tend to lean into undefined; and many rules are boolean, basically all the simple ones...

```rego
default allow := false
allow if user_is_admin
allow if endpoint_is_public
allow if {
  # more complex stuff
}

user_is_admin if ...

endpoint_is_public if ...
```

Would you think that `user_is_admin` and `endpoint_is_public` needs a default rule?

anderseknert:
I think you're right — this is really only valuable for "main" rules. The only way we have of knowing which ones are would be via the `entrypoint` annotation. We currently provide no recommendation for that, but perhaps we should (in a different rule that is)? Having one serves as good documentation, and it also makes it easier to compile the policy to different targets. If we made it a best practice to always have at least one, we could have this rule enforce that at least boolean _entrypoints_ have a default rule.

We'd need a new rule type for this I guess, which would check the aggregate of all files rather than each file individually.

srenatus:
I guess there's another heuristic: boolean rules that aren't called anywhere. It would cover my example nicely; and still doesn't require entrypoint metadata.

anderseknert:
Yeah I suppose so, but it could just be that they're unused and actually should be removed? 🤔

srenatus:
hmm warn about rules that aren't declared entrypoint and aren't called...?

```rego
package foo
baz := true
```

```rego
package bar
q := data.foo[input]
```

I guess we're too dynamic to reliably do this... 🤔

anderseknert:
Let's make entrypoint annotations a recommendation then. Best practices born out of laziness 😅

anderseknert:
Thinking about it, this would make sense as a `custom` rule where the user could simply say "for any rule named `allow`, there must be an accompanying default rule". Same for entrypoint annotations: "for all modules, there must be at least one entrypoint defined".

See https://github.com/StyraInc/regal/issues/282 for some design ideas.

Issue #51, opened 2023-03-21
Directory tree policy rules?


anderseknert:
One thing that would be interesting (and potentially useful) is if Regal included some optional rules to check the directory structure of a policy repo, and reported on things that could be improved:

- corresponding `*_test.rego` for any rego file
- or a separate test directory
- file names match package names
- `.editorconfig`
- "bundle mode" checking for `.manifest`

And so on... I'm sure we could come up with more. Low priority, but noting it down for later :)

anderseknert:
Adding `aggregate` tag for this one, as we're unlikely to want to run this agains each input file, but something like the `aggregate_report` rule introduced recently.

PR #215, opened 2023-07-26
add testing helper for creating policy bundles

bdjgs:
I was creating a custom rule implemented in Go, just like OpaFmtRule.
The rule lints an entire policy bundle. In this case, it asserts that masking is implemented under system/log/log.rego.

When testing the rule, I was missing a way to provide a set of policy files as input.

**Usage**
```
policyFileName := "api/authz/policy.rego"
policyFileContent := `
package api.authz

##content of policy##
`
systemLogFileName := "system/log/log.rego"
systemLogFileContent := `
package system.log

##content of policy##
`

bundle := map[string]string{
    policyFileName:    policyFileContent,
    systemLogFileName: systemLogFileContent,
}

result, err := rules.NewMaskingRule(config.Config{}).Run(ctx, test.InputBundle(bundle))
```

I am new to regal, so I am open for any suggestions :)

### DIFF
--- a/internal/test/test.go
+++ b/internal/test/test.go
@@ -45,3 +45,18 @@ func InputPolicy(filename string, policy string) rules.Input {
 
 	return rules.NewInput(content, modules)
 }
+
+// InputBundle allows for constructing a policy bundle
+// for testing rules that perform linting of an entire policy bundle.
+// policyBundle represents the bundle as a map from filename to filecontent.
+func InputBundle(policyBundle map[string]string) rules.Input {
+	content := make(map[string]string)
+	modules := make(map[string]*ast.Module)
+
+	for filename, filecontent := range policyBundle {
+		content[filename] = filecontent
+		modules[filename] = parse.MustParseModule(filecontent)
+	}
+
+	return rules.NewInput(content, modules)
+}


anderseknert:
Hey @bdjgs! And thanks for submitting this 👍  I love seeing Regal not just used, but extended for custom use cases! 

As for this case, and others where we need Regal to be aware of more than just the AST of a single input file, I think we're gonna need to flesh out a solution that will cover these requirements without giving up on Rego. While it's perfectly fine to extend Regal via Go, such extensions basically require the project to be forked, and the fork to be maintained over time. If someone wants to do that, that's cool. It's open source after all :) But I'd rather provide an easier option for this category of rules — not just for authors of custom rules, but just as much for Regal itself, as I imagine we'll want to add quite a few builtin rules that make use of this. This has been discussed in the past, both [here](https://github.com/StyraInc/regal/issues/152#issuecomment-1582530032), [there](https://github.com/StyraInc/regal/issues/51) and elsewhere. 

I've got about a week left of my vacation, but perhaps we could sync on this once I'm back? I have a pretty good idea of what I think would be needed here, and I don't think it would be that demanding to add, but I'd love to have your use case(s) in mind, to make sure we cover all angles. If we could work together on a solution, even better! 😃

bdjgs:
That sounds great 👍

For now, I will just close the PR and then we can sync on this in the future :)

PR #216, opened 2023-07-27
ignore violations with a level lower than fail-level

bdjgs:
I personally dont think that the final violation report should contain violations below the configured fail-level. If I use regal to linter my policy, but are only interested in error level violations, I dont want the final report to contain all the warning level violations.

This change makes it so that if I run:
`go run main.go lint test.rego --fail-level warning` the violation report will contain violations with level warning or error.
and if I run:
`go run main.go lint test.rego --fail-level error` the violation report will only contain violations with level error. 

I might have overseen something related to the fields inside the Summary object :) 

Also, I am complety new to regal, so I might be wrong on this and the report should just contain all violations no matter the fail-level :)

### DIFF
--- a/cmd/lint.go
+++ b/cmd/lint.go
@@ -259,7 +259,9 @@ func lint(args []string, params lintCommandParams) (report.Report, error) {
 		return report.Report{}, fmt.Errorf("failed to get reporter: %w", err)
 	}
 
-	return result, rep.Publish(result) //nolint:wrapcheck
+	filteredReport := filterReport(result, params.failLevel)
+
+	return result, rep.Publish(filteredReport) //nolint:wrapcheck
 }
 
 func getReporter(format string, outputWriter io.Writer) (reporter.Reporter, error) {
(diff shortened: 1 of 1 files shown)


bdjgs:
I will be unavailable for the next couple of weeks. So for now, I will just close the PR and then reopen it at a later point.

anderseknert:
This is a good discussion to have, so let's get back to it once you're back from vacation (and enjoy that first, of course!)

I don't think "level" here should be thought of as something as levels in logging. IMHO, there are only two "real" levels at play here — one either thinks a rule is valid and should be included in linting, or it shouldn't. As such, the `warning` "level" should not be thought of as rules of lesser importance, but as an option for violations one wants to be made aware of in e.g. CI/CD, but that should _not fail the build_ (e.g. not themselves set a non-zero exit code) for the time being.

This would most likely be a temporary measure for when you introduce Regal in a large existing policy repo and might find yourself with thousands of violations from e.g. the style category. You could either choose to ignore them for the moment, or have them included in the report but ensure the project still builds unless rules configured as errors (or not at all, as error is the default) yield violations.

That said, I would not mind adding options to make the various output formats more flexible / configurable... but if we add support for that I think it should be done separately, and independently, of `--fail-level`.

PR #221, opened 2023-08-02
feat: add pre-commit hook

adam-moss:
Add a `.pre-commit-hooks.yaml` file so that `regal`, more specifically `regal lint` can be utilised as a pre-commit hook via the popular [pre-commit](https://pre-commit.com) project.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,17 @@
+- id: regal-lint
+  stages:
+    - pre-commit
+  language: golang
+  name: policy file linting
+  description: policy file linting with Regal from Styra built from source
+  entry: regal lint
+  files: (\.rego)$
+
+- id: regal-lint-use-path
+  stages:
+    - pre-commit
+  language: system
+  name: policy file linting
+  description: policy file linting with Regal from Styra using system $PATH
+  entry: regal lint
+  files: (\.rego)$
--- a/README.md
+++ b/README.md
@@ -281,6 +281,40 @@ Note that at this point in time, Regal only considers the line following the ign
 entire blocks of code (like rules, functions or even packages). See [configuration](#configuration) if you want to
 ignore certain rules altogether.
 
+## Using Regal with Pre-Commit
+
+[Pre-Commit](https://pre-commit.com) is a framework for managing and maintaining multi-language pre-commit hooks.
+
+To use Regal with pre-commit, add this to your `.pre-commit-config.yaml`
+
+```yaml
+- repo: https://github.com/StyraInc/regal
+  rev: v0.5.0 # Use the ref you want to point at
+  hooks:
+    - id: regal-lint
+  # -   id: ...
+```
+
+### Hooks Available
+
+#### `regal-lint`
+
+![commit-msg hook](https://img.shields.io/badge/hook-pre--commit-informational?logo=git)
+
+Runs Regal against all staged `.rego` files, aborting the commit if any fail.
+
+- requires the `go` build chain is installed and available on `$PATH`
+- will build and install the tagged version of Regal in an isolated `GOPATH`
+- ensures compatibility between versions
+
+#### `regal-lint-use-path`
+
+![commit-msg hook](https://img.shields.io/badge/hook-pre--commit-informational?logo=git)
+
+Runs Regal against all staged `.rego` files, aborting the commit if any fail.
+
+- requires the `regal` package is already installed and available on `$PATH`.
+
 ## Resources
 
 ### Documentation


anderseknert:
Hey Adam! And thanks for this 👍 I'll take a closer look tomorrow

```
language: golang
```
  name: policy file lint
  description: policy file linting with regal from Styra
  entry: regal lint
  files: (\.rego)$
```

File: .pre-commit-hooks.yaml [hunk 1]
```diff
@@ -0,0 +1,8 @@
+- id: regal-lint
+  stages:
+    - pre-commit
+  language: golang
```
anderseknert, line 4: This should be `system`, IIRC. We're calling the `regal` binary here, not depending on the Golang tooling.

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,8 @@
...
+  name: policy file lint
+  description: policy file linting with regal from Styra
```
anderseknert, line 6: regal -> Regal

anderseknert (commented):
Hey! And thanks for submitting this 👍 It's been a while since I last worked on [pre-commit hooks](https://github.com/anderseknert/pre-commit-opa), so this was a good reminder :) This seems small enough to warrant inclusion here, or else I'd probably prefer it in a separate repo. Left two minor suggestions for fixes.

anderseknert:
Sorry, not sure why this got closed as I submitted the comments. That was not my intention.

.pre-commit-hooks.yaml
adam-moss: Amended 👍

.pre-commit-hooks.yaml
adam-moss: I've went with `golang` as that ensures `pre-commit` installs the `regal` binary for the given tag in this repository, thus maintenance of the hook is simplified if anything needs to change / be added.

If we use `system` it requires the user has a pre-existing installation of `regal` that they'd need to keep updated.

anderseknert: I suppose we could have to options, one which was `regal-lint-go` (or whatever) and one `regal-lint` or something like that.

c-wygoda: I'd be in favor of `system` - the casual OPA user or policy editor might not even have Golang installed.

c-wygoda: probably a wrapper script to install the right binary for the current architecture might be handy though.

adam-moss: To summarise:

1) `language: golang`

`pre-commit` will install the tagged version by doing a `go install ./...` in an isolated `GOPATH`.

In this scenario the consumer doesn't have to have `regal` on their `PATH`, but needs the `go` build chain.  `pre-commit` will use the tagged version of `regal` from the repo.  If `regal` does exist on `PATH` the user _may_ experience differences in functionality if the versions are misaligned.

2) `language: system`

`pre-commit` will not install `regal` and will simply execute whatever version is on `PATH`.

In this scenario the consumer doesn't have to have `go` build chain on their `PATH`, but needs the `regal` binary.  `pre-commit` will fail if `regal` does not exist.  There is an additional risk, likely small (?), that if the _interface_ of the `regal` binary changed the `pre-commit` hook would fail due to version mismatching.

Happy to take your guidance here; personally I favour (1).

adam-moss: I could add hooks with `id: regal-lint` and `id: regal-lint-from-system` to give both options?

anderseknert: That sounds like a good middle ground to me 👍

adam-moss: Updated 👍

anderseknert: Thanks! Do you think you could add a section on this to the README? I'm thinking a new one called "Pre-Commit Hooks" below the "Inline Ignore Directives" section, where you provide instructions for how to install the two different versions, and how they're different?

If not, I can have it added later.

adam-moss: Updated with README as request

README.md
```diff
@@ -281,6 +281,38 @@ Note that at this point in time, Regal only considers the line following the ign
 entire blocks of code (like rules, functions or even packages). See [configuration](#configuration) if you want to
 ignore certain rules altogether.
 
+## Using Regal with Pre-Commit
+
+Add this to your `.pre-commit-config.yaml`
```
anderseknert, line 286: For people unfamiliar with the pre-commit project, it'd be good to add a link here.

README.md
```diff
@@ -281,6 +281,38 @@ Note that at this point in time, Regal only considers the line following the ign
...
+
+```yaml
+- repo: https://github.com/StyraInc/regal
+  rev: v.5.0 # Use the ref you want to point at
```
anderseknert, line 290: Should be v0.5.0 right?

anderseknert (commented):
Thanks! Just two comments, then we're good to merge this :)

README.md
adam-moss: 🤦 , fixed!

README.md
adam-moss: Updated with link and minor verbiage, wdyt?

c-wygoda:
Once this is merged, I have a follow-up PR with a `regal-download` hook which fetches the binary from Github.

anderseknert (approved):
Thanks Adam!

PR #225, opened 2023-08-03
Rule: non-raw-regex-pattern

### LINKED ISSUE #222: Enforce backtick strings for regex literal
As the style guide foretold, thou shall [use raw strings for regex patterns](https://github.com/StyraInc/rego-style-guide#use-raw-strings-for-regex-patterns).

Can't get this from the AST, but we can find the location of the calls, and try to parse the text from there.

anderseknert:
Following the Rego Style Guide, this rule ensures regex patterns are provided as raw strings in order to avoid having to escape things like `\` inside of the regex.

Fixes #222

### DIFF
--- a/docs/rules/idiomatic/non-raw-regex-pattern.md
+++ b/docs/rules/idiomatic/non-raw-regex-pattern.md
@@ -1,0 +1,60 @@
+# non-raw-regex-pattern
+
+**Summary**: Use raw strings for regex patterns
+
+**Category**: Idiomatic
+
+**Avoid**
+```rego
+all_digits if {
+    regex.match("[\\d]+", "12345")
+}
+```
+
+**Prefer**
+```rego
+all_digits if {
+    regex.match(`[\d]+`, "12345")
+}
+```
+
+## Rationale
+
+[Raw strings](https://www.openpolicyagent.org/docs/edge/policy-language/#strings) are interpreted literally, allowing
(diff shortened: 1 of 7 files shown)


docs/rules/idiomatic/non-raw-regex-pattern.md
```diff
@@ -0,0 +1,60 @@
...
+package policy
+
+import future.keywords.if
+
+# Pattern assigned to variable
+pattern := "[\\d]+"
+
+# This won't trigger a violation
+allow if regex.match(pattern, "12345")
+```
```
charlieegan3, line 43: is it worth opening an issue for this too?

charlieegan3 (approved):
Looks good to me!

docs/rules/idiomatic/non-raw-regex-pattern.md
anderseknert: Yeah, not sure about this specifically, but we could definitely add a feature request for being able to track values in scope. I don't think it can be easily done in Rego alone, at least not for more advanced cases where the var could be assigned the value of another var, which could be assigned the value of another var, etc... but perhaps a simpler solution would be good enough for most use cases.

anderseknert: https://github.com/StyraInc/regal/issues/229

PR #226, opened 2023-08-04
Add self-downloading pre-commit hook

c-wygoda:
- Using a shell script, check if `regal` binary is not yet present in pre-commit repo clone and download right binary for system and architecture.
  - should work for Linux and MacOS, only tested on Linux atm.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -15,3 +15,12 @@
   description: policy file linting with Regal from Styra using system $PATH
   entry: regal lint
   files: (\.rego)$
+
+- id: regal-download
+  stages:
+    - pre-commit
+  language: script
+  name: policy file linting
+  description: policy file linting with Regal downloaded from Github
+  entry: build/download-and-run.sh lint
+  files: (\.rego)$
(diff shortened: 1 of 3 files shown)


download-and-run.sh
anderseknert, line 1: Could you please move this to the `build` directory? I'd rather keep the number of files in the top-level directory to a minimum. 

Also, please add a comment in this file explaining its purpose, as that's not clear unless you're aware of the pre-commit configuration.

anderseknert (commented):
Looks good to me! Just a minor request.

anderseknert:
Not related to the PR, but I think the docs around this are claiming a bit too much space in the README, and even more so once this is merged. So I'll follow up with a change later moving these instructions to a sub-page under `docs/`, and will link to that from the README. I hope you don't mind. Also FYI @adam-moss.

c-wygoda:
> Not related to the PR, but I think the docs around this are claiming a bit too much space in the README, and even more so once this is merged. So I'll follow up with a change later moving these instructions to a sub-page under `docs/`, and will link to that from the README. I hope you don't mind. Also FYI @adam-moss.

I don't mind. We can probably also condense to one paragraph with giving the three different methods as a short list.

anderseknert (approved):

anderseknert:
Thanks @c-wygoda 👍

PR #237, opened 2023-08-09
Use H1 for pre-commit hook title

charlieegan3:
This matches the other top level docs pages.


anderseknert (approved):
Good catch! Yep, copy-paste mistake :)

PR #245, opened 2023-08-10
Move "Exceptions" to after "Rationale"

anderseknert:
IMHO it makes sense to first present *why*, then the cases for when exceptions might make sense.

Also made H2's for these sections.


charlieegan3 (approved):
Yeah, agree that this makes sense.

PR #249, opened 2023-08-10
Use docs.styra.com for published docs

anderseknert:
While GitHub pages are OK for reading up on a single rule, they don't provide a very nice experience for more exploratory browsing, and we definitely want to encourage that.

Use docs.styra.com as the public facade for the docs. The GitHub docs are still the _source_ and obviously available for browsing too, should someone prefer that.


charlieegan3 (approved):
Looks good to me!

Issue #3, opened 2023-01-24
Rule: unification used for variable assignment


anderseknert:
**Avoid**
```rego
x = 5
```
**Prefer**
```rego
x := 5
```
Motivation: while there are a few valid cases for unification, the unification operator (`=`) should not be used for simple assignment. TBD — more comprehensive explanation.

anderseknert:
Closing in favor of #4, which will be remodeled to be "Unnecessary unification".

PR #254, opened 2023-08-14
Rule: use-some-for-output-vars

### LINKED ISSUE #250: Use `some` for output variables
**Avoid**
```rego
package policy

allow {
    userinfo := data.users[id]
    # ...
}
```

**Prefer**
```rego
allow {
    some id
    userinfo := data.users[id]
    # ...
}

# alternatively, and arguably better, but there are other rules for that
allow {
    some id, userinfo in data.users
    # ...
}
```

https://www.openpolicyagent.org/docs/latest/policy-language/#some-keyword

anderseknert:
This commit adds AST functions to find vars in scope of a location, which means scanning "up" in the scope for e.g. rules or other vars to check if they are present. This allows us to see if a possible "output var" is defined using some, or if it is an "input" var that's assigned elsewhere in the policy.

Fixes #250

### DIFF
--- a/bundle/regal/ast.rego
+++ b/bundle/regal/ast.rego
@@ -24,6 +24,11 @@ tests := [rule |
 	startswith(rule.head.name, "test_")
 ]
 
+rule_names := {rule.head.name |
+	some rule in input.rules
+	not rule.head.args
+}
+
 # METADATA
 # description: parse provided snippet with a generic package declaration added
 policy(snippet) := regal.parse_module("policy.rego", concat("", [
@@ -103,6 +108,13 @@ _find_every_vars(_, value) := var if {
 	var := array.concat(key_var, val_var)
 }
 
+_find_set_or_array_comprehension_vars(value) := [value.value.term | value.value.term.type == "var"]
+
+_find_object_comprehension_vars(value) := array.concat(key, val) if {
+	key := [value.value.key | value.value.key.type == "var"]
+	val := [value.value.value | value.value.value.type == "var"]
+}
+
 _find_vars(path, value, last) := _find_assign_vars(path, value) if {
 	last == "terms"
 	value[0].type == "ref"
@@ -125,16 +137,60 @@ _find_vars(path, value, last) := _find_every_vars(path, value) if {
 	value.domain
 }
 
+_find_vars(_, value, _) := _find_set_or_array_comprehension_vars(value) if {
+	value.type in {"setcomprehension", "arraycomprehension"}
+}
+
+_find_vars(path, value, _) := _find_object_comprehension_vars(value) if {
+	value.type == "objectcomprehension"
+}
+
 # METADATA
 # description: |
 #   traverses all nodes under provided node (using `walk`), and returns an array with
-#   all variables declared via assignment (:=), `some` or `every`
+#   all variables declared via assignment (:=), `some`, `every` and in comprehensions
 find_vars(node) := [var |
 	# regal ignore:function-arg-return
 	walk(node, [path, value])
 
 	some var in _find_vars(path, value, regal.last(path))
 ]
+
+_function_arg_names(rule) := {arg.value |
+	some arg in rule.head.args
+	arg.type == "var"
+}
+
+# METADATA
+# description: |
+#  finds all vars declared in `rule` *before* the `location` provided
+#  note: this isn't 100% accurate, as it doesn't take into account `=`
+#  assignments / unification, but it's likely good enough since other rules
+#  recommend against those
+find_vars_in_local_scope(rule, location) := [var |
+	some var in find_vars(rule)
+	_before_location(var, location)
+]
+
+_before_location(var, location) if {
+	var.location.row < location.row
+}
+
+_before_location(var, location) if {
+	var.location.row == location.row
+	var.location.col < location.col
+}
+
+# METADATA
+# description: |
+#  similar to `find_vars_in_local_scope`, but returns all variable names in scope
+#  of the given location *and* the rule names present in the scope (i.e. module)
+find_names_in_scope(rule, location) := names if {
+	fn_arg_names := _function_arg_names(rule)
+	var_names := {var.value | some var in find_vars_in_local_scope(rule, location)}
+
+	names := (rule_names | fn_arg_names) | var_names
+}
 
 # METADATA
 # description: |
--- a/docs/rules/idiomatic/use-some-for-output-vars.md
+++ b/docs/rules/idiomatic/use-some-for-output-vars.md
@@ -1,0 +1,102 @@
+# use-some-for-output-vars
+
+**Summary**: Use `some` to declare output variables
+
+**Category**: Idiomatic
+
+**Avoid**
+```rego
+package policy
+
+allow {
+    userinfo := data.users[id]
+    # ...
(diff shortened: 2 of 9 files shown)


bundle/regal/ast.rego
```diff
@@ -136,6 +141,32 @@ find_vars(node) := [var |
...
+_before_location(var, location) if {
+	var.location.row == location.row
+	var.location.col < location.col
+}
+
+# METADATA
+# description: |
+#  similar to `find_vars_in_local_scope`, but returns all variable names in scope
+#  of the given location *and* the rule names present in the scope (i.e. module)
+find_names_in_scope(rule, location) := rule_names | {var.value | some var in find_vars_in_local_scope(rule, location)}
```
srenatus, line 168: oh that location for a set union took me a few seconds to understand. 😃

docs/rules/idiomatic/use-some-for-output-vars.md
```diff
@@ -0,0 +1,102 @@
...
+    # ...
+}
+```
+
+## Rationale
+
+An interesting, and likely unfamiliar aspect of Rego for developers coming from other languages, is the concept of
+[unification](https://en.wikipedia.org/wiki/Unification_(computer_science)). Unification happens not just explicitly via
+the unification operator (`=`), but is an integral part of Rego. In the context of this rule, unification means that a
+variable can either be an _input_ or an _output_. What does that mean?
```
srenatus, line 37: I suppose some of that would make sense to be more explicitly and thoroughly explained in the OPA docs, too.

srenatus (commented):
Those docs are good! I haven't payed much attention to the implementation _yet_.

bundle/regal/ast.rego
```diff
@@ -103,6 +108,13 @@ _find_every_vars(_, value) := var if {
 	var := array.concat(key_var, val_var)
 }
 
+_find_set_or_array_comprehension_vars(value) := [value.value.term | value.value.term.type == "var"]
```
srenatus, line 111: 🤔 there's no iteration here. Why an array?
```suggestion
_find_set_or_array_comprehension_vars(value) := value.value.term if value.value.term.type == "var"
```

bundle/regal/ast.rego
anderseknert: Hehe! I made it a bit more explicit now by using temporary vars before the union.

bundle/regal/ast.rego
```diff
@@ -136,6 +141,32 @@ find_vars(node) := [var |
 	some var in _find_vars(path, value, regal.last(path))
 ]
 
+# METADATA
+# description: |
+#  finds all vars declared in `rule` *before* the `location` provided
+#  note: this isn't 100% accurate, as it doesn't take into account `=`
+#  assignments / unification, but it's likely good enough since other rules
+#  recommend against those
```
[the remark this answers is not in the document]
anderseknert: Yep! "Luckily" we don't yet complain about `=` for [var assignments](https://github.com/StyraInc/regal/issues/3).

But that should be easier to implement now that we have the plumbing in place for determining whether `=` is used for assignment or comparison (i.e. whether the vars are inputs or outputs)... so that would be up next I guess.

bundle/regal/ast.rego
```diff
@@ -125,17 +137,61 @@ _find_vars(path, value, last) := _find_every_vars(path, value) if {
...
+
+# METADATA
+# description: |
+#  finds all vars declared in `rule` *before* the `location` provided
+#  note: this isn't 100% accurate, as it doesn't take into account `=`
+#  assignments / unification, but it's likely good enough since other rules
+#  recommend against those
+find_vars_in_local_scope(rule, location) := [var |
+	some var in find_vars(rule)
+	_before_location(var, location)
```
srenatus, line 172: Are we working with `ast.Module`s that have passed through the compiler...? Because IIRC we do some reordering there based on output variables, that might be relevant to consider. I suppose it could be possible that some policies are valid for OPA but icky in Regal...? 🤔

docs/rules/idiomatic/use-some-for-output-vars.md
anderseknert: Agreed.

srenatus (approved):
Just commentary and questions. Impressive work!

bundle/regal/ast.rego
anderseknert: No, we don't leverage the compiler in Regal at this point, but it's been considered for having something like the strict mode checks included. At that point we'd probably need to revisit some assumptions like you're on to 😅

bundle/regal/ast.rego
```diff
@@ -125,17 +137,61 @@ _find_vars(path, value, last) := _find_every_vars(path, value) if {
...
+
+# METADATA
+# description: |
+#  similar to `find_vars_in_local_scope`, but returns all variable names in scope
+#  of the given location *and* the rule names present in the scope (i.e. module)
+find_names_in_scope(rule, location) := names if {
+	fn_arg_names := _function_arg_names(rule)
+	var_names := {var.value | some var in find_vars_in_local_scope(rule, location)}
+
+	names := (rule_names | fn_arg_names) | var_names
```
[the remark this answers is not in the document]
anderseknert: No, it's `opa fmt` adding thos 😓 I think there's an open issue somewhere about that. Or was it perhaps another related one where that tool would actually _remove_ parenthesis and thus alter the actual policy? I can't remember. They are not added by me.

bundle/regal/ast.rego
```diff
@@ -103,6 +108,13 @@ _find_every_vars(_, value) := var if {
 	var := array.concat(key_var, val_var)
 }
 
+_find_set_or_array_comprehension_vars(value) := [value.value.term | value.value.term.type == "var"]
+
+_find_object_comprehension_vars(value) := array.concat(key, val) if {
+	key := [value.value.key | value.value.key.type == "var"]
+	val := [value.value.value | value.value.value.type == "var"]
+}
```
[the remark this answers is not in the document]
anderseknert: But then the assignment would fail if either the key or the value aren't vars, no?

srenatus (approved):

docs/rules/idiomatic/use-some-for-output-vars.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+Even though `name` is defined in the global scope, the `some` keyword will ensure it's now considered as an output
+variable in the local scope of the `usernames` rule.
+
+## Related Resources
+
+- Rego Style Guide [Don't use undeclared variables](https://github.com/StyraInc/rego-style-guide#dont-use-undeclared-variables)
+- OPA docs: [The `some` keyword](https://www.openpolicyagent.org/docs/latest/policy-language/#some-keyword)
+- Wikipedia [Unification](https://en.wikipedia.org/wiki/Unification_(computer_science))
```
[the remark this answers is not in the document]
anderseknert: You're right, but it's extensive enough as it is, I think, so maybe introducing even more articles would be too much for this page? 🤔 I suppose we could have a possibly optional rule to detect variable shadowing, so that even this:

```rego
foo := "bar"

allow {
    some foo
    bar[foo]
}
```
Was considered a violation?

PR #263, opened 2023-08-16
Rule: prefer-some-in-iteration

### LINKED ISSUE #243: Prefer some ... in for iteration
One of the few remaining items from the [style guide](https://github.com/StyraInc/rego-style-guide#prefer-some--in-for-iteration). This is obviously opinonated, and will come with a few (perhaps configurable) exceptions, like allowing nested iteration using `a[_].b[_].c[_]` style.

The first iteration of this could cover just the simpler cases using wildcards. In order to know if this is an iteration though:

```rego
a[i].b[j].c[k]
```
We'd need to know whether i, j or k was previously defined and in scope — i.e. if they are used as inputs our outputs — which we cannot yet. This would however be simpler than #229, as we wouldn't need to know the values, just that there's a var in scope.

anderseknert:
Building on top of previous work to find vars and to determine the "scope" of a variable, this commit adds a new rule in the style category to prefer `some x in y` over `x := y[_]`.

There are also a few fixups here related to the new rule, and some that were forgotten about in the previous rule added.

Note that there are some situations where a variable may thought of as being in scope while not actually being in scope — such as when iteration happens after a comprehension, as we have no way of keeping the nesting level "state" currently. This should not lead to false positives but could lead to a few cases being missed, which I think is OK for the time being. Will need to investigate better methods for determining scope, and likely outside of Rego.

Fixes #243

### DIFF
--- a/docs/rules/style/prefer-some-in-iteration.md
+++ b/docs/rules/style/prefer-some-in-iteration.md
@@ -1,0 +1,117 @@
+# prefer-some-in-iteration
+
+**Summary**: Prefer `some .. in` for iteration
+
+**Category**: Style
+
+**Avoid**
+```rego
+package policy
+
+engineering_roles = {"engineer", "dba", "developer"}
+
+engineers[employee] {
+    employee := data.employees[_]
+    employee.role in engineering_roles
+}
+```
+
+**Prefer**
+```rego
+package policy
+
+import future.keywords.in
+
+engineering_roles = {"engineer", "dba", "developer"}
+
+engineers[employee] {
+    some employee in data.employees
+    employee.role in engineering_roles
+}
+```
+
+## Rationale
+
+Using the `some .. in` construct for iteration removes ambiguity around iteration vs. membership checks, and is
+generally more pleasant to read. Consider the following example:
+
+```rego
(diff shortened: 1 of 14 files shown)


anderseknert:
False positive case fixed. There are more... fundamental issues with our approach, as documented here: https://github.com/StyraInc/regal/issues/264

But it's arguably better if the linter misses a few violations than that it reports violations that aren't. This should be good enough for now.

docs/rules/style/prefer-some-in-iteration.md
```diff
@@ -0,0 +1,117 @@
...
+    some role in attribute.roles
+    role == "admin"
+}
+
+any_user_is_admin if {
+    input.users[_].attributes[_].roles[_] == "admin"
+}
+
+# Using "if", we may even omit the brackets for single line rules
+any_user_is_admin if input.users[_].attributes[_].roles[_] == "admin"
```
srenatus, line 74: It's quite a monster in one line 🙈

anderseknert: Oh, yes 😆

srenatus (approved):

PR #268, opened 2023-08-17
Allows multiple # comments

### LINKED ISSUE #266: no-whitespace-comment should allow ##
It's not uncommon to see comments like:

```rego
##########
# Foobar # 
##########
```
These are currently flagged by the `no-whitespace-comment` rule. We should make an exception for `#` immediately following the first `#`.

Parsifal-M:
Hey Team! 👋 

I hope I understood, and am doing this right!

If not please let me know so I can adjust 😅 

This _**should**_ fix #266

Thanks! 💪


Parsifal-M:
Do I need to run `opa fmt` before pushing? 🤔 

I can do that if needed 👍

anderseknert:
Indeed :)

Thinking about it now, I think we should forbid something like

```rego
##foo
```
too, as it's just as ugly as
```rego
#foo
```
It'd be slightly more complex, but perhaps a regex would work?

Something like:

```rego
_whitespace_comment(text) if regex.match(`^#*\s+.*$`, text)
```

That should allow " comment" and "# comment" and "## comment" and so on, but not "#comment" or "##comment".

I imagine that would be able to replace all of the other `_whitespace_comment` functions, but it would need to be tested :)

anderseknert:
I thought `opa fmt` would strip trailing whitespace from comments, but apparently it doesn't. I'll file an issue in OPA, as I think that should be dealt with there. We can ignore that in this rule for now.

Parsifal-M:
> Indeed :)
> 
> Thinking about it now, I think we should forbid something like
> 
> ```rego
> ##foo
> ```
> 
> too, as it's just as ugly as
> 
> ```rego
> #foo
> ```
> 
> It'd be slightly more complex, but perhaps a regex would work?
> 
> Something like:
> 
> ```rego
> _whitespace_comment(text) if regex.match(`^#*\s+.*$`, text)
> ```
> 
> That should allow " comment" and "# comment" and "## comment" and so on, but not "#comment" or "##comment".
> 
> I imagine that would be able to replace all of the other `_whitespace_comment` functions, but it would need to be tested :)

Something like the below seems to work or at least the tests pass, however, I am no regex expert (does such a thing exist? 😅 )

```
_whitespace_comment(text) {
	regex.match("^(#*)(\\s+.*|$)", text)
}
```

Will check to see if any additional tests are needed and report back 👍

anderseknert:
Nice! Tests are passing, but both your regex and mine seem to fail on `" foo # "` 😅

Parsifal-M:
> Nice! Tests are passing, but both your regex and mine seem to fail on `" foo # "` 😅

Just thinking on this, how would you regex this? Is this considered bad practice in rego?

```rego
package testing

####################
# This is a comment #
####################

default allow := true # We default allow here
```

Just trying to think of how you could regex it 🤔 💭

anderseknert (approved):

anderseknert:
Thanks Peter! 👍 

I got derailed thinking about trailing whitespace even though I had previously said we should fix that in OPA instead. That's done now: https://github.com/open-policy-agent/opa/pull/6163

Parsifal-M:
> Thanks Peter! 👍
> 
> I got derailed thinking about trailing whitespace even though I had previously said we should fix that in OPA instead. That's done now: [open-policy-agent/opa#6163](https://github.com/open-policy-agent/opa/pull/6163)

No worries! 

This is a good way to learn some `rego` and help out at the same time 😄

PR #270, opened 2023-08-21
Enable pre-commit hook for all hook types

jharrisonSV:
The current pre-commit hook configuration only allows the hook to run on a `pre-commit` hook. The default configuration enables the hook for all hook types supported by pre-commit as per the [docs](https://pre-commit.com/#confining-hooks-to-run-at-certain-stages), e.g. on a `push` hook which is very useful for CI.

Also a minor fix for the docs to use a ref that works.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,6 +1,4 @@
 - id: regal-lint
-  stages:
-    - pre-commit
   language: golang
   name: policy file linting
   description: policy file linting with Regal from Styra built from source
@@ -8,8 +6,6 @@
   files: (\.rego)$
 
 - id: regal-lint-use-path
-  stages:
-    - pre-commit
   language: system
   name: policy file linting
   description: policy file linting with Regal from Styra using system $PATH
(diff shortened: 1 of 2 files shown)


c-wygoda:
Other linters we use in this way also on CI (ruff, black in the Python world for example), don't define their stages and leave that to the user.

anderseknert:
Alright, sounds good to me 👍 We'd need to amend the docs though, wouldn't we? And have the example updated to include a stage.

https://github.com/StyraInc/regal/blob/main/docs/pre-commit-hooks.md

anderseknert:
> Users won't need to specify anything. The hook will run at any stage they wish to use.

So the default would be that the linter won't run on any stage unless the user provides one? Or that it runs on all stages unless the user says otherwise?

jharrisonSV:
Runs on all stages unless the user says otherwise.

anderseknert:
OK. Perhaps that's the better default then. We can always change that later if it turns out not to be.

anderseknert (approved):

PR #272, opened 2023-08-23
Filter out files _before_ parsing

### LINKED ISSUE #265: Files ignored may still cause parse errors
This came up when testing Regal against a repo where a few files were used as templates, i.e. something like:
```
package {{org_name}}.authz

# ...
```
Now this is obviously not a valid Rego file, and should be ignored for evaluation. But since we do the ignore check *after* we've parsed the files, this fails even when ignoring the directory or these files.

Doing the ignore check in Rego is likely the best option for rule level ignores, as you won't want to ignore a parse error only for a certain rule, but perhaps we'd need to make the global ignore check in Go before the files are parsed.

Alternatively, we could still leverage the check in Rego, but have Regal do something like:

- Gather all files lined up for evaluation.
- Pass the list for evaluation using the config policy, which...
- Returns the list with any files and directories that should be ignored removed.

This options would be more arduous, but it would benefit from having the same ignore logic for both global ignore directives as well as for rule level ignores.

anderseknert:
Also:
* Allow files to be provided either from file paths, *or* programmatically as modules.
* Marshal config directly to Config struct, and prefer to use the struct rather than passing around the config as a map.

Fixes #265

### DIFF
--- a/internal/io/io.go
+++ b/internal/io/io.go
@@ -72,6 +72,13 @@ func JSONRoundTrip(from any, to any) error {
 	return json.Unmarshal(bs, to) //nolint:wrapcheck
 }
 
+// MustJSONRoundTrip convert any value to JSON and back again, exit on failure.
+func MustJSONRoundTrip(from any, to any) {
+	if err := JSONRoundTrip(from, to); err != nil {
+		log.Fatal(err)
+	}
+}
+
 // MustYAMLToMap creates a map from reader, expecting YAML content, or fail.
 func MustYAMLToMap(from io.Reader) (m map[string]any) {
 	if err := yaml.NewDecoder(from).Decode(&m); err != nil {
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -8,25 +8,27 @@ import (
 	"path/filepath"
 	"strings"
 
+	"gopkg.in/yaml.v3"
+
 	rio "github.com/styrainc/regal/internal/io"
 )
 
 type Config struct {
-	Rules  map[string]Category `json:"rules"`
-	Ignore Ignore              `json:"ignore"`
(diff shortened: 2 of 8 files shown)


internal/io/io.go
```diff
@@ -72,6 +72,23 @@ func JSONRoundTrip(from any, to any) error {
 	return json.Unmarshal(bs, to) //nolint:wrapcheck
 }
 
+// YAMLRoundTrip convert any value to YAML and back again.
+func YAMLRoundTrip(from any, to any) error {
```
kristiansvalland, line 76: While I am not very familiar with (un)marshalling YAML, I do not immediately see how making a roundtrip via YAML as the intermediate is any different from JSON when considering the end result. Or is there some go types that would behave differently in this case?

pkg/config/config.go
```diff
@@ -135,3 +145,45 @@ func (rule *Rule) UnmarshalJSON(data []byte) error {
...
+
+		rule.Ignore = dst
+	}
+
+	delete(result, "enabled")
+
+	rule.Level = level
+	rule.Extra = result
+
+	return nil
```
kristiansvalland, line 188: Probably not a big improvement, but could/should this part be shared with `UnmarshalJSON` to reduce code duplication?

kristiansvalland (approved):
This is a great refactor and fix of the ignore logic. 🙂

I had just one question and one minor suggestion, but those are in no way blocking for this PR and can be ignored if they are wrong or you disagree.

internal/io/io.go
anderseknert: No, you're right. This remained from a point where I considered ditching JSON altogether, but I realized later we need that for output formats. I'll have this removed before merge.

pkg/config/config.go
anderseknert: Fair enough! Done.

PR #273, opened 2023-08-23
Add GitHub output format

### LINKED ISSUE #260: Add GitHub report output format
For when Regal is running in GitHub actions, it'll be useful to have a report format adhering to their [workflow commands](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions).

This would display the issues in-lined in the code view on the location given. All the data necessary is already in the report, so this should be fairly easy to do. Important to include a link to the docs for each rule, as that allows people to learn more about the issues reported.

anderseknert:
This will have violations printed in PRs as annotations in the source code at the location of the violation.

In addition to this, the GitHub output format also prints the summary as a "Job Summary", which is a rather nice touch!

Fixes #260 

**Annotations**
annotation

**Job Summary**
report

### DIFF
--- a/pkg/reporter/reporter.go
+++ b/pkg/reporter/reporter.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"os"
 	"strings"
 
 	"github.com/fatih/color"
@@ -33,6 +34,11 @@ type JSONReporter struct {
 	out io.Writer
 }
 
+// GitHubReporter reports violations in a format suitable for GitHub Actions.
+type GitHubReporter struct {
+	out io.Writer
+}
+
 // NewPrettyReporter creates a new PrettyReporter.
 func NewPrettyReporter(out io.Writer) PrettyReporter {
 	return PrettyReporter{out: out}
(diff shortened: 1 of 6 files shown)


pkg/reporter/reporter.go
```diff
@@ -156,6 +167,77 @@ func (tr JSONReporter) Publish(r report.Report) error {
...
+		summaryFile, err := os.OpenFile(summaryFileLoc, os.O_APPEND|os.O_WRONLY, 0o644)
+		if err != nil {
+			return err //nolint:wrapcheck
+		}
+
+		defer func() {
+			_ = summaryFile.Close()
+		}()
+
+		_, _ = fmt.Fprintf(summaryFile, "### Regal Lint Report\n\n")
```
srenatus, line 206: [nit] Does the linter forbid us to just do this?
```suggestion
		fmt.Fprintf(summaryFile, "### Regal Lint Report\n\n")
```

pkg/reporter/reporter_test.go
```diff
@@ -221,3 +221,46 @@ func TestJSONReporterPublishNoViolations(t *testing.T) {
 		t.Errorf("expected %q, got %q", `{"violations":[]}`, buf.String())
 	}
 }
+
+//nolint:paralleltest
+func TestGitHubReporterPublish(t *testing.T) {
+	// Can't use t.Parallel() here because os.Setenv() forbids that
```
srenatus, line 227: 🤔 os.Setenv forbids it, but does t.Setenv, too?

srenatus (approved):
LGTM

pkg/reporter/reporter_test.go
anderseknert: Oh, yes that it does — the comment was wrong. I'll fix, thanks!

pkg/reporter/reporter.go
anderseknert: I thought it did but it seems not. I'll clean this up!

PR #277, opened 2023-08-29
Fix: Scaffolding now properly populates templatesge nam…

ThomasKingoTM:
The scaffolding "regal new rule" was adding incorrectly formatted package names in both the regal policy and the test. 
This PR should fix the following: 
1. The package name of the regal policy and the test file.
2. The import of the regal policy into the test file.
3. The 'title' attribute in the output of rule.report.


cmd/new.go
```diff
@@ -215,21 +215,27 @@ func templateValues(params newRuleCommandParams) TemplateValues {
...
 
 	if strings.Contains(params.name, "-") {
-		tmplNameTestValue = `["` + params.name + `_test"]`
+		tmplNameTestValue = `["` + params.name + `-test"]`
+	} else if strings.Contains(params.name, "_") {
+		var tempNameTest = strings.ReplaceAll(params.name, "_", "-")
+		tmplNameTestValue = `["` + tempNameTest + `-test"]`
 	} else {
-		tmplNameTestValue = "." + params.name + "_test"
+		tmplNameTestValue = "." + params.name + "-test"
```
anderseknert, line 233: This shouldn't change — the `_test` suffix is idiomatic for test packages.

anderseknert (commented):
Looks good! Just a minor correction required. I'm also not sure what the newly added `.regal` directory is there for?

cmd/new.go
ThomasKingoTM: I ran "regal new rule -t custom -c test -n some_name" a few times, so it might have created it.

ThomasKingoTM: But why doesn't it build correctly?

anderseknert: Some rather strict linter rules for Go as well :P 
Screenshot 2023-08-29 at 23 52 06

anderseknert: > I ran "regal new rule -t custom -c test -n some_name" a few times, so it might have created it.

What I mean is — even if your linte rule would be called "foo-bar-baz", the idiomatic naming for its _test package_ would be "foo-bar-baz_test". So the code should not rewrite that to be "foo-bar-baz-test".

anderseknert: As for the suggested switch statement the linter wants, we could just do this instead:

```go
tmplNameValue = strings.ReplaceAll(params.name, "_", "-")

if strings.Contains(params.name, "-") {
    tmplNameValue = `["` + tempName + `"]`
} else {
    tmplNameValue = "." + params.name
}
```

ThomasKingoTM: > This shouldn't change — the `_test` suffix is idiomatic for test packages.

I believe it depends what we are referring to:
1. If we look at the name of the file itself, you are correct 
2. If we look at the package name of the test file, it should contain a dash, not an underscore.

ThomasKingoTM: As far as I can tell, the two cases I mentioned are both currently handled in the same if-else, which should not be the case, as they are not identical - or am I wrong?

anderseknert: No, while this is obviously opinionated, _test is definitely more common for test packages as it avoids having to quote the name. In our case we almost always quote the names anyway since we use - as a separator, but if you choose a single word as title, like “foo”, then neither foo nor foo_test would need to be quoted. So the _test suffix would be consistent for all test package names.

anderseknert: > As far as I can tell, the two cases I mentioned are both currently handled in the same if-else, which should not be the case, as they are not identical - or am I wrong?

Not sure I follow you here. Can you elaborate?

ThomasKingoTM: I see - I might have missed the memo on the "_test" in package names as well.
I'll make the quick change.

anderseknert: The linter is complaining that we're not using the first assignment, i.e.

```go
// This assignment is never used
tmplNameValue := strings.ReplaceAll(params.name, "_", "-")
```

And it's correct about that. It was my code, so apologies for that! We should of course us it in the contains check:

```go
tmplNameValue = strings.ReplaceAll(params.name, "_", "-")

if strings.Contains(tmplNameValue, "-") {
```

Same thing for the test template name.

ThomasKingoTM: Should be fixed now. Still doesn't build - a ton of "Cannot open: file exists" errors.

ThomasKingoTM: Finally works - had to decuddle assignments (their linter sure is rough to get by) :)

anderseknert: Hehe, sure is! Yeah, those warnings are annoying but nothing to worry about. Some bug in the Go linter as I understood it. 

I'd suggest getting `golangcilint` yourself so that you can run `golangci-lint run ./...` yourself before pushing your commits.

anderseknert (approved):

anderseknert:
Thanks!

PR #278, opened 2023-08-30
Ensure rule "name" logic works without rule.head.name

### LINKED ISSUE #276: Ref-head rules aren't currently handled by Regal
Running Regal against some modern Rego where ref-head rules are used, I'm seeing some false positives. This isn't particularly strange considering we've not really had those in mind before, but that's definitely something we should do, and I'd suggest going through all rules where this would be relevant and add test cases to assert these rules are dealt with correctly. Example:

**p.rego**
```rego
package p

import future.keywords.if

a["b"] if input.c
```

```shell
$ regal lint p.rego
Rule:         	use-assignment-operator
Description:  	Prefer := over = for assignment
Category:     	style
Location:     	p.rego:5:1
Text:         	a["b"] if input.c
Documentation:	https://docs.styra.com/regal/rules/style/use-assignment-operator

1 file linted. 1 violation found.
```

This is obviously not right, as there's no explicit assignment happening here. I'm sure there are more false positives in other rules, so let's try to fix this.

anderseknert:
The name attribute is not present in ref-head rules, so make sure things keep working when those are around.

Fixes #276

Note though that the rules don't yet necessarily take all ref names into account. Will look at that separately. Just fix the bug for now.

### DIFF
--- a/bundle/regal/ast.rego
+++ b/bundle/regal/ast.rego
@@ -18,13 +18,23 @@ package_name := concat(".", [path.value |
 	i > 0
 ])
 
+# METADATA
+# description: |
+#   returns the name of the rule, i.e. "foo" for "foo { ... }"
+#   note that what constitutes the "name" of a ref-head rule is
+#   ambiguous at best, i.e. a["b"][c] { ... } ... in those cases
+#   we currently return the first element of the ref, i.e. "a"
+name(rule) := rule.head.name if {
+	rule.head.name
+} else := rule.head.ref[0].value
+
 tests := [rule |
 	some rule in input.rules
 	not rule.head.args
-	startswith(rule.head.name, "test_")
+	startswith(name(rule), "test_")
 ]
 
-rule_names := {rule.head.name |
+rule_names := {name(rule) |
 	some rule in input.rules
 	not rule.head.args
 }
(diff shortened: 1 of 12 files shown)


bundle/regal/ast.rego
```diff
@@ -297,6 +307,23 @@ implicit_boolean_assignment(rule) if {
...
+	# `a.b if true`, which is "rewritten" to `a.b = true` *and*  where a location is still added to the value
+	# see https://github.com/open-policy-agent/opa/issues/6184 for details
+	#
+	# Do note that technically, it is possible to write a rule where the `true` value actually is on column 1, i.e.
+	#
+	# a.b =
+	# true
+	# if true
+	#
+	# If you write Rego like that — you're not going to use Regal anyway, are you? ¯\_(ツ)_/¯
```
srenatus, line 321: 😄

bundle/regal/rules/style/prefer_snake_case.rego
```diff
@@ -23,3 +27,13 @@ report contains violation if {
 
 	violation := result.fail(rego.metadata.chain(), result.location(var))
 }
+
+_is_name(ref, pos) if {
+	pos == 0
+	ref.type == "var"
+}
```
srenatus, line 34: [golf]
```suggestion
_is_name(ref, 0) if ref.type == "var"
```

srenatus (approved):

bundle/regal/rules/style/prefer_snake_case.rego
anderseknert: I like it :) And in this case it won't matter, but I remember how "pattern matching" on function args was noticeably slower when done in the hot path of `walk` traversals.

PR #279, opened 2023-08-30
bundle: rearrange embed.FS setup

srenatus:
No functional changes, I hope!

### DIFF
--- a/bundle/bundle.go
+++ b/bundle/bundle.go
@@ -1,0 +1,11 @@
+package bundle
+
+import (
+	"embed"
+)
+
+// Bundle FS will include the tests as well, but since that has negligible impact on the size of the binary,
+// it's preferable to filter them out from the bundle than to e.g. create a separate directory for tests
+//
+//go:embed *
+var Bundle embed.FS
--- a/internal/io/io.go
+++ b/internal/io/io.go
@@ -94,7 +94,7 @@ func CloseFileIgnore(file *os.File) {
 }
 
 func ExcludeTestFilter() filter.LoaderFilter {
-	return func(abspath string, info files.FileInfo, depth int) bool {
+	return func(_ string, info files.FileInfo, _ int) bool {
 		return strings.HasSuffix(info.Name(), "_test.rego") &&
 			// (anderseknert): This is an outlier, but not sure we need something
 			// more polished to deal with this for the time being.
--- a/cmd/lint.go
+++ b/cmd/lint.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/fs"
 	"log"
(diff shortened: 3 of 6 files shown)


bundle/bundle.go
```diff
@@ -0,0 +1,11 @@
+package bundle
+
+import (
+	"embed"
+)
+
+// Note: this will bundle the tests as well, but since that has negligible impact on the size of the binary,
+// it's preferable to filter them out from the bundle than to e.g. create a separate directory for tests
+//
+//go:embed *
```
srenatus, line 10: 💭 Probably means that _this_ file makes it into the bundle, too. But it should be ignored...?

anderseknert (approved):
That's an interesting approach, and not one I had considered :)
Having a `.go` file in the bundle may not be koscher, but yeah, it should just get ignored.

bundle/bundle.go
anderseknert: https://pkg.go.dev/embed#hdr-Directives

> If a pattern names a directory, all files in the subtree rooted at that directory are embedded (recursively), except that files with names beginning with ‘.’ or ‘_’ are excluded. 

Does that mean the file could be named `_bundle.go`, or would that not work with Go? 🤔

bundle/bundle.go
```diff
@@ -0,0 +1,11 @@
...
+var Bundle embed.FS
```
anderseknert, line 11: >  Error: ST1022: comment on exported var Bundle should be of the form "Bundle ..." (stylecheck)

Adding an ignore on package level should be fine here.

internal/io/io.go
```diff
@@ -94,7 +94,7 @@ func CloseFileIgnore(file *os.File) {
 }
 
 func ExcludeTestFilter() filter.LoaderFilter {
-	return func(abspath string, info files.FileInfo, depth int) bool {
+	return func(_ string, info files.FileInfo, _ int) bool {
```
srenatus, line 97: Sorry, unrelated change. I had just happened to look into how the bundle is loaded...

bundle/bundle.go
srenatus: I was hoping that `*` explicitly included files starting with `.` -- we need it for `.manifest`.

srenatus:
Tests fail, back to draft. I'm looking into it.

bundle/bundle.go
anderseknert: Oh right, you'll need to use `all` for that, I think

srenatus: ✔️ Double-checked: `.manifest` is there, and `bundle.go` is, too.

anderseknert: Yay!

PR #283, opened 2023-08-31
Rule: file-length

anderseknert:
Add a rule to limit the length of file to a configured value, where the default is 500 lines.

### DIFF
--- a/bundle/regal/rules/style/file_length.rego
+++ b/bundle/regal/rules/style/file_length.rego
@@ -1,0 +1,19 @@
+# METADATA
+# description: Max file length exceeded
+package regal.rules.style["file-length"]
+
+import future.keywords.contains
+import future.keywords.if
+import future.keywords.in
+
+import data.regal.ast
+import data.regal.config
+import data.regal.result
+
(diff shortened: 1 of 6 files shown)


bundle/regal/rules/style/file_length.rego
```diff
@@ -0,0 +1,19 @@
...
+import data.regal.ast
+import data.regal.config
+import data.regal.result
+
+cfg := config.for_rule("style", "file-length")
+
+report contains violation if {
+	count(input.regal.file.lines) > cfg["max-file-length"]
+
+	violation := result.fail(rego.metadata.chain(), result.location(input["package"]))
```
srenatus, line 18: [style] I guess it's just me but I'd default to non-`["..."]` variant where possible: `input.package`

srenatus (approved):

bundle/regal/rules/style/file_length.rego
anderseknert: You can’t, as package is a keyword :)

srenatus: aaaah ignore me. 🤦

PR #286, opened 2023-09-04
test: allow running "regal" on its own bundle 

srenatus:
...by rewriting some refs on loading: data.regal becomes data.internal.regal,
both in package definitions and imports, e.g.

    package regal.main
    import data.regal.config

becomes

    package regal.internal.main
    import data.internal.regal.config

**Update** ☝️ This didn't work. Custom rules might use `data.regal` in their code, and direct refs would have to be rewritten too (so `data.regal.config.whatever` without an import).

⏩ We're now using a ModuleLoader to load packages from the embedded Regal module into `data.regal` as needed. This allows for running all of

- `regal test bundle` # all of it
- `regal test bundle/regal/main_test.rego` # individual test files

### DIFF
--- a/cmd/test.go
+++ b/cmd/test.go
@@ -109,13 +109,12 @@ func opaTest(args []string) int {
 		Ignore: testParams.ignore,
 	}
 
-	var modules map[string]*ast.Module
-
-	var bundles map[string]*bundle.Bundle
-
-	var store storage.Store
-
-	var err error
+	var (
+		modules map[string]*ast.Module
+		bundles map[string]*bundle.Bundle
+		store   storage.Store
+		err     error
+	)
 
 	if testParams.bundleMode {
 		bundles, err = tester.LoadBundles(args, filter.Apply)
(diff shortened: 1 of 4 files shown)


cmd/test.go
```diff
@@ -109,13 +109,12 @@ func opaTest(args []string) int {
...
-
-	var store storage.Store
-
-	var err error
+	var (
+		modules map[string]*ast.Module
+		bundles map[string]*bundle.Bundle
+		store   storage.Store
+		err     error
+	)
```
srenatus, line 117: I can be persuaded to revert this chunk, I just find the linter rule enforcing newlines between this horrendous 🙃

anderseknert: Yeah, I don't really like it for vars either, and perhaps that's configurable. I do think it's pretty good for other things though, like separating return values, and so on.

anderseknert: But this is fine with me!

e2e/cli_test.go
```diff
@@ -318,6 +318,32 @@ func TestLintRuleNamingConventionFromCustomCategory(t *testing.T) {
 	}
 }
```
srenatus, line 320: It's a bit simplistic, but I've verified that it fails on main.

anderseknert (approved):
This is GREAT! Thanks @srenatus 😃

PR #287, opened 2023-09-04
Add `--debug` flag to `regal lint` command

anderseknert:
More capabilities to be added in time, but for now:

* Disable `print` output unless the debug flag is provided
* Print the location of the user configuration file found
* Print the final configuration merged from provided and user conf
* Fix bug that made configuration be parsed and merged twice

### DIFF
--- a/bundle/regal/rules/style/file_length.rego
+++ b/bundle/regal/rules/style/file_length.rego
@@ -6,7 +6,6 @@ import future.keywords.contains
 import future.keywords.if
 import future.keywords.in
 
-import data.regal.ast
 import data.regal.config
 import data.regal.result
 
--- a/bundle/regal/rules/style/file_length_test.rego
+++ b/bundle/regal/rules/style/file_length_test.rego
@@ -3,7 +3,6 @@ package regal.rules.style["file-length_test"]
 import future.keywords.if
 import future.keywords.in
 
-import data.regal.ast
 import data.regal.config
 
 import data.regal.rules.style["file-length"] as rule
--- a/cmd/lint.go
+++ b/cmd/lint.go
@@ -31,6 +31,7 @@ type lintCommandParams struct {
 	failLevel       string
 	rules           repeatedStringFlag
 	noColor         bool
+	debug           bool
 	disable         repeatedStringFlag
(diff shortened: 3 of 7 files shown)


srenatus (approved):
Looks good to me. Would look even better with an e2e test, though 😉

anderseknert:
@srenatus True :) I'll add an e2e test.

anderseknert:
Ugh! Good call... 🤦 

```
found user config file: e2e/testdata/configs/ignore_files_prefer_snake_case.yaml
panic: cannot have key "ignore" in inlined map: conflicts with struct field [recovered]
	panic: cannot have key "ignore" in inlined map: conflicts with struct field

goroutine 1 [running]:
gopkg.in/yaml%2ev3.handleErr(0x14000539f08)
	/Users/anderseknert/go/pkg/mod/gopkg.in/yaml.v3@v3.0.1/yaml.go:294 +0x80
panic({0x10331aec0?, 0x1400029f1c0?})
	/opt/homebrew/Cellar/go/1.21.0/libexec/src/runtime/panic.go:914 +0x218
gopkg.in/yaml%2ev3.(*encoder).structv.func1()
	/Users/anderseknert/go/pkg/mod/gopkg.in/yaml.v3@v3.0.1/encode.go:245 +0x398
gopkg.in/yaml%2ev3.(*encoder).mappingv(0x1400046e800, {0x0?, 0x0}, 0x14000539118)
	/Users/anderseknert/go/pkg/mod/gopkg.in/yaml.v3@v3.0.1/encode.go:265 +0xfc
gopkg.in/yaml%2ev3.(*encoder).structv(0x1400046e800, {0x0, 0x0}, {0x1033e2b00?, 0x1400028fbf0?, 0x102a80dbc?})
```

Not sure how to best serialize the extra attributes without having a key named "extra" in the YAML output then. Back to the drawing board.

anderseknert:
Huh, so the custom marshaller was never invoked due to it being a poiinter receiver, and apparently it shouldn't have been. Marshalling now works as intended, and I added another test to make sure it continues to do so.

srenatus (approved):

PR #289, opened 2023-09-05
Allow not providing level in user configuration

### LINKED ISSUE #285: Overriding rule configuration requires providing level attribute
While I personally prefer explicit configuration, it should be possible to override only attributes of interest, and inherit the `level` attribute from the Regal-provided configuration.
```yaml
rules:
  testing:
    metasyntactic-variable:
      ignore:
        files:
          - "*test.rego"
```

But this configuration will fail with:
```
failed to decode user config from /project/.regal: value of 'level' must be string
```
And that's a bug.

anderseknert:
Previously, this would lead to the rule level getting set to "" (empty string), which "worked" as it would still print the violation, but since it wasn't set to `error` it would not change the exit code of the lint command.

If the level is not provided in user configuration now, we simply inherit the value from the provided configuration.

Fixes #285

### DIFF
--- a/bundle/regal/config/config.rego
+++ b/bundle/regal/config/config.rego
@@ -14,7 +14,29 @@ docs["resolve_url"](url, category) := replace(
 
 user_config := data.regal_user_config
 
-merged_config := object.union(data.regal.config.provided, user_config)
+config_union := object.union(data.regal.config.provided, user_config)
+
+merged_config["ignore"] := config_union.ignore
+
+merged_config["rules"] := merged_rules
+
+# Iterate over the unioned rules and check for any level set to "". This means the user has not provided
+# the level for this rule, so we should fall back on the provided level, i.e. that which was
+merged_rules[category] := rules if {
+	some category, _rules in config_union.rules
+	rules := {rule_name: empty_level_to_provided(category, rule_name, rule_conf) |
+		some rule_name, rule_conf in _rules
+	}
+}
+
+empty_level_to_provided(_, _, conf) := conf if conf.level != ""
+
+empty_level_to_provided(category, title, conf) := object.union(
+	conf,
+	{"level": data.regal.config.provided.rules[category][title].level},
+) if {
+	conf.level == ""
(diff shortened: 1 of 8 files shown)


e2e/cli_test.go
```diff
@@ -447,6 +447,23 @@ func TestCreateNewBuiltinRuleFromTemplate(t *testing.T) {
...
+func TestMergeRuleConfigWithoutLevel(t *testing.T) {
+	t.Parallel()
+
+	stdout := bytes.Buffer{}
+	stderr := bytes.Buffer{}
+
+	cwd := must(os.Getwd)
+
+	// No violations from the built-in configuration in the policy provided, but
+	// the user --config-file changes the nax-file-length to 1, so this should fail
```
charlieegan3, line 459:
```suggestion
	// the user --config-file changes the max-file-length to 1, so this should fail
```

bundle/regal/config/config.rego
```diff
@@ -14,7 +14,27 @@ docs["resolve_url"](url, category) := replace(
...
+merged_rules[category] := rules if {
+	some category, _rules in config_union.rules
+	rules := {rule_name: empty_level_to_provided(category, rule_name, rule_conf) |
+		some rule_name, rule_conf in _rules
+	}
+}
+
+empty_level_to_provided(_, _, conf) := conf if conf.level != ""
+
+empty_level_to_provided(category, title, conf) := object.union(
```
charlieegan3, line 32: I spent a little while thinking about this, I wonder if an explanatory comment might be handy for the future since the level is being processed as a special case.

charlieegan3 (approved):
Looks good to me 👍

e2e/cli_test.go
anderseknert: Thanks! Fixing.

bundle/regal/config/config.rego
charlieegan3: Nah, I'm just having a panic at attack at how slow I am to read Rego, still! 😅

Issue #220, opened 2023-08-02
Dubious use of print + sprintf


anderseknert:
**Avoid**
```rego
allow {
    # if any of input.name or input.domain are undefined, this will just print <undefined>
    print(sprintf("name is: %s domain is: %s", [input.name, input.domain]))

    input.name == "admin"
}
```

**Prefer**
```rego
allow {
    # if any of input.name or input.domain are undefined, this will still print the whole
    # sentence, with the value undefined printed as such, e.g.
    # name is: admin domain is: <undefined>
    print("name is:", input.name, "domain is:", input.domain)

    input.name == "admin"
}
```
Since `print` allows any number of arguments, there's rarely any benefit to using `sprintf` for formatting. But more importantly, the `print` function is unique in that it will allow any arguments passed to be undefined without terminating, but will print such values as `<undefined>`. Using `sprintf` will however nullify this benefit, and just print `<undefined>` without the context.

It should be added to the notes on this that the use of `print` is discouraged outside of development, and other rules exists to check for its use. But in the context of development, one may of course choose to disable such rules, or have them ignored for e.g. `_test.rego` files... so I think having rules on how to make best use of `print` still makes sense.

Ronnie-personal:
@anderseknert 
I'm getting started on this issue.

I run following commands to create the skeleton code in my local dev environment. 

`go build`   
`./regal new rule --category testing --name dubious-print-sprintf`

Two files are created under '.regal/rules/testing'.
dubious_print_sprintf.rego  dubious_print_sprintf_test.rego

Did I miss anything? The rego files are not be created under bundle directory.

anderseknert:
That's great to hear! The `regal new rule` command assumes most people using it will do so for custom rules, so by default the `--type` is set to `custom`. Change this to `--type builtin` and the files should end up in the right location.

Issue #231, opened 2023-08-08
Add docs and config when running `regal new rule --type builtin`


anderseknert:
We currently add only the rule and a test policy. Adding an entry to the [provided configuration](https://github.com/StyraInc/regal/blob/main/bundle/regal/config/provided/data.yaml), as well as a new page for documentation in that category would be a nice addition.

Ronnie-personal:
I have started to work on the issue.

PR #291, opened 2023-09-07
print sprintf testing rule

Ronnie-personal:
Adding dubious-print-sprintf under testing category.

This PR is to address #220

### DIFF
--- a/bundle/regal/rules/testing/dubious_print_sprintf.rego
+++ b/bundle/regal/rules/testing/dubious_print_sprintf.rego
@@ -1,0 +1,20 @@
+# METADATA
+# description: dubious print sprintf
+package regal.rules.testing["dubious-print-sprintf"]
+
+import future.keywords.contains
+import future.keywords.if
+import future.keywords.in
+
+import data.regal.ast
+import data.regal.result
+
+report contains violation if {
+	some rule in input.rules
+	some rule_body in rule.body
+	rule_body.terms[0].value[0].value == "print"
+    rule_body.terms[1].type == "call" 
+	rule_body.terms[1].value[0].value[0].value == "sprintf" 
+
+	violation := result.fail(rego.metadata.chain(), result.location(rule_body.terms[1].value[0].value[0]))
+}
--- a/bundle/regal/rules/testing/dubious_print_sprintf.rego
+++ b/bundle/regal/rules/testing/dubious_print_sprintf.rego
@@ -1,5 +1,5 @@
 # METADATA
-# description: dubious print sprintf
+# description: dubious use of print and sprintf
 package regal.rules.testing["dubious-print-sprintf"]
 
 import future.keywords.contains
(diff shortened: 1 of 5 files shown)


Ronnie-personal:
@anderseknert 
I guess I still need to work on readme file and data.yaml.

Does the code changes overall make sense to you?

Also I'm not sure why 'config_test.rego' failed, that's not part of this PR code change.

Thanks!

anderseknert:
The `config_test.rego` test checks that all rules have an entry in the `data.yaml` file, which like you said you still need to to :) There's an [open issue](https://github.com/StyraInc/regal/issues/231) to have the `regal new rule` command add this automatically, along with a page for documentation of the rule. If you're up for it later, that'd be a good one to work on!

Other than that, the code looks good to me! The way it currently works however means it'll only find `print` statements in the top level of a rule — i.e. if you add a test for this rule, you'll see that it won't fail:

```rego
y {
    comprehension := [x |
        x := input[_]
        print(sprintf("x is: %s", [x]))
    ]
}
```

To fix this, you could use the `walk` built-in function to traverse the whole rule. See the code for [non-raw-regex-pattern](https://github.com/StyraInc/regal/blob/main/bundle/regal/rules/idiomatic/non_raw_regex_pattern.rego) for an example, but basically this:

```rego
	# regal ignore:unused-return-value,function-arg-return
	walk(input.rules, [_, value])

	value[0].type == "ref"
	value[0].value[0].type == "var"
	value[0].value[0].value == "print"

        # the rest of your code here
}
```

Traversing everything like this is quite slow, and I plan to investigate alternative approaches in the future. But let's go with that for now as at least it will be correct :)

For the docs, I think the description in the issue would work well: https://github.com/StyraInc/regal/issues/220

bundle/regal/rules/testing/dubious_print_sprintf.rego
```diff
@@ -0,0 +1,20 @@
+# METADATA
+# description: dubious print sprintf
```
anderseknert, line 2: Let's change this to "Dubious use of `print` and `sprintf`"

Ronnie-personal: Thanks for the comment. I will fix it.

bundle/regal/rules/testing/dubious_print_sprintf.rego
```diff
@@ -0,0 +1,24 @@
+# METADATA
+# description: dubious use of print and sprintf
```
anderseknert, line 2: dubious -> Dubious

bundle/regal/rules/testing/dubious_print_sprintf_test.rego
```diff
@@ -0,0 +1,55 @@
...
+			"ref": config.docs.resolve_url("$baseUrl/$category/dubious-print-sprintf", "testing"),
+		}],
+		"title": "dubious-print-sprintf",
+	}}
+}
+
+test_fail_bodies_print_sprintf if {
+	test_policy := ast.policy(`
+y {
+comprehension := [x |
```
anderseknert, line 37: Please indent the body of this comprehension

e2e/testdata/violations/most_violations.rego
```diff
@@ -114,3 +114,8 @@ use_some_for_output_vars {
 
 # metasyntactic variable
 foo := "bar"
+
+# dubious print sprintf
+y {
+print(sprintf("name is: %s domain is: %s", [input.name, input.domain]))
```
anderseknert, line 120: Indent

anderseknert (commented):
Looks good! Just some small fixes and then we can have this merged 👍

bundle/regal/rules/testing/dubious_print_sprintf_test.rego
Ronnie-personal: Sure, will do.

anderseknert (approved):
This is looking really good! There's still the docs to have added, but I can take care of that in a follow-up fix.

Thanks for contributin to Regal! 👏

PR #295, opened 2023-09-07
Rule: chained-rule-body

### LINKED ISSUE #292: Chained rule body
From the [OPA docs](https://www.openpolicyagent.org/docs/latest/policy-language/#incremental-definitions).

> If the head of the rule is same, we can chain multiple rule bodies together to obtain the same result. We don’t recommend using this form anymore.

```rego
instances contains instance if {
    server := sites[_].servers[_]
    instance := {"address": server.hostname, "name": server.name}
} {
    container := containers[_]
    instance := {"address": container.ipaddress, "name": container.name}
}
```

I don't see this used very often, but it'd be good to add as a style category rule.

anderseknert:
While `opa-fmt` fixes this, this is peculiar enough that the rule docs alone make for an informative source to point at.

Also some minor improvements here and there:
* Fix other rules that would report false positives when chained rules bodies were present
* Remove a trailing quote in the builtin test template used by `rego new rule`
* Add some guiding principles for new rules in development.md

Fixes #292

### DIFF
--- a/bundle/regal/rules/testing/metasyntactic_variable.rego
+++ b/bundle/regal/rules/testing/metasyntactic_variable.rego
@@ -32,6 +32,14 @@ report contains violation if {
 
 	lower(ref.value) in metasyntactic
 
+	# In case we have chained rule bodies — only flag the location where we have an actual name:
+	# foo {
+	#    input.x
+	# } {
+	#    input.y
+	# }
+	not ast.is_chained_rule_body(rule, input.regal.file.lines)
+
 	violation := result.fail(rego.metadata.chain(), result.location(location_of(ref, rule)))
 }
 
--- a/docs/development.md
+++ b/docs/development.md
@@ -19,6 +19,18 @@ for testing it. The code here will be a pretty simple template, but contains all
 rule. A good idea for learning more about what's needed is to take a look at some previous PRs adding new rules to
 Regal.
 
+### Guiding principles for new built-in rules
+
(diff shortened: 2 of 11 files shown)


bundle/regal/rules/testing/metasyntactic_variable.rego
```diff
@@ -32,6 +32,8 @@ report contains violation if {
 
 	lower(ref.value) in metasyntactic
 
+	not ast.is_chained_rule_body(rule, input.regal.file.lines)
```
charlieegan3, line 35: I feel like some of these early ejects triggered by other rules are worthy of explanatory comments.

anderseknert: I wouldn't say it's triggered by another _rule_, just that in this case if you say:

```rego
foo {
    true
} {
    false
}
```
There is only really one "foo" to point at. But I'll add a comment to explain this in the code 👍

docs/development.md
```diff
@@ -16,6 +16,18 @@ for testing it. The code here will be a pretty simple template, but contains all
 rule. A good idea for learning more about what's needed is to take a look at some previous PRs adding new rules to
 Regal.
 
+### Guiding principles for new built-in rules
+
+- All rules should have unique names, even across categories
```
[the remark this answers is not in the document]
anderseknert: Thanks! Fixing.

PR #299, opened 2023-09-07
Except walk from function-arg-return

### LINKED ISSUE #298: Add default exception for `walk` in `function-arg-return`
The `walk` built-in function is a special one, as it's the only one producing a relation. Therefore, it is OK that it is treated as one even in style, and:
```rego
walk(object, [path, value])
```
is arguably more idiomatic than
```rego
[path, value] := walk(object)
```

We should simply add `walk` to the `except-functions` configuration option for this rule in the provided configuration, and have all comment directives where we ignore this rule in our code removed.

anderseknert:
Fixes #298

### DIFF
--- a/bundle/regal/ast.rego
+++ b/bundle/regal/ast.rego
@@ -73,7 +73,6 @@ import future.keywords.in
 ]))
 
 _find_nested_vars(obj) := [value |
-	# regal ignore:function-arg-return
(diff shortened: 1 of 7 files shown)


anderseknert:
Merging this as it's small in scope and I depend on this for #301

Issue #282, opened 2023-08-31
Allow a broader scope for custom rules


sesponda:
Currently, the evaluation context for each custom rule is per-file (i.e. you get the AST for a single file).

We'd like to code a custom rule in the form of "_if your package is name `alfa`, then your code must have a rule named `export_rule`_"  (more context on _why_ we do this below at the end).

With a per-file scope is not possible to encode this, as two files can contribute code to the same package and trigger a false positive. Example:

File`example1.rego`:

``` 
package alfa

export_rule = x { .... } 
.. other code...

```

File `example2.rego`:

```
package alfa

.. other code...
```

In this situation, `example2.rego` fails the control. But, if both files are part of the same bundle, then the rule is not violated. For the time being, the workaround is asking Rego developers to ensure each package's code is not scattered in multiple files.

_More context about this:_
- We are building a validation system where some teams consume policy logic encoded by other teams. 
- To support standard policy aggregation and discovery patterns, have a guidelines in terms of package names and rules that packages should always have available. 
- We enforce this in runtime and unit tests. Having a linter rule will allow to converge in Regal.

anderseknert:
Hi @sesponda! 

And thanks for filing this. It's definitely been something I'd want to do, so I guess this is a good catalyst to get started :) As for the design, I can think of two rather different approaches here.

The first one entails sending **all** parsed policy as input for evaluation, and let the policy author deal with that massive bunch of ASTs as they see fit. This would be the most flexible choice from the POV of the policy author, as you have access to all information you could possibly need from inside policy evaluation. However, this would likely perform very poorly in large repos, and the design would not easily lend itself to parallel evaluation like how we currently do in Regal.

The second option, and this is the one I've been thinking the most about, would be to keep the scope limited to a single file, but allow a new and different _aggregate report_ to be returned from linter rules. These would not work as violations as they'd merely collect some data of interest and send it back to Regal. Once the "normal" linter rules have been evaluated, the aggregate data collected from these rules would be used as input for evaluating the aggregates.

A linter rule collecting data for an aggregate could look (very roughly!) look something like this:

```rego
# METADATA
# description: export_rule required in package alfa
package example

import future.keywords

# METADATA
# custom:
#   aggregate_type: any
aggregate contains entry if {
    path := input["package"].path

    count(path) == 2
    path[1].value == "alfa"

    some rule in input.rules
    rule.head.name == "export_rule"

    entry := report.aggregate(rego.metadata.chain(), {
[...]

PR #306, opened 2023-09-11
Introduce broader scope for custom rules

sesponda:
This (preliminary) addresses https://github.com/StyraInc/regal/issues/282

I'd recommend reviewing it commit by commit.

1) First commit: introduced the change in the simplest possible way. Unit and e2e tests still green.
2) Second change: the added function (`collectAggregates`) duplicates the file iteration logic happening goroutines in `lintWithRegoRules`. The common logic is extracted into a new third function, `evalAndCollect`. This makes the diff a little messy, being the reason why it's in another comment.

Next actions: if this implementation is OK, I'll be happy to add more commits with unit tests and documentation updates (some expectations/guidance welcome)

### DIFF
--- a/pkg/linter/linter.go
+++ b/pkg/linter/linter.go
@@ -426,36 +426,20 @@ func (l Linter) prepareRegoArgs(query ast.Body) []func(*rego.Rego) {
 }
 
 func (l Linter) lintWithRegoRules(ctx context.Context, input rules.Input, aggregates []report.Aggregate) (report.Report, error) {
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
+	aggregate := report.Report{}
 
 	regoArgs := l.prepareRegoArgs(query)
-
 	linterQuery, err := rego.New(regoArgs...).PrepareForEval(ctx)
 	if err != nil {
 		return report.Report{}, fmt.Errorf("failed preparing query for linting: %w", err)
 	}
 
-	aggregate := report.Report{}
-
-	var wg sync.WaitGroup
-
-	var mu sync.Mutex
-
-	errCh := make(chan error)
-	doneCh := make(chan bool)
-
-	for _, name := range input.FileNames {
-		wg.Add(1)
-
-		go func(name string) {
-			defer wg.Done()
-
+	err = l.evalAndCollect(ctx, input, linterQuery,
+		//input builder for each file eval
+		func(name string, content string, module *ast.Module) (map[string]any, error) {
 			enhancedAST, err := parse.EnhanceAST(name, input.FileContent[name], input.Modules[name])
 			if err != nil {
-				errCh <- fmt.Errorf("failed preparing AST: %w", err)
-
-				return
+				return nil, err
 			}
 			if aggregates == nil {
 				// if the value is nil, inserting it "as is" will make it null-defined, and
@@ -467,40 +451,18 @@ func (l Linter) lintWithRegoRules(ctx context.Context, input rules.Input, aggreg
 			} else {
 				enhancedAST["aggregate"] = aggregates
 			}
-			resultSet, err := linterQuery.Eval(ctx, rego.EvalInput(enhancedAST))
-			if err != nil {
-				errCh <- fmt.Errorf("error encountered in query evaluation %w", err)
-
-				return
-			}
-
-			result, err := resultSetToReport(resultSet)
-			if err != nil {
-				errCh <- fmt.Errorf("failed to convert result set to report: %w", err)
-
-				return
-			}
-
-			mu.Lock()
-			aggregate.Violations = append(aggregate.Violations, result.Violations...)
-			aggregate.Aggregates = append(aggregate.Aggregates, result.Aggregates...)
-			mu.Unlock()
-		}(name)
+			return enhancedAST, nil
+		},
+		// result collector
+		func(report report.Report) {
+			aggregate.Violations = append(aggregate.Violations, report.Violations...)
+			aggregate.Aggregates = append(aggregate.Aggregates, report.Aggregates...)
+		},
+	)
+	if err != nil {
+		return report.Report{}, err
 	}
-
-	go func() {
-		wg.Wait()
-		doneCh <- true
-	}()
-
-	select {
-	case <-ctx.Done():
-		return report.Report{}, fmt.Errorf("context cancelled: %w", ctx.Err())
-	case err := <-errCh:
-		return report.Report{}, fmt.Errorf("error encountered in rule evaluation %w", err)
-	case <-doneCh:
-		return aggregate, nil
-	}
+	return aggregate, nil
 }
 
 func resultSetToReport(resultSet rego.ResultSet) (report.Report, error) {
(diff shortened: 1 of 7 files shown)


pkg/linter/linter.go
```diff
@@ -369,7 +374,7 @@ func (l Linter) paramsToRulesConfig() map[string]any {
 	}
 }
 
-func (l Linter) prepareRegoArgs() []func(*rego.Rego) {
+func (l Linter) prepareRegoArgs(query ast.Body) []func(*rego.Rego) {
```
anderseknert, line 377: This looks unused?

pkg/linter/linter.go
```diff
@@ -676,3 +662,103 @@ func (l Linter) getBundleByName(name string) (*bundle.Bundle, error) {
...
+
+	return result, nil
+}
+
+// Process each file in input.Filenames in a goroutine, with the given Rego query and building the eval input using the
+// provided function. Collects the results via the provided collector. The collector is guaranteed to
+// run sequentially via a mutex.
+func (l Linter) evalAndCollect(ctx context.Context, input rules.Input, query rego.PreparedEvalQuery,
+	queryInputBuilder func(name string, content string, module *ast.Module) (map[string]any, error),
+	reportCollector func(report report.Report),
```
anderseknert, line 705: Could we perhaps define types for these callback functions to help keep the args list readable?

pkg/linter/linter.go
```diff
@@ -206,7 +206,12 @@ func (l Linter) Lint(ctx context.Context) (report.Report, error) {
 
 	aggregate.Violations = append(aggregate.Violations, goReport.Violations...)
 
-	regoReport, err := l.lintWithRegoRules(ctx, input)
+	aggregates, err := l.collectAggregates(ctx, input)
```
anderseknert, line 209: I think guarding this with a:

```go
	if len(input.Modules) > 1 {
		// ...
	}
```
Would be enough to avoid having these rules called when only a single file is to be linted.

anderseknert (commented):
Looks like a great start! Some tests would help me understand exactly how this behaves, but I reckon that'll be the next step. Thanks a lot for looking into this! 🚀

pkg/linter/linter.go
```diff
@@ -206,7 +206,12 @@ func (l Linter) Lint(ctx context.Context) (report.Report, error) {
...
+	if err != nil {
+		return report.Report{}, fmt.Errorf("failed to collect aggregates using Rego rules: %w", err)
+	}
+
+	regoReport, err := l.lintWithRegoRules(ctx, input, aggregates)
```
charlieegan3, line 214: question: I'm missing something I think. What I see is that the aggregates rules are collected (L209) before being passed to `l.lintWithRegoRules`. This in turn passes them to `evalAndCollect`. These are then all collected onto the report.

What wasn't clear to me in this implementation was when aggregate rules were evaluated on the results of other rules.

In my head, I can't see where step 4 from https://github.com/StyraInc/regal/issues/282#issuecomment-1708528606 is being done.

What I'm not sure about is if I've missed this, or if it's coming in another commit! 😅

charlieegan3 (commented):
Thanks for having a look into this, I think it's going to be a great feature.

pkg/linter/linter.go
sesponda: It's used by `lintWithRegoRules` (where query is `violations = data.regal.main.report`) and by `collectAggregates` (where query is `aggregates = data.regal.main.aggregate`)

pkg/linter/linter.go
sesponda: > What wasn't clear to me in this implementation was when aggregate rules were evaluated on the results of other rules.

They are evaluated only if a violation rule reads the aggregates. This is the file I'm using for my local testing:

```
# METADATA
# description: Could not find a rule satisfying query pattern (some_package).public.export
package custom.regal.rules.acmecorp["require-export-rule"]

import future.keywords
import data.regal.result

# First, let's collect which files satisfy the rule.
aggregate contains entry if {
    input["package"].path[2].value == "public"
    some i
	input["rules"][i].head.name == "export"
	input["rules"][i].head.assign == true
	input["rules"][i].head.value.type == "array"
    entry := { "export_rule_found" : input.regal.file.name }
}

    not there_is_at_least_one_export_rule
    violation := result.fail(rego.metadata.chain(), {})
}

there_is_at_least_one_export_rule {
	input.aggregate[_].export_rule_found
}
```

Notes: 
* This could be simplified by simple checking if we have >0 aggregates. I wanted to check the negation as a way to see how one could have multiple aggregates collecting different data, and having different violation rules not interfering with each other.
* The aggregate record is an object, which makes it very flexible. One could abuse it, though. For example:

```
# Danger zone below - just to should the potential
aggregate contains entry if {
    # A way to collect all ASTs, providing the author understands the performance limits and risks.
    entry := { "AST" : input }
}
    input.aggregate[_].AST["package"].path[2].value == "public"
    violation := result.fail(rego.metadata.chain(), {})
}
```

sesponda:
Hi @anderseknert @charlieegan3 I've addressed the PR comment and pushed a new version, rebased from the recent changes in the main branch (newbie question: is it OK to rebase form main mid-PR, or you prefer a merge commit? - feel free to point me to best practices, etc.).

I have some test failures:
```
--- FAIL: TestLintRuleIgnoreFiles (0.23s)
    cli_test.go:236: expected two rules to be missing from reported violations, but there were 34 out of 38 violations reported
    cli_test.go:253: expected violation for rule "dubious-print-sprintf"
    cli_test.go:253: expected violation for rule "chained-rule-body"
--- FAIL: TestLintAllViolations (0.26s)
    cli_test.go:187: expected violation for rule "chained-rule-body"
    cli_test.go:187: expected violation for rule "dubious-print-sprintf"
--- FAIL: TestTestRegalBundledBundle (0.76s)
    cli_test.go:333: expected exit status 0, got 2
```

... which happen also in my updated main branch, so I'm not sure what's happening, I'll investigate. Maybe unrelated to my change? 

Next actions:
* If the recent update is OK -> continue with tests (once I sort out why I have other tests failing).
* Do you need help with documentation updates? How do you coordinate the release of a new build with docs? (probably I can learn that deeply studying the CI files but I'm not ashamed to recognise (a) I'm a newbie in Go (including Github CI), (b) I'm not afraid of asking dumb questions (there is a fine line between this and laziness, I know 😸 )

anderseknert:
@sesponda those are e2e tests failing, and the reason is they require a fresh binary built from the head of your changes (after you've rebased from main).

I recently added some code to make sure there _is_ a binary before starting the tests, but there's currently no way to know if it's stale or not. Perhaps it'd be best if we built the binary before starting those tests after all 🤔 Either way, I'll have to figure something out there. To solve the immediate problem, just `go build` and they should work.

(CC @srenatus as we've talked about this before, and you know the problems I ran into).

Help with docs would be great! I don't expect there'd have to be much to start with, and I can expand on them later, but something basic explaining how this works in the docs on custom rule development would be great!

pkg/linter/linter.go
anderseknert: Ah, all good!

anderseknert:
Continuing with some tests seems like a good next step to me, if @charlieegan3 considers his questions resolved.

Awesome work! 👏

pkg/linter/linter.go
charlieegan3: > They are evaluated only if a violation rule reads the aggregates

Ahh 🤦, gotcha! I knew there was something I was missing.

Getting a test in with that example make sense as a good next step to me.

> One could abuse it, though

I think we can live with that. We're not likely to include rules in upstream Regal of that format, but if people want to write and run their own rules that seems ok.

charlieegan3:
All good from me, thanks for taking the time to respond in detail 🙂

pkg/linter/linter.go
anderseknert: From your example,

```rego
entry := { "export_rule_found" : input.regal.file.name }
```

It'd be good if we had an `result.aggregate` function added which always included `input.regal.file.name` in the returned object, so that we'd have a unique value per aggregate object. If we don't there's a risk they'll be munged as a single value when added to the set. But that's an optimization we can save for later too :)

sesponda:
Hi @anderseknert @charlieegan3 thanks for your time so far. I started my work in spare time but later it became company-sponsored contribution (I work for Atlassian), so to keep things tidy I'm creating the same PR (same diff) from a different fork. Let's continue on https://github.com/StyraInc/regal/pull/323 (I've added the e2e tests)

sesponda:
Moved to: https://github.com/StyraInc/regal/pull/323

Issue #248, opened 2023-08-10
Go API


anderseknert:
At some point, we'll want to allow programmatic access via a well defined non-breaking API. We obviously won't stop anyone from using it today, but it'd be good to settle on something stable eventually, and document integrating Regal in Go application.

PR #307, opened 2023-09-11
Expose rules and parser opts for use as library

charlieegan3:
This is just one way to make progress towards: https://github.com/StyraInc/regal/issues/248.

### DIFF
--- a/internal/parse/parse.go
+++ b/internal/parse/parse.go
@@ -39,6 +39,16 @@ func MustParseModule(policy string) *ast.Module {
 	return ast.MustParseModuleWithOpts(policy, ParserOptions())
 }
 
+// Module works like ast.ParseModule but with the Regal parser options applied.
+func Module(filename, policy string) (*ast.Module, error) {
+	mod, err := ast.ParseModuleWithOpts(filename, policy, ParserOptions())
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse module: %w", err)
+	}
+
+	return mod, nil
+}
+
 // EnhanceAST TODO rename with https://github.com/StyraInc/regal/issues/86.
 func EnhanceAST(name string, content string, module *ast.Module) (map[string]any, error) {
 	var enhancedAst map[string]any
--- a/pkg/linter/linter.go
+++ b/pkg/linter/linter.go
@@ -18,6 +18,7 @@ import (
 	"github.com/open-policy-agent/opa/rego"
 	"github.com/open-policy-agent/opa/topdown"
 
+	rbundle "github.com/styrainc/regal/bundle"
 	rio "github.com/styrainc/regal/internal/io"
 	"github.com/styrainc/regal/internal/parse"
 	"github.com/styrainc/regal/internal/util"
@@ -49,7 +50,11 @@ const regalUserConfig = "regal_user_config"
 
 // NewLinter creates a new Regal linter.
 func NewLinter() Linter {
-	return Linter{}
+	regalRules := rio.MustLoadRegalBundleFS(rbundle.Bundle)
+
+	return Linter{
+		ruleBundles: []*bundle.Bundle{&regalRules},
+	}
 }
 
 // WithInputPaths sets the inputPaths to lint. Note that these will be
--- a/cmd/lint.go
+++ b/cmd/lint.go
@@ -15,7 +15,6 @@ import (
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v3"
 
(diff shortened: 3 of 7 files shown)


anderseknert:
Looks good to me! Some things that come to mind:

1. Would it be more convenient if we simply offered a `regal.ParseModule()` function, where the user need not be concerened about parser options? We already have a `regal parse` command that does this, so this would be the programmatic equivalent call.
2. That the user needs to call `regalRules := rio.MustLoadRegalBundleFS(rbundle.Bundle)` and `WithAddedBundle(regalRules)` feels like a bit of cludge to me. I'm wondering if there are any reasons not to make that an internal detail, and just always load the provided bundle in the return of `linter.NewLinter()`? Regal can't really work without that bundle, so it seems senisble to me that the onus to provide that isn't on the user.

WDYT?

anderseknert:
Also, whichever way we go, perhaps we could add an `integration.md` file to the docs directory, providing an example similar to what you did above? We should state in that document that the Go API is still subject to change, and anyone who wants to integrate Regal programmatically is encouraged to reach out before they do so... or something like that :)

charlieegan3:
> Would it be more convenient if we simply offered a regal.ParseModule() function

Maybe, though I can imagine some use cases for not always using the default, particularly around capabilities? Just a thought, not sure if that's worth supporting in our first pass.

> Regal can't really work without that bundle

True, and it doesn't look that pretty. I guess I was just wondering what custom rules/ bundles might look like as a go API user. I feel like we want to make using the default bundle easy but optional. Perhaps we have `WithDefaultBundle` as a linter option, in addition to `WithAddedBundle`?

anderseknert:
IIRC, the parser doesn't make use of capabilities... or does it?

Even custom rules will depend on the provided bundle to be loaded, as that contains the entrypoint for evaluation even of custom rules. So I'd say we remove that requirement altogether and just have Regal load it internally.

charlieegan3:
> IIRC, the parser doesn't make use of capabilities... or does it?

We don't in Regal, but it's an option, along with annotations in OPA which folk might want to use?

charlieegan3:
> entrypoint for evaluation even of custom rules

Ah, ok that makes sense then 👍

anderseknert:
> We don't in Regal, but it's an option, along with annotations in OPA which folk might want to use?

From what I can see, those are only used to decide if future keywords should be allowed, right? If so, I don't think we'll need to take that into account in Regal. Annotations processing OTOH should always be enabled, as many rules will check for the presence of those, validate attributes, etc.

So unless we can foresee some real case where custom parse options are needed, I think providing a way to have Regal rather than OPA do the parsing. We already have this functionality in a custom built-in, so extracting that logic to a public function should be easy, I think. But it's fine to postpone this to later if you want to.

charlieegan3:
> only used to decide if future keywords should be allowed,

Ahh, I didn't realise that. I thought it was also used for checking for valid builtins.

> providing a way to have Regal rather than OPA do the parsing

Generally this does seem to make sense, so I'll look to add this.

internal/parse/parse.go
```diff
@@ -39,6 +39,11 @@ func MustParseModule(policy string) *ast.Module {
 	return ast.MustParseModuleWithOpts(policy, ParserOptions())
 }
 
+// ParseModule works like ast.ParseModule but with the Regal parser options applied.
+func ParseModule(policy string) (*ast.Module, error) {
```
charlieegan3, line 43: This is not public, should it be? It's hard to manually build data for with NewInput without these, but InputFromText and InputFromPaths might make exposing it surplus.

anderseknert: Let’s hide it for now then.

pkg/linter/linter.go
```diff
@@ -68,7 +73,7 @@ func (l Linter) WithInputModules(input *rules.Input) Linter {
 	return l
 }
 
-// WithAddedBundle adds a bundle of rules and data to include in evaluation.
+// WithAddedBundle overrides the default Regal rules with the provided bundle.
```
anderseknert, line 76: It will add to the rules, not override them.

anderseknert (approved):
Great! Just to comments, but feel free to merge whenever.

internal/parse/parse.go
charlieegan3: 👍

pkg/linter/linter.go
charlieegan3: Oops, have fixed that now. I've also opted to use this in our tests: d98d51095527fdc76474169b48587b362100d5a1

PR #309, opened 2023-09-11
Add documentation on architecture

anderseknert:
Some risk that this gets outdated, obviously, so I've tried to avoid referencing locations in code and such. Will still need to revisit every once in a while, but hopefully this is useful for people looking to contribute.

Also made some minor updates to related documentation.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -1,0 +1,47 @@
+# Architecture
+
+Or "How does Regal work?"
+
+As you might have [read](https://www.styra.com/blog/guarding-the-guardrails-introducing-regal-the-rego-linter/), Regal
+[uses Rego for linting Rego](https://www.styra.com/blog/linting-rego-with-rego/) — or rather, Rego policies turned into
+a JSON representation of their abstract syntax tree (AST).
+
+## High-level Overview
+
+When running Regal against a directory, like `regal lint my-policies/`, Regal does the following:
+
+- For each source file provided for linting, Regal parses the Rego into its AST representation. This AST representation
+  is then turned into JSON and provided as the **input** variable to the linter rules.
+- Each linter rule (and there are almost 40 of them at the time of writing this) uses the **input**, which contains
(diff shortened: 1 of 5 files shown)


docs/architecture.md
```diff
@@ -0,0 +1,47 @@
...
+  built-in functions and operators.
+- Since rule bodies aren’t necessarily flat, but may contain nested bodies of constructs such as
+  [comprehensions](https://www.openpolicyagent.org/docs/latest/policy-language/#comprehensions) or
+  [every](https://www.openpolicyagent.org/docs/latest/policy-language/#every-keyword) blocks, many linter rules need to
+  traverse all expressions in order to find what they are looking for. This is normally done with the help of the
+  built-in [walk](https://www.openpolicyagent.org/docs/latest/policy-reference/#graph) function.
+- Traversing huge AST structures — and some policies contain millions of AST nodes! — takes time. This isn’t noticeable
+  when linting a single file, but for some of the largest policy repositories out there, with several thousands of
+  policy files and tests, the cost may be prohibitive. To alleviate that, Regal evaluates as many files as it possibly
+  can in parallel.
```
charlieegan3, line 28: I would have described this as:

> To alleviate this, Regal is implemented to process files concurrently to minimise the impact of IO bound tasks and to make use of multiple cores when available.

> many files as it possibly can in parallel.

When I read this I thought we might get some nit picking from folk talking about goroutine scheduling yada yada.

charlieegan3:
> obviously, so I've tried to avoid referencing locations in code and such

This is always gonna be an impossible task, but I wonder if we left some 'hints' on key functions in the code mentioning the architecture docs as a reminder to keep it updated as and when key code paths are changed? Perhaps it's futile 😅, I have not seen a system like that work before for docs, but I have also never tried.

charlieegan3 (approved):

docs/architecture.md
anderseknert: Sounds reasonable to me! I'll make the change as suggested.

PR #311, opened 2023-09-11
Add note about prod use of Go API

charlieegan3:
Fixes https://github.com/StyraInc/regal/issues/248


charlieegan3:
Thanks Stephan!

PR #312, opened 2023-09-12
Add `--compare-to-readme` flag to `regal table`

### LINKED ISSUE #258: Add check to verify rules table in README is up to date
Basically, as part of the build pipeline, run:

```
go run main.go table --write-to-readme bundle
```

and fail the build if there are any changes in staging.

anderseknert:
And add check in CI so that we don't forget about this when any change would update the table, like when a new rule is added, or the name/description of a current one is modified.

Fixes #258

### DIFF
--- a/cmd/table.go
+++ b/cmd/table.go
@@ -4,6 +4,7 @@ package cmd
 import (
 	"bytes"
 	"errors"
+	"io"
 	"io/fs"
 	"log"
 	"os"
@@ -24,12 +25,13 @@ import (
 )
 
 type tableCommandParams struct {
-	writeToReadme bool
+	compareToReadme bool
+	writeToReadme   bool
 }
 
 func init() {
 	params := tableCommandParams{}
-	parseCommand := &cobra.Command{
+	tableCommand := &cobra.Command{
 		Hidden: true,
 		Use:    "table <path> [path [...]]",
 		Long:   "Create a markdown table from rule annotations found in provided modules",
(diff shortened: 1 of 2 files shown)


cmd/table.go
```diff
@@ -43,15 +44,30 @@ func init() {
 		},
 
 		Run: func(_ *cobra.Command, args []string) {
-			if err := createTable(args, params); err != nil {
-				log.SetOutput(os.Stderr)
-				log.Println(err)
-				os.Exit(1)
+			log.SetOutput(os.Stderr)
```
srenatus, line 47: [nit] I believe that's the default ([see here](https://cs.opensource.google/go/go/+/refs/tags/go1.21.1:src/log/log.go;l=87))

cmd/table.go
```diff
@@ -63,12 +79,12 @@ func unquotedPath(path ast.Ref) []string {
 	return ret
 }
 
-func createTable(args []string, params tableCommandParams) error {
+func createTable(args []string) (bytes.Buffer, error) {
```
srenatus, line 82: [nit] I think it would be slightly more idiomatic to return `io.Reader` here. Since it's an interface, you can then also use the much-beloved `return nil, err` instead

cmd/table.go
```diff
@@ -164,21 +180,44 @@ func writeTable(tableData [][]string, params tableCommandParams) error {
 	table.AppendBulk(tableData)
 	table.Render()
 
-	if !params.writeToReadme {
-		_, err := os.Stdout.Write(buf.Bytes())
-
-		return err
-	}
+	return buf
```
srenatus, line 183: (continued) `&buf` would be an `io.Reader`

cmd/table.go
```diff
@@ -164,21 +180,44 @@ func writeTable(tableData [][]string, params tableCommandParams) error {
...
-	file, err := os.ReadFile("README.md")
-	if err != nil {
-		return err
-	}
+func renderREADME(buf bytes.Buffer) string {
+	file := string(must(os.ReadFile("README.md")))
 
-	first := strings.Split(string(file), "<!-- RULES_TABLE_START -->")[0]
-	last := strings.Split(string(file), "<!-- RULES_TABLE_END -->")[1]
+	first := strings.Split(file, "<!-- RULES_TABLE_START -->")[0]
```
srenatus, line 189: 💭 Looks like a good occasion for using https://pkg.go.dev/strings#CutPrefix and https://pkg.go.dev/strings#CutSuffix? But really, it's fine and working as-is, no need to change stuff.

srenatus (approved):

cmd/table.go
anderseknert: Ah, good! I'll have it removed 👍 

(I guess there's a theoretical chance some other command would have set this, but not in this case)

cmd/table.go
anderseknert: Updated in all locations 👍

cmd/table.go
anderseknert: I don't think that would work here, as those strings don't "start with prefix" or suffix but are rather somewhere deep inside of the README?

srenatus: Hmm I might be wrong 😅 Yeah, looks like I misread the docs. We could do two cuts, but it. doesn't get much better -- https://play.golang.com/p/u8OtalFaste

PR #316, opened 2023-09-12
Store list of refs for use by rules

anderseknert:
And use this to guard expensive `walks` when possible

This gives a modest performance improvement:

**main**
```
go run main.go lint --enable-all ~/tmp/kics/assets  199.62s user 8.53s system 604% cpu 34.459 total
```

**change**
```
go run main.go lint --enable-all ~/tmp/kics/assets  183.42s user 8.19s system 588% cpu 32.560 total
```

But importantly, it'll allow more rules to use this guard in the future, as well as quickly retrieve all built-in functions called in a policy.

Also, fix a correctness issue in `function-arg-return`, where it previously would only report functions with a single part in their name, i.e. `indexof` but not `regex.match`.

### DIFF
--- a/bundle/regal/ast.rego
+++ b/bundle/regal/ast.rego
@@ -263,19 +263,27 @@ is_output_var(rule, ref, location) if {
 	not ref.value in (find_names_in_scope(rule, location) - find_some_decl_names_in_scope(rule, location))
 }
 
-# METADATA
-# description: |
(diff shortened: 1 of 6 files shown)


anderseknert:
Merging without review as this is needed for a subsequent PR. Feel free to take a look anyway and let me know if something looks odd.

Issue #271, opened 2023-08-23
Figure out how to best display errors


anderseknert:
We're doing a lot of wrapping around errors in Regal, and I think that's good for providing context and trace capabilities. It however leads to some error messages being printed in a way that's not really friendly to an end-user. Example:

```
$ regal lint foobar
error(s) ecountered while linting: failed to filter ignored files: failed to filter paths: 1 error occurred during loading: stat foobar: no such file or directory
```

The above error would be better presented as just:

```
$ regal lint foobar
foobar: no such file or directory
```

Or perhaps (and optionally, perhaps when provided the future `--debug` flag) with something that at least presented each "step" of wrapping:

```
$ regal lint foobar
error(s) ecountered while linting:
    failed to filter ignored files: 
        failed to filter paths: 1 error occurred during loading:
            stat foobar: no such file or directory
```

I'm sure there are libraries out there to deal with this, so maybe look into that.

charlieegan3:
Another consideration here is how we allow users of the Go API to get structured error data from Regal, and OPA. This includes both listing errors but also parser errors regal returns. Our functions in https://github.com/StyraInc/regal/blob/main/pkg/rules/rules.go return wrapped parser errors from OPA for example. These need to be accessed like this:

```go
var astErrs ast.Errors
var astErr *ast.Error
response.ErrorMessage = "unknown error"
if errors.As(err, &astErrs) {
	myErrs = astErrs
} else if errors.As(err, &astErr) {
	myErrs = ast.Errors{astErr}
}
```

Some other thoughts:

* it feels like there are two kinds of error, errors related to a line, and errors related to the operation which are not related to a particular line. Either class of error could occur 0 or more times.
* OPA parse errors and Regal lint errors both seem to hold for the above. Might there be some value in standardising parse and lint messages to make them easier for downstream consumption/presentations?

anderseknert:
For the API case — but likely also for us to display nice output of errors in the CLI, I suppose it would make sense for us to create, and be stringent about using, our own error types? We could also provide helpers for filtering out Regal errors from those in underlying libs, and vice versa. But I'd be curious to learn what others are doing here. I'm sure there are many best practices, and possibly some useful libraries, to help deal with this.

PR #317, opened 2023-09-12
Surface parse errors from input building

charlieegan3:
This makes it possible to work with the parse errors when building integrations.

This change is also consistent with the behavior of: InputFromPaths.


anderseknert:
Does this not mean we lose location data outside of OPA? Can’t the client unwrap the error?

I don’t know a lot about how errors are normally surfaced in Go APIs, but I think even OPA does some wrapping of errs here and there?

Not meant as leading questions either, I’m genuinely clueless 😅

charlieegan3:
Closing as we can use errors.Unwrap for now. We can revisit in #271 if we need to.

PR #319, opened 2023-09-13
build: use rq

srenatus:
I haven't updated the readme yet -- let's first discuss if this is a direction we'd like to take 😃

### DIFF
--- a/build/do.rq
+++ b/build/do.rq
@@ -1,16 +1,17 @@
 #!/usr/bin/env rq
 # rq: query data.script.main
 # rq: output null
+import future.keywords
 
 what := rq.args()[0]
 main := do[what] { true }
 else := job[what]
 
 do.pull_request {
-	job.test
-	job.lint
-	job.e2e
-	job.check_readme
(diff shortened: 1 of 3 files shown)


srenatus:
Gave it some optional structure there when run on github
image

build/do.rq
```diff
@@ -0,0 +1,57 @@
...
+do.pull_request {
+	some x in ["test", "lint", "e2e", "check_readme"]
+	github("::group::", x)
+	job[x]
+	github("::endgroup::", x)
+}
+
+job.pr {
+	print("thanks for contributing")
+}
```
charlieegan3, line 19: This isn't used right?

charlieegan3 (commented):
This is cool! 😎 Some explanatory comments in do.rq or in some docs would be helpful I think.

anderseknert (approved):

build/do.rq
anderseknert: No, I'll add those steps later to help run all steps needed to prepare a PR.

PR #323, opened 2023-09-15
Implement Aggregates Collection and Usage

sesponda:
Feature described by: https://github.com/StyraInc/regal/issues/282

Summary: 

This change allows custom rules to collect aggregate data while processing a file AST, which can be used later by other linter rules that require a global context.

Highlights of the change:

1. Added a new feature in the linter to collect aggregates. This is done by creating a new type `Aggregate` that collects data while processing a file AST and can be used later by other rules that need a global context.
2. Modified the `lintWithRegoRules` function to include the aggregates in the evaluation process.
3. Created a helper function `evalAndCollect` to extract common code that processes each file in a goroutine.
4. Updated the end-to-end testing to include the aggregates in the testing process. This includes positive and negative tests.

### DIFF
--- a/e2e/cli_test.go
+++ b/e2e/cli_test.go
@@ -319,6 +319,52 @@ func TestLintRuleNamingConventionFromCustomCategory(t *testing.T) {
 	}
 }
 
+func TestAggregatesAreCollectedAndUsed(t *testing.T) {
+	t.Parallel()
+	cwd := must(os.Getwd)
+	basedir := cwd + "/testdata/aggregates"
+
+	t.Run("Zero violations expected", func(t *testing.T) {
+		stdout := bytes.Buffer{}
+		stderr := bytes.Buffer{}
+
+		err := regal(&stdout, &stderr)("lint", "--format", "json", basedir+"/rego", "--rules", basedir+"/rules/custom_rules_using_aggregates.rego")
+
+		if exp, act := 0, ExitStatus(err); exp != act {
+			t.Errorf("expected exit status %d, got %d", exp, act)
+		}
+
+		if exp, act := "", stderr.String(); exp != act {
+			t.Errorf("expected stderr %q, got %q", exp, act)
+		}
+	})
+
+	t.Run("One violation expected", func(t *testing.T) {
+		stdout := bytes.Buffer{}
+		stderr := bytes.Buffer{}
+		// By sending a single file to the command, we skip the aggregates computation, so we expect one violation
+		err := regal(&stdout, &stderr)("lint", "--format", "json", basedir+"/rego/policy_1.rego", "--rules", basedir+"/rules/custom_rules_using_aggregates.rego")
+
+		if exp, act := 3, ExitStatus(err); exp != act {
+			t.Errorf("expected exit status %d, got %d", exp, act)
+		}
+
+		if exp, act := "", stderr.String(); exp != act {
+			t.Errorf("expected stderr %q, got %q", exp, act)
+		}
+
+		var rep report.Report
+
+		if err = json.Unmarshal(stdout.Bytes(), &rep); err != nil {
+			t.Fatalf("expected JSON response, got %v", stdout.String())
+		}
+
+		if rep.Summary.NumViolations != 1 {
+			t.Errorf("expected 1 violation, got %d", rep.Summary.NumViolations)
+		}
+	})
+}
+
 func TestTestRegalBundledBundle(t *testing.T) {
 	t.Parallel()
 
--- a/pkg/linter/linter.go
+++ b/pkg/linter/linter.go
@@ -46,6 +46,10 @@ type Linter struct {
 	ignoreFiles      []string
 }
 
+type QueryInputBuilder func(name string, content string, module *ast.Module) (map[string]any, error)
+
+type ReportCollector func(report report.Report)
+
 const regalUserConfig = "regal_user_config"
 
 // NewLinter creates a new Regal linter.
@@ -160,7 +164,7 @@ var query = ast.MustParseBody("violations = data.regal.main.report") //nolint:go
 
 // Lint runs the linter on provided policies.
 func (l Linter) Lint(ctx context.Context) (report.Report, error) {
-	aggregate := report.Report{}
+	aggregateReport := report.Report{}
 
 	if len(l.inputPaths) == 0 && l.inputModules == nil {
 		return report.Report{}, errors.New("nothing provided to lint")
@@ -209,23 +213,33 @@ func (l Linter) Lint(ctx context.Context) (report.Report, error) {
 		return report.Report{}, fmt.Errorf("failed to lint using Go rules: %w", err)
 	}
 
-	aggregate.Violations = append(aggregate.Violations, goReport.Violations...)
+	aggregateReport.Violations = append(aggregateReport.Violations, goReport.Violations...)
 
-	regoReport, err := l.lintWithRegoRules(ctx, input)
+	var aggregates []report.Aggregate
+
+	if len(input.Modules) > 1 {
+		// No need to collect aggregates if there's only one file
+		aggregates, err = l.collectAggregates(ctx, input)
+		if err != nil {
+			return report.Report{}, fmt.Errorf("failed to collect aggregates using Rego rules: %w", err)
+		}
+	}
+
+	regoReport, err := l.lintWithRegoRules(ctx, input, aggregates)
 	if err != nil {
 		return report.Report{}, fmt.Errorf("failed to lint using Rego rules: %w", err)
 	}
 
-	aggregate.Violations = append(aggregate.Violations, regoReport.Violations...)
+	aggregateReport.Violations = append(aggregateReport.Violations, regoReport.Violations...)
 
-	aggregate.Summary = report.Summary{
+	aggregateReport.Summary = report.Summary{
 		FilesScanned:  len(input.FileNames),
-		FilesFailed:   len(aggregate.ViolationsFileCount()),
+		FilesFailed:   len(aggregateReport.ViolationsFileCount()),
 		FilesSkipped:  0,
-		NumViolations: len(aggregate.Violations),
+		NumViolations: len(aggregateReport.Violations),
 	}
 
-	return aggregate, nil
+	return aggregateReport, nil
 }
 
 func (l Linter) lintWithGoRules(ctx context.Context, input rules.Input) (report.Report, error) {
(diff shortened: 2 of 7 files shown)


e2e/cli_test.go
```diff
@@ -319,6 +319,52 @@ func TestLintRuleNamingConventionFromCustomCategory(t *testing.T) {
...
+
+		if exp, act := "", stderr.String(); exp != act {
+			t.Errorf("expected stderr %q, got %q", exp, act)
+		}
+	})
+
+	t.Run("One violation expected", func(t *testing.T) {
+		stdout := bytes.Buffer{}
+		stderr := bytes.Buffer{}
+		// By sending a single file to the command, we skip the aggregates computation, so we expect one violation
```
anderseknert, line 345: I understand why this happens, but it's a bit of a mind bend to me having test fail by _not_ running aggregate computation. I'd expect `report` rules that need aggregates to just fail out as undefined, likely by referencing `input.aggregate` which would be undefined if no aggregate rules had run?

The way it's done now, there's no way fo a report rule to know if aggregates have been collected or not, or is there?

anderseknert:
Looks great! I'm not sure about the logic around evaluation of `report` without `aggregate` rules having been collected, but besides that... awesome! 👏

pkg/linter/linter.go
```diff
@@ -421,72 +435,53 @@ func (l Linter) prepareRegoArgs() []func(*rego.Rego) {
...
-			if err != nil {
-				errCh <- fmt.Errorf("failed to convert result set to report: %w", err)
-
-				return
+			if aggregates == nil {
+				// if the value is nil, inserting it "as is" will make it null-defined, and
+				// `count(input.aggregate) == 0` and `not input.aggregate` will not evaluate.
+				// For simplicity and to avoid error-prone code, ensure it's always defined as an array,
+				// so that authors can always check `count(input.aggregate)`
+				// not worrying about nil/null/undefined nuances.
```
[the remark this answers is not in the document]
sesponda: The rationales of ensuring it's always defined as part of the public contract is to avoid common mistakes (as I've seen) like this one:

1. Rego author accidentally breaks the collection of some aggregate data.
2. `input.aggregage` becomes undefined.
3. Violations using `input.aggregate[_]....` will not evaluate, and the linter will pass.

By making #2 impossible, we simplify the work/validations Rego authors need to make.

e2e/cli_test.go
sesponda: > The way it's done now, there's no way fo a report rule to know if aggregates have been collected or not, or is there?

It is. `input.aggregate` is always defined, which allows devs to check if any aggregate is collected simply via  `count(input.aggregate[_]) > 0` without worrying if it's defined or not. This also allows a kind of typed check, e.g. "check that there are N aggregates of specific type" via `not count([x | x = input.aggregate[_].myType]) == N`

anderseknert: Right, but that'll only tell you if any of the aggregate rules has something to report — it's not going to tell you that the aggregate rules were not evaluated at all, which is why I'd prefer it if `input.aggregate` was undefined in that case.

sesponda: > it's not going to tell you that the aggregate rules were not evaluated at all

I did not fully followed this. Maybe some code example? I don't know if by "aggregate rules" you mean rules collecting aggregates (i.e. `aggregate contains entry if`) or the rules evaluating aggregates (i.e. `report contains violation if { .... input.aggregate ... }`

> I'd prefer it if input.aggregate was undefined in that case.

The rationale of merging `not input.aggregates` and `count(input.aggregates[_]) == 0` in to the latter was explained in the [preliminary PR](https://github.com/StyraInc/regal/pull/306#discussion_r1323875777) (in summary: simplify the how one can check aggregates), but that's just my personal preference. If this is a fundamental problem blocking a merge, I don't mind tweaking the PR as needed.
What do you have in mind? Would you like to default on having it `undefined` (if the array is zero), or something different?

anderseknert: Sorry, I should have elaborated more on the why, but I was short on time :)

Consider a rule like this: In package `foo` (which could be distributed across multiple files) I need to ensure that there is exactly **one** rule named `bar`. Some pseudo-code below, but bear with me:

```rego
aggregate contains entry if {
    input["package"].name == "foo"
    some rule in input.rules

    rule.head.name == "bar"

    entry := {
        "file" : input.regal.file.name,
        "message": "found rule named 'bar'"
    }
}

    count(input.aggregate) != 1
    violation := result.fail(rego.metadata.chain(), {})
}
```

If `input.aggregate` is always an array, this rule will fail even when aggregate collection hasn't happened, like when evaluating only a single file, or pehaps has been disabled by other means. In this case, `count(input.aggregate)` will give me `0`, which is of course not equal to 1. If `input.aggregate` is undefined, evluation will stop at that line, and nothing will be reported.

sesponda: > is always an array, this rule will fail even when aggregate collection hasn't happened, like when evaluating only a single file

😮 This discussion has surfaced a bug: your example above makes sense and (regardless of the original point of the discussion), it seems valid to me to write an aggregate that I want to collect always. The whole point of aggregates was to write business rules I can enforce without concerning how I package the code in multiple `.rego` files. In the current version, a refactor merging all files into one would suddenly cause aggregates not being collected, and thus violations could be missed, violating the principle of least surprise. I think we should:

1) When the docs are added, clearly state that aggregates are collected only if files > 1. Ideally, display a warning if we see aggregate rules unused.
2) Alternatively... be more consistent and remove the optimisation, and always collect aggregates. 

I prefer #2. If you agree I'll make the change. If you prefer #1, no worries (however, I can't commit to making those extra changes soon in follow up PRs due to time constraints).

Now, coming back to the original point, i.e. should be undefined or not:

I think that if we encourage writing rules that attempt to find data more specifically, for example:

```
aggregate contains entry if {
    input["package"].name == "foo"
    some rule in input.rules
    rule.head.name == "bar"
    entry := {
        "bar_rule_found" : input.regal.file.name
    }
}

    not count([x | x = input.aggregate[_].bar_rule_found]) == 1
    violation := result.fail(rego.metadata.chain(), {})
}

```
... the logic will work OK for both cases (empty array or undefined).

While I'm still (slightly) inclined to keep it as an empty array, I'm not strong on this one (and I'm here to help the project contributing code and following your lead). I'd be happy to change things. Let me know what changes you'd like and I'll make then soon this week. 

Thanks 🙏

anderseknert: > The whole point of aggregates was to write business rules I can enforce without concerning how I package the code in multiple .rego files.

The point of aggregate rules is to provide a way to lint Rego at the level of a _project_, which is something we can't currently do. If your project consists of a single file, there's really no point in using aggregate rules, as there's nothing to aggregate. That case is covered by regular rules. Indeed, this distinction will need to be well documented, and I'd be happy to flesh that out in follow-up PRs.

See the Regal integration that went out in the [Rego Playground](https://play.openpolicyagent.org/) yesterday for an example of single file linting. This will be increasingly common as we start to look into editor integrations later.

> In the current version, a refactor merging all files into one would suddenly cause aggregates not being collected, and thus violations could be missed, violating the principle of least surprise. 

Yes, a refactor merging all files into one would render aggregate rules moot. That's not surprising to me, although I'd be very surprised if someone did that and called it refactoring :)

> I think that if we encourage writing rules that attempt to find data more specifically

Agreed! Even better than encouraging, I think, is a helper function, similar to how we provide `result.fail`. I suggested this [previously](https://github.com/StyraInc/regal/issues/282#issuecomment-1708528606), but much have been said since then, so that it got lost along the way is understandable :) 

Rather than:

```rego
entry := {
    "bar_rule_found" : input.regal.file.name
}
```

We'd provide a `result.aggregate` helper:
```
entry := result.aggregate(rego.metadata.chain(), {"some_other_interesting": "value"})
```

Since we pass `rego.metadata.chain()`, this helper would be able to find the name of the rule from the annotation, and group the aggregates accordingly, i.e. `input.aggregate["my-rule"]`. It would also add `input.regal.file.name` to each aggregate, and anything else we could think might be useful. The second argument would be for the rule author to populate with whatever is useful in the context of their rule.

So yeah 1. would be me preferred choice here:

> 1. When the docs are added, clearly state that aggregates are collected only if files > 1. Ideally, display a warning if we see aggregate rules unused.

And I'd be happy to help fill in some of the blanks in follow-up fixes, like docs, the above mentioned helper function, (optionally) display warnings if aggregate rules are skipped, and so on. Just let me know how you want to proceed 😃

sesponda: > Just let me know how you want to proceed 😃

IIUC, we have two changes:

1) Make aggregates undefined if not collected, or collected for still of zero-length (i.e. normalise to undefined)
2) Add the helper function `result.aggregate(...)`

IMHO, we could leave 2 for later. Initially, Rego authors can put anything they want in the entry. Later, the project can release the helper function to make it easier to auto-insert the rule name, etc.

I'm happy to help with the minimum change you'd accept a merge, leaving any optionals/wish-to-have for a post-merge. If you think we cannot merge this without #2, no worries... I'll commit to add it (not soon because I have too much on my plate recently, but I still want to end what I started). 

The reason I'm leaning towards speed over coding optionals is that the PR has been open for a a while and I have now non-trivial conflicts to merge. I'll work on them while waiting for your answer.

sesponda:
Hi @anderseknert I've pushed an update. Changes:
- Aggregates not collected now means the collection will be undefined.
- Addressed merge conflicts. I've undo the refactoring extracting common code, because the other feature merged (time metrics) was too attached to the `lintWithRegoRules` method, and I did not want to risk an accidental change in the logic. 

The only missing think is deciding if the helper function `result.aggregate(...)`is a must-have or can be done port-merge at another time as a future improvement, feel free to decide and I'll help. Cheers,

anderseknert (approved):

anderseknert:
Awesome! Let's get this merged, and I'll create some follow-up issues to address before the next release, but I don't expect it to be much. Great work on this @sesponda 👏

anderseknert:
@sesponda 

There were a couple of issues open with the [aggregate](https://github.com/StyraInc/regal/labels/aggregate) label previously, and I've just created some more for things I think we'll need to consider for this type of rule. I may also add a few more later when I've had more time to think about this. I don't consider any of them critical, but perhaps we should call this feature a beta in the next release, so that we at least have some freedom to work on the design in the following releases. But first, let's see how much we are be able to accomplish until then :)

If you have the bandwidth to work on any issue in the list, that's great! But regardless of that, I'd value your thoughts on the issues that involve some design decisions. Please leave a comment on those if you can, and let me know if you want any issue assigned to you.

PR #324, opened 2023-09-15
build/do.rq: build binary for tasks list if needed

srenatus:
`build` is not `build(fresh)`, and will only rebuild the ./regal binary if passed `true`. If passed `false`, it will not rebuild the binary if some binary is present in the given location.

`build(false)` is thus to be used when any regal version will do, like for printing tasks.

### DIFF
--- a/build/do.rq
+++ b/build/do.rq
@@ -56,6 +56,7 @@ do.pull_request {
 # title: tasks
 # description: Prints the name of all available tasks
 job.tasks {
+	build(false)
 	some task in tasks
 	print("-", sprintf("%-20s", [task[0]]), "\t", strings.replace_n({"\n": ""}, task[1]))
 }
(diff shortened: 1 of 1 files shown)


anderseknert (approved):
Neat! Some UNIX assumptions made here, but I guess we’ve got plenty of those already in this build (and I’m not sure rq is even intended to work on Windows) so I think that’s fine.

srenatus:
> I’m not sure rq is even intended to work on Windows

It's not? That's surprising. Many of our users surely use windows for their stuff. Contributing to Regal _should_ also be open to everyone.

🤔 I'll see if I can't do this better with `rq.tree` instead.

anderseknert:
> It's not? That's surprising. 

Oh, `rq` most certainly works, but I meant running as a script like this. I haven't used Windows in ages, but outside of emulation, do even things like:
```
#!/usr/bin/env rq
```
work? Or then you'd just call the binary explicitly I suppose.

PR #325, opened 2023-09-15
do.rq: don't depend on 'ls'

srenatus:
I believe there's more to be done for windows support, but it's one barrier less, maybe.


anderseknert (approved):
Nice!

PR #332, opened 2023-09-18
File filtering improvements

### LINKED ISSUE #330: Long time spent filtering `.git` when linting a project root directory
I was curious to see what took so "long" (hundreds of milliseconds) when linting even with all rules disabled. Turns out that when linting a project root directory, like Regal's own one, more than a hundred milliseconds is being spent filtering out files in the `.git` directory. We're _never_ going to want to look for Rego files there, so that it's even included in the search makes no sense. Adding a counter/timer around the [loader.FilteredPaths](https://github.com/StyraInc/regal/blob/3011243c84a7efca4bb7326769c729034cafc03d/pkg/config/filter.go#L18C3-L20) call:

```go
filteredPaths, err := loader.FilteredPaths(paths, func(_ string, info os.FileInfo, depth int) bool {
    return !info.IsDir() && !strings.HasSuffix(info.Name(), bundle.RegoExt)
})
```

Reveals that a staggering 5904 files have been considered, and this process took ~125 ms.

```shell
$ go run main.go lint --disable-all .
processed 5904 paths in 125.86ms
99 files linted. No violations found.
```

If we remove the `.git` directory _before_ the call:

```shell
$ go run main.go lint --disable-all .
processed 267 paths in 2.82ms
99 files linted. No violations found.
```

And we're down below 3 ms! The reason that this is so expensive is likely how the filter function makes a stat system call on each entry it traverses, which is of course redundant for directories which shouldn't be considered in the first place.

Perhaps this could be made more efficient in OPA? And if so, we should look into upstreaming a fix there too eventually, but as usual, we'll need to take more things into account in that context. For now, let's just remove `.git`, `.idea` and whatever other directory people might have in their project root which isn't relevant to Regal.

anderseknert:
* Ignore .git and .idea directories
* Use more efficient algorithm for filtering than the one provided by OPA

Fixes #330

### DIFF
--- a/pkg/config/filter.go
+++ b/pkg/config/filter.go
@@ -1,35 +1,63 @@
 package config
 
 import (
+	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/gobwas/glob"
 
 	"github.com/open-policy-agent/opa/bundle"
-	"github.com/open-policy-agent/opa/loader"
 )
 
 func FilterIgnoredPaths(paths, ignore []string, checkFileExists bool) ([]string, error) {
-	policyPaths := paths
-
 	if checkFileExists {
-		filteredPaths, err := loader.FilteredPaths(paths, func(_ string, info os.FileInfo, depth int) bool {
-			return !info.IsDir() && !strings.HasSuffix(info.Name(), bundle.RegoExt)
-		})
-		if err != nil {
-			return nil, fmt.Errorf("failed to filter paths: %w", err)
+		filtered := make([]string, 0, len(paths))
+
+		if err := walkPaths(paths, func(path string, info os.DirEntry, err error) error {
+			if info.IsDir() && (info.Name() == ".git" || info.Name() == ".idea") {
+				return filepath.SkipDir
+			}
+			if !info.IsDir() && strings.HasSuffix(path, bundle.RegoExt) {
+				filtered = append(filtered, path)
+			}
+
+			return err
+		}); err != nil {
+			return nil, fmt.Errorf("failed to filter paths:\n%w", err)
 		}
 
-		policyPaths = filteredPaths
+		return filterPaths(filtered, ignore)
 	}
 
 	if len(ignore) == 0 {
-		return policyPaths, nil
+		return paths, nil
 	}
 
-	return filterPaths(policyPaths, ignore)
+	return filterPaths(paths, ignore)
+}
+
+func walkPaths(paths []string, filter func(path string, info os.DirEntry, err error) error) error {
+	var errs error
+
+	for _, path := range paths {
+		// We need to stat the initial set of paths, as filepath.WalkDir
+		// will panic on non-existent paths.
+		_, err := os.Stat(path)
+		if err != nil {
+			errs = errors.Join(errs, err)
+
+			continue
+		}
+
+		if err := filepath.WalkDir(path, filter); err != nil {
+			errs = errors.Join(errs, err)
+		}
+	}
+
+	return errs
 }
 
 func filterPaths(policyPaths []string, ignore []string) ([]string, error) {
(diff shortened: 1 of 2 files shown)


anderseknert:
Heh, pretty much all of my findings from this experiment summarized here: https://benhoyt.com/writings/go-readdir/
It would have saved me some time finding that _before_ I did this, but good to get some validation nonetheless.

srenatus (approved):
LGTM

.idea makes we wonder if we should read .gitignore, or introduce .regalignore.

anderseknert:
Insipired by that blog, I rewrote the implementation using `filepath.WalkDir`. It performs about the same, but it feels more idiomatic than that recursive algo I took from OPA. It moves a little more responsibility over to the filter function, but the upside is that it also provides a whole lot more control 😃  The pattern they use for the callback function here, which now returns an `error` rather than a `bool` is quite nice — by returning a `filepath.SkipDir` "error", the user can stop the walk from traversing a directory, and `return filepath.SkipAll` to stop the traversal entirely.

@srenatus please take another look and let me know what you think :) 

Using `.gitignore` is a good idea! I'll create an issue for that 👍

pkg/config/filter.go
```diff
@@ -3,33 +3,70 @@ package config
...
+	return filterPaths(paths, ignore)
+}
+
+// walkPaths is a simplified version of the same function in OPA, but one that
+// avoids stat-ing files in favor of DirEntry introspection. This is faster due to
+// the reduced number of system calls. Since we avoid stat-ing files, we don't get
+// to work with FileInfo objects, but will rather work directly with the files as
+// string paths. Finally, this function differs from OPA in that we never will
+// traverse down into .git or .idea directories, as that's always a waste of time.
+func walkPaths(paths []string, filter DirEntryFilter) []error {
```
srenatus, line 45: [nit] It doesn't matter, but golang can join errors now: https://pkg.go.dev/errors#Join

pkg/config/filter.go
```diff
@@ -108,3 +145,22 @@ func excludeFile(pattern string, filename string) (bool, error) {
...
+	if len(e) == 1 {
+		return "1 error occurred during loading: " + e[0].Error()
+	}
+
+	buf := make([]string, len(e))
+	for i := range buf {
+		buf[i] = e[i].Error()
+	}
+
+	return fmt.Sprintf("%v errors occurred during loading:\n", len(e)) + strings.Join(buf, "\n")
```
srenatus, line 159: [nit] Inconsequential, but I've played with it so I might as well post the comment 🙃 
```suggestion
	buf := make([]string, len(e)+1)
	buf[0] = fmt.Sprintf("%v errors occurred during loading:", len(e))
	for i := range buf {
		buf[i+1] = e[i].Error()
	}

	return strings.Join(buf, "\n")
```

pkg/config/filter.go
```diff
@@ -3,33 +3,64 @@ package config
...
 
 	"github.com/gobwas/glob"
 
 	"github.com/open-policy-agent/opa/bundle"
-	"github.com/open-policy-agent/opa/loader"
 )
 
-func FilterIgnoredPaths(paths, ignore []string, checkFileExists bool) ([]string, error) {
-	policyPaths := paths
+type DirEntryFilter = func(path string, info os.DirEntry, err error) error
```
srenatus, line 14: [nit] DirEntryFilter doesn't need to be exported. I wonder if it even warrants its own type, but it won't hurt.

srenatus (approved):

pkg/config/filter.go
anderseknert: Oh, that's interesting! Perhaps we could skip that whole `errorsToString` function entirely then? I'll play around with it.

pkg/config/filter.go
```diff
@@ -1,35 +1,63 @@
...
 
 	if len(ignore) == 0 {
-		return policyPaths, nil
+		return paths, nil
+	}
+
+	return filterPaths(paths, ignore)
+}
+
+func walkPaths(paths []string, filter func(path string, info os.DirEntry, err error) error) error {
```
anderseknert, line 42: Screenshot 2023-09-20 at 09 42 21

LOL, errlang strikes again! But you're right, that type was probably superfluous.

srenatus (approved):

PR #333, opened 2023-09-19
Metrics

### LINKED ISSUE #26: Collect and report metrics
In order to know how the linter performs, and if there are certain groups of rules that could be improved, it would be useful if the linter collected and reported on metrics from evaluation. This could be included in the output if some verbose setting was toggled on, or a specific flag for metrics.

anderseknert:
Add Regal-specific metrics in order to know what's taking the most time in linter evaluation.

Also:
* Pass metrics to EvalMetrics to get insights from OPA too
* Have Publish functions accept a context. Unused for now but could be useful later.

Fixes #26

### DIFF
--- a/cmd/lint.go
+++ b/cmd/lint.go
@@ -15,7 +15,10 @@ import (
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v3"
 
+	"github.com/open-policy-agent/opa/metrics"
+
 	rio "github.com/styrainc/regal/internal/io"
+	regalmetrics "github.com/styrainc/regal/internal/metrics"
 	"github.com/styrainc/regal/pkg/config"
 	"github.com/styrainc/regal/pkg/linter"
 	"github.com/styrainc/regal/pkg/report"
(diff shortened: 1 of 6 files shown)


cmd/lint.go
```diff
@@ -224,6 +240,14 @@ func lint(args []string, params lintCommandParams) (report.Report, error) {
 		regal = regal.WithIgnore(params.ignoreFiles.v)
 	}
 
+	if params.metrics {
+		regal = regal.WithMetrics(m)
+	}
+
+	if params.metrics {
+		m.Timer(regalmetrics.RegalConfigParse).Start()
+	}
```
srenatus, line 249: [nit]
```suggestion
	if params.metrics {
		regal = regal.WithMetrics(m)
		m.Timer(regalmetrics.RegalConfigParse).Start()
	}
```

srenatus (approved):

cmd/lint.go
anderseknert: Ah, yes, of course! Thanks, will fix before merge.

PR #335, opened 2023-09-19
Unmarshalling: Remove extra attributes "level" and "ignore"

anderseknert:
As these are assigned to "real" struct attributes, they should not be present in extra attributes too.

Add test to verify.

From discussion in #304

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -160,6 +160,9 @@ func (rule *Rule) UnmarshalYAML(value *yaml.Node) error {
 	return rule.mapToConfig(result) //nolint:errcheck
 }
 
+// Note that this function will mutate the result map. This isn't a problem right now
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -179,5 +179,8 @@ func (rule *Rule) mapToConfig(result map[string]any) error {
 
 	rule.Extra = result
 
+	delete(rule.Extra, "level")
+	delete(rule.Extra, "ignore")
```
srenatus, line 183: 💭 does it alter `result`? Would we care?

srenatus (approved):

pkg/config/config.go
anderseknert: It does, and it's a good point! But at least for now it's an internal method which is only called after a string has been unmarshalled into a map. I'll add a comment to make sure we're aware of that if its use case is extended later.

PR #347, opened 2023-09-25
Allow parameters and constants in top level rule

zregvart:
The `top-level-iteration` is meant to detect rules that have iteration without function body. Referencing parameters or constants should be okay without a function body. This change allows that.


anderseknert (approved):
Excellent. Thanks @zregvart! 😃

PR #349, opened 2023-09-27
Rule: `prefer-package-imports`

### LINKED ISSUE #326: Rule: `prefer-package-imports`
The [Rego Style Guide](https://github.com/StyraInc/rego-style-guide) recommends importing packages rather than rules, as referencing packages makes for easier to read code:

**Avoid**
```rego
package policy

import data.jwt.claims.first_name
import data.jwt.claims.last_name

# ...
# 350 lines of policy code later
# ...

allow {
    # where is first_name and last_name defined?
    name := concat(" ", [first_name, last_name])
    # ...
}
```

**Prefer**
```rego
package policy

import data.jwt

# ...
# 350 lines of policy code later
# ...

allow {
    # obvious where first name and last name comes from
    name := concat(" ", [jwt.claims.first_name, jwt.claims.last_name])
    # ...
}
``` 

I, who wrote that recommendation, consider it to be a good recommendation. Let's include a rule for that in Regal. That we can't "resolve" dynamic rule/function references doesn't really matter in this case — if it's not defined as a _package_ it's not likely a package. I imagine an aggregate rule would aggregate package declaration and import of each provided file, and then a collector rule that tried to match them across all aggregated results.

### LINKED ISSUE #343: Provide `result.aggregate` helper function
Similarly to how we provide ´result.fail` for easily creating well-formed violations in rules, we should provide a `result.aggregate` function that does the same for aggregate rules .This function should:

1. Take the metadata chain as one of its args, and use this to add the full path/name of the package to an entry.
2. Always add `input.regal.file.name` to the returned object, so that we may trace the origin of any aggregate.
3. Take an object of arbitrary attributes that are useful for aggregate collection and reporting.

We should use this function in all examples in the docs, but also document exactly what it provides and returns, and why.

anderseknert:
This builds upon the design laid out in #282, and on @sesponda's excellent work on that. My approach for testing this was to implement an actual "real world" linter rule from the backlog, and `prefer-package-imports` seemed like a good
choice. The rule aggregates package names (or paths) from each input file, along with any import paths.

The report rule then checks each import collected against the list of package paths to determine if the import matches a package. If not, the rule will consider that a violation.

As is often the case when real world meets the original design, "some" modifications proved to be necessary.

The collection of aggregate data is now run alongside the usual linter rule evaluation, as we conveniently may use the same input files for aggregating data. Once that's done, we now send another eval call using only the data aggregated as input. Each aggregate rule is now provided only the data aggregated by _that rule_, so there's no need for policy authors to traverse through all of the aggregates, but one may simply refer to the `input.aggregate` array for the same set of entries as one collected previously.

I've also had to add an alternative schema for the input when these rules are evaluated (currently only enforced by `regal test`). Perhaps `anyOf` would be a better option than two separate schemas, but I'll explore that at a later point.

Fixes #326
Fixes #343


internal/test/rego_test.go
```diff
@@ -16,8 +15,11 @@ import (
 	"github.com/styrainc/regal/pkg/builtins"
 )
 
+// nolint:paralleltest,tparallel
 func TestRunRegoUnitTests(t *testing.T) {
-	t.Parallel()
+	// Temporary until this requirement is removed from OPA
+	// Once removed, re-add t.Parallel()
+	t.Setenv("EXPERIMENTAL_GENERAL_RULE_REFS", "true")
```
charlieegan3, line 22: suggestion: prefix comment with TODO to make it a clearer reminder?

main.go
```diff
@@ -9,9 +9,14 @@ import (
 
 func main() {
 	// Remove date and time from any `log.*` calls, as that doesn't add much of value here
-	// Evaluate options for logging later..
+	// Evaluate options for logging later
 	log.SetFlags(0)
 
+	// Temporary until this requirement is removed from OPA
+	if err := os.Setenv("EXPERIMENTAL_GENERAL_RULE_REFS", "true"); err != nil {
```
charlieegan3, line 16: suggestion: prefix comment with TODO to make it a clearer reminder?

pkg/linter/linter.go
```diff
@@ -565,15 +571,76 @@ func (l Linter) lintWithRegoRules(
 	}
 }
 
+func (l Linter) lintWithRegoAggregateRules(
```
charlieegan3, line 574: Does it makes sense to cover this in linter_test too?

e2e/testdata/violations/most_violations.rego
```diff
@@ -22,6 +22,9 @@ import data.redundant.alias as alias
 # redundant-data-import
 import data
 
+# prefer-package-imports
+import data.rule_named_if.allow
```
charlieegan3, line 26: Worth a comment that this will only fail when there is more than one file linted?

bundle/regal/rules/imports/prefer_package_imports.rego
```diff
@@ -0,0 +1,63 @@
...
+	})
+}
+
+# METADATA
+# schemas:
+#   - input: schema.regal.aggregate
+aggregate_report contains violation if {
+	all_package_paths := {pkg |
+		some entry in input.aggregate
+		pkg := entry.aggregate_data.package_path
```
charlieegan3, line 38: Even if a file has no imports, its package path will still be set here right? I think this is the case, but just wanted to check that `all_package_paths` will be set to all packages, rather than all packages with imports.

charlieegan3 (approved):
Looks good to me!

internal/test/rego_test.go
anderseknert: Will add, thanks!

bundle/regal/rules/imports/prefer_package_imports.rego
anderseknert: Yep!

e2e/testdata/violations/most_violations.rego
anderseknert: Can do 👍

pkg/linter/linter.go
anderseknert: I've added a simple test 👍 And as it's covered by an e2e test too it's probably good enough for the time being.

PR #350, opened 2023-09-27
Add `non-breakable-word-threshold` option to `line-length` rule

### LINKED ISSUE #327: Add an `allow-non-breakable-words` option in `line-length` à la `yamllint` to avoid rule violation in cases where reformatting/splitting line is not possible
There are some cases were splitting a long line is not possible. For example in metadata annotations we could have long URLs:

```rego
# METADATA
# title: This is just an example with a long reference in related resource
# related_resources:
# - ref: https://www.example.org/this/is/very/long/url/unfortunately/sometimes/happens/and/using/url/shortener/could/be/not/nice
#   description: A problematic long URL for regal line-length
package play
```

(OK, we could lie here and use a URL shortener! :))

[yamllint](https://yamllint.readthedocs.io/) [`line-length`](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.line_length) rule has an option called `allow-non-breakable-words` - that is enabled by default - that permit a longer line in such cases.

It would be maybe nice to have something like that in Regal as well.

anderseknert:
Fixes #327

### DIFF
--- a/bundle/regal/rules/imports/prefer_package_imports.rego
+++ b/bundle/regal/rules/imports/prefer_package_imports.rego
@@ -6,7 +6,6 @@ import future.keywords.contains
 import future.keywords.if
 import future.keywords.in
 
-import data.regal.ast
 import data.regal.config
 import data.regal.result
 
--- a/bundle/regal/rules/style/line_length.rego
+++ b/bundle/regal/rules/style/line_length.rego
@@ -9,13 +9,15 @@ import future.keywords.in
 import data.regal.config
 import data.regal.result
 
-report contains violation if {
-	cfg := config.for_rule("style", "line-length")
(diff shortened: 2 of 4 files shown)


anderseknert:
@iamleot let me know what you think 😊

iamleot:
Hello Anders,
I've just tested it and works perfectly!

Thank you very much!

iamleot:
Hello Anders,
I've just tested it and works perfectly!

Thank you very much!

anderseknert:
Awesome @iamleot 😃 Will make it into the next release.

charlieegan3 (commented):
LGTM, nice addition 😎

charlieegan3 (approved):

PR #353, opened 2023-09-28
Remove code that set the ref head env var

anderseknert:
We don't need that anymore since @johanfylling bumped us up to v0.57.0


anderseknert:
Merging without review as this is a trivial change.

PR #354, opened 2023-09-29
Rule: `rule-length`

### LINKED ISSUE #236: Rule length
Similar to the line-length rule, but counting number of _lines_ (or perhaps better, expressions) in a rule, and warn on excessively long ones, like over 25 lines or expressions or something. This ensures proper use of helper functions, rules and other abstractions.

Same thing for packages, or modules/files, which should not be longer than XXX lines.

anderseknert:
Made easy by now having the `text` attribute present in the AST. Thanks for that @charlieegan3!

Fixed existing violations against this rule, which were almost exlusively in tests. Sprinkled a few ignore directives in the places where it wasn't possible to make rules any shorter.

Fixes #236


anderseknert:
Merging without review as the changes around `text` attributes in location is needed in current development tasks.

PR #355, opened 2023-09-30
Rule: `no-defined-entrypoint`

### LINKED ISSUE #154: Require entrypoint annotation
Having one or more entrypoint annotations helps document the... well, entrypoints of a policy, and additionally makes it easy to compile to different targets without command line flags. This could be considered a best practice, and as such worthy of a rule.

anderseknert:
Aggregate rule number two, in which we aggregate entrypoint annotations, and fail if we don't find any.

Fixes #154

### DIFF
--- a/bundle/regal/rules/idiomatic/no_defined_entrypoint.rego
+++ b/bundle/regal/rules/idiomatic/no_defined_entrypoint.rego
@@ -1,0 +1,26 @@
+# METADATA
+# description: Missing entrypoint annotation
+package regal.rules.idiomatic["no-defined-entrypoint"]
+
+import future.keywords.contains
+import future.keywords.if
+import future.keywords.in
+
+import data.regal.ast
+import data.regal.result
+
+aggregate contains entry if {
+	some annotation in input.annotations
+	annotation.entrypoint == true
+
+	entry := result.aggregate(rego.metadata.chain(), {"entrypoint": annotation.location})
+}
+
+# METADATA
+# schemas:
+#   - input: schema.regal.aggregate
+aggregate_report contains violation if {
+	count(input.aggregate) == 0
+
+	violation := result.fail(rego.metadata.chain(), {})
+}
--- a/docs/rules/idiomatic/no-defined-entrypoint.md
+++ b/docs/rules/idiomatic/no-defined-entrypoint.md
@@ -1,0 +1,99 @@
+# no-defined-entrypoint
+
+**Summary**: Missing entrypoint annotation
+
+**Category**: Idiomatic
+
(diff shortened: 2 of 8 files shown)


bundle/regal/rules/idiomatic/no_defined_entrypoint.rego
```diff
@@ -0,0 +1,26 @@
+# METADATA
+# description: No defined entrypoint
```
anderseknert, line 2: This message is a little short. Suggestions welcome!

bundle/regal/rules/idiomatic/no_defined_entrypoint.rego
```diff
@@ -0,0 +1,26 @@
...
+}
+
+# METADATA
+# schemas:
+#   - input: schema.regal.aggregate
+aggregate_report contains violation if {
+	count(input.aggregate) == 0
+
+	violation := result.fail(rego.metadata.chain(), {})
+}
```
anderseknert, line 26: These rules are just _so_ simple! @sesponda note how input aggregate always is defined and at minimum an empty set/array, like how you wanted it to be. This is possible now since we're not even evaluating these rules if aggregate rules are disabled (because only one file linted or for other reasons).

docs/rules/idiomatic/no-defined-entrypoint.md
```diff
@@ -0,0 +1,83 @@
...
+import future.keywords.in
+
+default allow := false
+
+# Nothing wrong with this rule, but an
+# entrypoint should be documented as such
+allow if {
+    some role in input.user.roles
+    role in data.permissions.admin_roles
+}
```
charlieegan3, line 23: Would these examples be clearer if the unnamed entrypoint `allow` showed how entrypoints often gather the results from various other rules?

charlieegan3 (approved):

docs/rules/idiomatic/no-defined-entrypoint.md
anderseknert: Yeah.. I agree that'd be good! I've wanted to make all examples self-contained, so that they'd be copy-pasteable... but might be worth expanding a bit on this example and have a helper rule or two. I'll take a look!

bundle/regal/rules/idiomatic/no_defined_entrypoint.rego
anderseknert: I went with `Missing entrypoint annotation` 👍

docs/rules/idiomatic/no-defined-entrypoint.md
anderseknert: Updated to include two helper rules :)

PR #360, opened 2023-10-01
Rule: `default-over-else`

### LINKED ISSUE #211: Prefer default function to conditionless else
Jumping the gun here :) But adding this as a reminder once the next OPA version is out, and we'll have [default functions](
https://github.com/open-policy-agent/opa/issues/2445) 🤤 Once we do, I'm inclined to say we should:

**Avoid**
```rego
package policy

city(person) := city {
    location := person.location
    city := location.city
} else := "unknown"
```

**Prefer**
```rego
package policy

default city(_) := "unknown"

city(person) := city {
    location := person.location
    city := location.city
} 
```
I'll definitely need to come up with a better example for the docs, but I hope the message is clear regardless 😄

@kristiansvalland @charlieegan3 @srenatus agreed? 

An alternative would be to file this under the "style" category, but I do think this should be considered more than that. EDIT: changed my mind. Considering this for the style category.

anderseknert:
Prefer default assignment over catch-all `else`. This rule was originally meant to endorse the new default function construct, but having tested it out thoroughly for this rule, I'm now more inclined to have it flagged. I find that counter-intuitive, and as such more likely to cause harm than good. Those who disagree may configure the `prefer-default-functions` option. Either way, default assignment over the equivalent `else` feels more idiomatic to me, so the rule feels relevant despite not doing what was originally intended.

Fixes #211

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,7 @@ The following rules are currently available:
 | imports   | [redundant-data-import](https://docs.styra.com/regal/rules/imports/redundant-data-import)         | Redundant import of data                                  |
 | style     | [avoid-get-and-list-prefix](https://docs.styra.com/regal/rules/style/avoid-get-and-list-prefix)   | Avoid get_ and list_ prefix for rules and functions       |
 | style     | [chained-rule-body](https://docs.styra.com/regal/rules/style/chained-rule-body)                   | Avoid chaining rule bodies                                |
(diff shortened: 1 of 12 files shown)


charlieegan3 (approved):
LGTM.

I had the same assumption about undefined args to functions 🤔

anderseknert:
Thanks @charlieegan3! Yeah, I feel like the OPA docs should be amended to mention that, as using default functions without being aware seems like it could lead to some unintended things happening. I'll create an issue for that in the OPA backlog.

PR #363, opened 2023-10-03
Some small tweaks

anderseknert:
* `rule-named-if`: Skip iteration if no rule named "if"
* Detect generated rule bodies without uncertainty
* Move commonly used idioms and helper to ast package
* Added a few comments for docs here and there

### DIFF
--- a/bundle/regal/rules/bugs/constant_condition.rego
+++ b/bundle/regal/rules/bugs/constant_condition.rego
@@ -17,15 +17,8 @@ _operators := {"equal", "gt", "gte", "lt", "lte", "neq"}
 
 _rules_with_bodies := [rule |
 	some rule in input.rules
-	not probably_no_body(rule)
+	not ast.no_body(rule)
 ]
-
-# i.e. allow {..}, or allow := true, which expands to allow = true { true }
-probably_no_body(rule) if {
-	count(rule.body) == 1
-	rule.body[0].terms.type == "boolean"
-	rule.body[0].terms.value == true
-}
 
 report contains violation if {
 	some rule in _rules_with_bodies
--- a/bundle/regal/rules/idiomatic/no_defined_entrypoint_test.rego
+++ b/bundle/regal/rules/idiomatic/no_defined_entrypoint_test.rego
@@ -3,7 +3,6 @@ package regal.rules.idiomatic["no-defined-entrypoint_test"]
 import future.keywords.if
 import future.keywords.in
 
-import data.regal.ast
 import data.regal.config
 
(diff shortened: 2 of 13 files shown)


bundle/regal/rules/bugs/constant_condition.rego
```diff
@@ -17,16 +17,9 @@ _operators := {"equal", "gt", "gte", "lt", "lte", "neq"}
 
 _rules_with_bodies := [rule |
 	some rule in input.rules
-	not probably_no_body(rule)
+	not ast.no_body(rule)
```
charlieegan3, line 20: wonder if there's a way to write this where the function is `has_body` and then not has_body might be easier to understand at a glance.

bundle/regal/rules/idiomatic/no_defined_entrypoint_test.rego
```diff
@@ -3,7 +3,6 @@ package regal.rules.idiomatic["no-defined-entrypoint_test"]
 import future.keywords.if
 import future.keywords.in
 
-import data.regal.ast
```
charlieegan3, line 6: I wonder if we could have a rule for unused imports?

charlieegan3 (approved):

bundle/regal/rules/idiomatic/no_defined_entrypoint_test.rego
anderseknert: We could, but it's in OPA strict mode already, which is why I've chosen not to duplicate this here. Running `opa check --strict` on Regal is a bit clunky though, as we need to provide a capabilities file including our custom built-in functions. We currently don't produce such a file, but we probably should. 

Another option would be to embed `opa check --strict` as a linter rule similar to the `opa fmt` one, but I'm [not pleased](https://github.com/open-policy-agent/opa/issues/5815) with the compiler API.

Seeing as this is likely going to be the default for OPA in some future release, the status quo remains :)

bundle/regal/rules/bugs/constant_condition.rego
anderseknert: Yeah, that's better. I'll merge this but can fix in the next round of tweaks :)

bundle/regal/rules/idiomatic/no_defined_entrypoint_test.rego
charlieegan3: 👍 I'd forgotten about that. Thanks!

anderseknert: Created https://github.com/StyraInc/regal/issues/369 :)

PR #365, opened 2023-10-03
Aggregate rules documentation

### LINKED ISSUE #342: Documentation for aggregate rules
We need to document the new aggregate rules for contributors of built-in rules as well as for authors of custom rules. Things to include:

1. How to author rules using aggregated data.
2. When and when not Regal considers aggregate rules (see #340)
3. Helper functions or helper rules available.
4. A simple example.

anderseknert:
I'm sure there's more to it, but we can iterate on this when we hear back from the first custom aggregate rules authors.

Fixes #342

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,6 +190,16 @@ The following rules are currently available:
 
 By default, all rules except for those in the `custom` category are currently **enabled**.
 
+#### Aggregate Rules
+
+Most Regal rules will use data only from a single file at a time, with no consideration for other files. A few rules
+however require data from multiple files, and will therefore collect, or aggregate, data from all files provided for
+linting. These rules are called *aggregate rules*, and will only be run when there is more than one file to lint, such
+as when linting a directory or a whole policy repository. One example of such a rule is the `prefer-package-imports`
+rule, which will aggregate package names and imports from all provided policies in order to determine if any imports
+are pointing to rules or functions rather than packages. You normally won't need to care about this distinction other
+than being aware of the fact that some linter rules won't be run when linting a single file.
+
 If you'd like to see more rules, please [open an issue](https://github.com/StyraInc/regal/issues) for your feature
 request, or better yet, submit a PR! See the [custom rules](/docs/custom-rules.md) page for more information on how to
 develop your own rules, for yourself or for inclusion in Regal.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -190,6 +190,16 @@ The following rules are currently available:
 
 By default, all rules except for those in the `custom` category are currently **enabled**.
 
+#### Aggregate Rules
+
+Most Regal rules will use data only from a single file at a time, with no consideration for other files. A few rules
+however require data from multiple files, and will therefore collect, or aggregate, data from all files provided for
+linting. These rules are called *aggregate rules*, and will only be run when there are more than one file to lint, such
```
charlieegan3, line 197:
```suggestion
linting. These rules are called *aggregate rules*, and will only be run when there is more than one file to lint, such
```

docs/custom-rules.md
```diff
@@ -164,6 +164,83 @@ Starting from top to bottom, these are the components comprising our custom rule
    will later be included in the final report provided by Regal.
 1. The `result.location` helps extract the location from the element failing the test. Make sure to use it!
 
+## Aggregate Rules
+
+Aggregate rules are a special type of rule that allows you to collect data from multiple files before making a decision.
+This is needed in some cases where a single policy won't be enough for a linter rule to make a decision. For example,
```
charlieegan3, line 170:
```suggestion
This is needed in some cases where a single policy file won't be enough for a linter rule to make a decision. For example,
```

docs/custom-rules.md
```diff
@@ -164,6 +164,83 @@ Starting from top to bottom, these are the components comprising our custom rule
...
+you may want to enforce that in a policy repository, there must be at least one package annotated with an `authors`
+attribute. This requires first collecting annotation data from all the provided policies, and then have the rule use
+this data to make a decision. The structure of an aggregate rule is similar to a regular rule, but with a few notable
+differences. The most significant one is that linting happens in two separate phases — one that aggregates data from
+files, and one that actually lints and reports violations using that data.
```
charlieegan3, line 175: Take or leave this one, I thought it made sense to strictly tie the explanation to the two rules in the code below.

```suggestion
files (`aggregate`), and one that actually lints and reports violations using that data (`aggregate_report`).
```

charlieegan3 (approved):

anderseknert:
Thanks for the review @charlieegan3 👍 I applied all of your suggestions before merging.

Issue #356, opened 2023-09-30
CI/CD: `do.rq pull_request` nullifies `--format github` benefits


anderseknert:
Since we're capturing stdout/stderr here, we're not getting the errors inlined, nor the summary report printed.
```
Run build/do.rq pull_request
 test
 lint
  go build
  
  ./regal lint --format github bundle
  rq: error: main.ScriptCmd.Run(): 
             stdout: ::error file=bundle/regal/rules/idiomatic/no_defined_entrypoint_test.rego,line=22,col=2::Call to print or trace function. To learn more, see: https://docs.styra.com/regal/rules/testing/print-or-trace-call
```

PR #366, opened 2023-10-04
Add run_quiet to rq build script

anderseknert:
So that we don't silence the GitHub workflow commands

Fixes https://github.com/StyraInc/regal/issues/356

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -19,8 +19,8 @@ jobs:
       - uses: actions/setup-go@v4
         with:
           go-version: 1.20.1
+      - run: go install git.sr.ht/~charles/rq/cmd/rq@latest
+      - run: build/do.rq pull_request
       - uses: golangci/golangci-lint-action@v3
         with:
           version: v1.52.2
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yaml
```diff
@@ -15,8 +15,8 @@ jobs:
       - uses: actions/setup-go@v4
         with:
           go-version: 1.20.1
+      - run: go install git.sr.ht/~charles/rq/cmd/rq@latest
+      - run: build/do.rq pull_request
```
anderseknert, line 19: Just moving this up as I'd rather fail fast on builds than golangci-lint, which is... not fast.

build/do.rq
```diff
@@ -215,6 +215,18 @@ run(cmd) {
 	print(out.stdout)
 }
 
+run_quiet(cmd) {
```
anderseknert, line 218: I guess one option is to just make this `run`?

charlieegan3: So the difference is that this will print the raw stdout and that allows GH to comment on lines again even when there's an error?

PR #372, opened 2023-10-05
Dogfood setup-regal action

charlieegan3:
Uses https://github.com/StyraInc/setup-regal

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -24,3 +24,19 @@ jobs:
       - uses: golangci/golangci-lint-action@v3
         with:
           version: v1.52.2
+  # lint-bundle is here to dogfood the setup-regal action from https://github.com/StyraInc/setup-regal
+  lint-bundle:
+    name: Lint Bundle
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+
+      - name: Setup Regal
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -24,3 +24,19 @@ jobs:
       - uses: golangci/golangci-lint-action@v3
         with:
           version: v1.52.2
+  # lint-bundle is here to dogfood the setup-regal action from https://github.com/StyraInc/setup-regal
+  lint-bundle:
+    name: Lint Bundle
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
```
anderseknert, line 33: Might as well make this v4 right away, as it's used elsewhere in this file

bundle/regal/ast.rego
```diff
@@ -1,9 +1,6 @@
 package regal.ast
 
-import future.keywords.contains
-import future.keywords.every
-import future.keywords.if
-import future.keywords.in
+import future.keywords
```
anderseknert, line 3: This should be reverted, I guess?

.github/workflows/build.yaml
charlieegan3: 👍 fixed

anderseknert (approved):

anderseknert:
Neato!

PR #380, opened 2023-10-08
Windows runner test

anderseknert:
Now with Regal working for Windows. Could not get `rq` to work here though so wer're not using that on the Windows runner for the time being.

Also fixed an issue in the e2e tests where stderr wasn't correctly read, and simplified the common case of checking for exit code with a helper function.

Two tests creating new files from templates using the `regal new rule` command are currently skipped on Windows, as they fail for some reason I don't have the time to look into now. The commands work, but `regal test` doesn't seem to find the created temp dir. Not critical at all, so I don't feel bad about skipping those for now.

### DIFF
--- a/e2e/cli_test.go
+++ b/e2e/cli_test.go
@@ -130,17 +130,16 @@ func TestLintNonExistentDir(t *testing.T) {
 	td := t.TempDir()
 
 	err := regal(&stdout, &stderr)("lint", td+filepath.FromSlash("/what/ever"))
-	if exp, act := 1, ExitStatus(err); exp != act {
-		t.Errorf("expected exit status %d, got %d", exp, act)
-	}
 
-	if exp, act := "", stderr.String(); exp != act {
+	expectExitCode(t, err, 1, &stdout, &stderr)
+
+	if exp, act := "", stdout.String(); exp != act {
 		t.Errorf("expected stderr %q, got %q", exp, act)
 	}
 
 	if exp, act := "error(s) encountered while linting: errors encountered when reading files to lint: "+
 		"failed to filter paths:\nstat "+td+filepath.FromSlash("/what/ever")+": no such file or directory\n",
-		stdout.String(); exp != act {
+		stderr.String(); exp != act {
 		t.Errorf("expected stdout %q, got %q", exp, act)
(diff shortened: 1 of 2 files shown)


anderseknert:
Build is green, but I don't think this is doing anything:

Screenshot 2023-10-06 at 09 11 53

anderseknert:
Build is green, but I don't think this rq task actually runs anything here:
Screenshot 2023-10-08 at 10 00 13

charlieegan3:
Might be a good use case for a matrix: https://github.com/StyraInc/setup-regal/blob/main/.github/workflows/test.yml#L17

anderseknert:
@charlieegan3 when I get it working, for sure! Wanted it to be a separate step here to easily make changes to the windows build separately, but it turns out there were more issues than I could debug in the runner.

e2e/cli_test.go
```diff
@@ -506,22 +487,22 @@ func TestCreateNewCustomRuleFromTemplate(t *testing.T) {
...
+
 	stdout := bytes.Buffer{}
 	stderr := bytes.Buffer{}
-
 	tmpDir := t.TempDir()
 
-	err := regal(&stdout, &stderr)("new", "rule", "--category", "naming", "--name", "foo-bar-baz", "--output", tmpDir)
+	expectExitCode(t, regal(&stdout, &stderr)(
+		"new", "rule", "--category", "naming", "--name", "foo-bar-baz", "--output", tmpDir,
+	), 0, &stdout, &stderr)
```
srenatus, line 500: Hmm I think this could go one step further? To let us write
```go
expectExitCode(t, 0, "new", "rule", "--category", "naming", "--name", "foo-bar-baz", "--output", tmpDir)
```
or something like that? 🤔 Deal with the buffers in the helper, etc.

srenatus (approved):

e2e/cli_test.go
anderseknert: I did something like that in the first iteration, but the problem is that many of the tests then go on to check for the contents of the buffers, so we'll need them outside of the helper. There's definitely some improvements to make here though, and I hope to see more in later PRs :)

PR #382, opened 2023-10-09
Fixes to run on Windows

anderseknert:
Could be more are needed, but I'll need this in main to test cloning from a blank state.

### DIFF
--- a/e2e/cli_test.go
+++ b/e2e/cli_test.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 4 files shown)


srenatus (approved):

e2e/cli_test.go
```diff
@@ -118,13 +119,17 @@ func TestLintEmptyDir(t *testing.T) {
...
+	}
+
 	t.Parallel()
 
 	stdout := bytes.Buffer{}
 	stderr := bytes.Buffer{}
 	td := t.TempDir()
 
-	err := regal(&stdout, &stderr)("lint", td+"/what/ever")
+	err := regal(&stdout, &stderr)("lint", td+filepath.FromSlash("/what/ever"))
```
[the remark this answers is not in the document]
anderseknert: Nah, just figured it won't hurt to have if we want to unskip it later :)

internal/parse/parse.go
```diff
@@ -62,7 +62,7 @@ func EnhanceAST(name string, content string, module *ast.Module) (map[string]any
 	enhancedAst["regal"] = map[string]any{
 		"file": map[string]any{
 			"name":  name,
-			"lines": strings.Split(content, "\n"),
+			"lines": strings.Split(strings.ReplaceAll(content, "\r\n", "\n"), "\n"),
```
charlieegan3, line 65: I wonder if it'd be faster to use a Regex? `\r?\n` (rather than 2 replacements)

Also, a comment to explain that this specifically to handle windows line endings would be welcome since that might not be immediately clear.

charlieegan3 (approved):

internal/parse/parse.go
anderseknert: Thanks! I'll add a comment in a follow-up PR. About the regex, I guess we'd need to measure 😅

PR #383, opened 2023-10-09
Fix issues blocking Windows usage

anderseknert:
The logic traversing directories looking for a `.regal` directory did not take a possible volume name (like `C:`) into account. With this, all tests (unit + e2e)  now pass on Windows as well as UNIX.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -50,8 +50,17 @@ func FindRegalDirectory(path string) (*os.File, error) {
 		dir = filepath.Dir(path)
 	}
 
+	// returns e.g. "C:" on windows, "" on other platforms
+	volume := filepath.VolumeName(dir)
+
 	for {
-		searchPath := filepath.Join(rio.PathSeparator, dir, regalDirName)
+		var searchPath string
+		if volume == "" {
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -61,14 +70,19 @@ func FindRegalDirectory(path string) (*os.File, error) {
...
 			// Stop traversing at the root path
 			return nil, fmt.Errorf("can't traverse past root directory %w", err)
 		}
 
 		// Move up one level in the directory tree
 		parts := strings.Split(dir, rio.PathSeparator)
 		parts = parts[:len(parts)-1]
+
+		if parts[0] == volume {
+			parts[0] = volume + rio.PathSeparator
```
charlieegan3, line 83: Is this needed? I thought that file path.Join would be smart enough to do that?

charlieegan3 (approved):

pkg/config/config.go
anderseknert: It did not, so you'd get `C:Users` rather than `C:\Users`. Maybe there's a cleaner way to do this, but this one at least works :)

charlieegan3: Good to know 😞

PR #386, opened 2023-10-10
`use-assignment-operator`: check for assignment operator in else assignment

### LINKED ISSUE #385: use-assignment-operator does not detect else-clauses declared using unification
**The "use-assignment-operator" does not detect else-cases like the following:** 

decision := {...} {
} **else =** {...} {}

The linter rule should suggest this instead:
decision := {...} {
} **else :=** {...} {}

I've parsed both and their ASTs are identical - as the unification operator is simply parsed into an assignment.

anderseknert:
Also simplified the logic of this rule for other types of rule assignment.

Fixes #385


charlieegan3 (approved):
Looks good to me!

PR #387, opened 2023-10-10
Rule: `import-after-rule`

### LINKED ISSUE #376: Import following rule
Having worked with Rego for years, I still had no idea that this was valid. @johanfylling showed me that it was:

**Avoid**
```rego
package p

rule := b

import data.b
```

It would surprise me if anyone _ever_ does this, so realistically we'll write code for this that no one will run, documentation about it that no one will read, and a test case no one will notice. But.. this is the life we chose.

anderseknert:
Fixes #376


charlieegan3 (approved):
Very clear implementation for this one! 👌

PR #402, opened 2023-10-14
Fix false positive in `top-level-iteration`

### LINKED ISSUE #401: `top-level-iteration`: false positive reported when assigning input
Curiously, this one is reported as iteration :)
```rego
resource := input
```
Likely an easy fix.

anderseknert:
Fixes #401


anderseknert:
Some irony that fixing this would surface another false positive, but in the `constant-condition` rule 😄

Screenshot 2023-10-14 at 22 54 32

Running `opa fmt` actually "fixes" the error reported, but not in a satisfying way, I think :/

```rego
package p

f("x")
```

```
$ opa fmt p.rego
package p

f("x") = true
```

Meh 😑 I don't think that should be rewritten, and if it should be, we should at least use `:=`. WDYT @srenatus?

I'll fix the false positive in `constant-condition` either way.

charlieegan3 (approved):

PR #407, opened 2023-10-16
Rule: `import-shadows-builtin`

### LINKED ISSUE #406: Forbid shadowing built-in function "namespaces"
As @johanfylling suggested [here](https://github.com/orgs/open-policy-agent/discussions/498#discussioncomment-7291074), it would be useful to have a rule that forbids shadowing of any built-in "namespace". A namespace in this context would be the word preceding built-in functions, like `http` in `http.send`, `array` in `array.slice`, `io` in `io.jwt.decode` and so on.

Using these namespaces won't cause OPA to complain until you try to also use one of the built-in functions "owning" that namespace. While OPA flags that case, we could be more proactive in Regal and let the user know that this will potentially be problematic.

If there is a better terminology to use here, I'm all ears :)

anderseknert:
Forbid imports that may shadow a built-in function, like `data.print`, or `input.http`.

Fixes #406


charlieegan3 (approved):
LGTM! 👏

PR #411, opened 2023-10-17
Rule: `equals-pattern-matching`

### LINKED ISSUE #397: Prefer function arg pattern matching for equals condition
Few know about this capability, so it'd be great if Regal would educate its users on this!

**Avoid** (is a strong word here, but...)
```rego
package policy

is_one(number) := number == 1

bool_to_string(bool) := "true" if bool == true
bool_to_string(bool) := "false" if bool == false
```

**Prefer**
```rego
package policy

is_one(1)

bool_to_string(true) := "true"
bool_to_string(false) := "false"
```

anderseknert:
Prefer to use pattern matching for equality in function args.

Currently limited to simple rules where the equality check is the only condition in the rule head or rule body (and this limitation is described in the rule docs). I'm hoping to improve this in the future, but as I think there's value in this already, so let's ship and iterate.

Fixes #397

### DIFF
--- a/bundle/regal/rules/idiomatic/equals_pattern_matching.rego
+++ b/bundle/regal/rules/idiomatic/equals_pattern_matching.rego
@@ -1,0 +1,82 @@
+# METADATA
+# description: Prefer pattern matching in function arguments
+package regal.rules.idiomatic["equals-pattern-matching"]
+
+import future.keywords.contains
+import future.keywords.if
+import future.keywords.in
+
+import data.regal.ast
+import data.regal.result
+
+# Current limitations:
+# Only works for single comparison either in head or in body
+
+# f(x) := x == 1
+# ->
+# f(1)
+report contains violation if {
+	some fn in ast.functions
+	ast.generated_body(fn)
+
+	arg_var_names := {arg.value |
+		some arg in fn.head.args
+		arg.type == "var"
+	}
+
+	val := fn.head.value
+	val.type == "call"
+	val.value[0].type == "ref"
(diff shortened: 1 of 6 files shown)


charlieegan3 (approved):
LGTM!

bundle/regal/rules/idiomatic/equals_pattern_matching.rego
```diff
@@ -0,0 +1,82 @@
...
+
+import future.keywords.contains
+import future.keywords.if
+import future.keywords.in
+
+import data.regal.ast
+import data.regal.result
+
+# Current limitations:
+# Only works for single comparison either in head or in body
```
srenatus, line 13:
```rego
f(x) := x if x == 1
```
Will this pass or fail?

srenatus (approved):
One small edge case question 👍

bundle/regal/rules/idiomatic/equals_pattern_matching.rego
anderseknert: It will fail, as it could be rewritten, although that case is arguable contrived:
```
f(1) := 1
```
I added a unit test to verify just now:
https://github.com/StyraInc/regal/pull/411/files#diff-2800e3666d00c08742a4a86e8a7557117cd8ee69172cd5bcc6f3fff410f79a3cR50-R55

Do you think we should ignore that?

srenatus: This is fine, thanks!

Issue #212, opened 2023-07-21
Support capabilities feature from OPA


anderseknert:
> We could even ship with a capabilities config that supported e.g. Gatekeeper, EOPA, and whatever else there is, out of the box. Having to bring that whole JSON file and edit it manually always felt somewhat "low-level" to me. We should look into some easier mechanics for Regal, like allowing capabilities to be defined in the Regal conf.

From https://github.com/StyraInc/regal/issues/211

Adding capabilities via config could look something like:

```yaml
rules: 
  testing:
    test-outside-test-package:
      level: ignore

capabilities:
  from:
    # could be opa, eopa, gatekeeper, or any other known project extending opa
    engine: opa
    version: v0.54.0
  plus:
    builtins:
      - name: ldap.query
        type: function
        decl:
          args:
            - type: string
        result:
          type: object
  minus:
    builtins:
      - name: http.send 
```
We should also accept a `capabilities.json` file from OPA, since some users likely already have one. 
```yaml
capabilities:
  from:
    file: ${project_root}/capabilities.json
```

CLI flags for the most common and quick options would be nice too I supoose, but no hard requirement IMHO.

anderseknert:
Would also be cool to have built-in support for linting Rego scripts from the [rq](https://sr.ht/~charles/rq/) tool by @charlesdaniels.

srenatus:
One paper cut with rq would be that its Rego files don't need a package declaration, while opa still requires that.

anderseknert:
Yeah, not sure how rq does that. I suppose just identify the file type first and then prepend a `package script` to the file before parsing it would work.

charlesdaniels:
Yeah, rq just prepends `package script` to the text of the Rego code if it does not find any line beginning with `package`, so you could do the same. Another option if you are OK with a subshell would be to do use `rq --save-bundle`, which dumps the code+data that `rq` would have run your query on to a normal OPA bundle (though it will still have rq-specific builtins).

anderseknert:
Given the many upcoming changes to OPA, like `import future.strict`, and eventually a 1.0 release, we'll need to support capabilities in Regal as soon as possible. 

While we obviously will want people to upgrade OPA to run the latest version, we'll likely see old versions around for a _long_ time. For that reason, I think we'll also need to allow rules to define their "requirements" on things like OPA (or project building on OPA) version, but could also be other things, like the existance of some built-in function. The latter would not be needed for a first iteration though.

[...]

PR #423, opened 2023-10-26
config: Allow configuration of capabilities

charlieegan3:
This is a first pass at making it possible to configure capabilities in regal.

I haven't yet updated the regoArgs to use this from the config since I wanted to get some feedback on the impl first.

Related: https://github.com/StyraInc/regal/issues/212

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -17,9 +17,9 @@ import (
 const capabilitiesEngineOPA = "opa"
 
 type Config struct {
-	Rules        map[string]Category `json:"rules"            yaml:"rules"`
-	Ignore       Ignore              `json:"ignore,omitempty" yaml:"ignore,omitempty"`
-	Capabilities ast.Capabilities    `json:"capabilities,omitempty"     yaml:"capabilities,omitempty"`
+	Rules        map[string]Category `json:"rules"                  yaml:"rules"`
+	Ignore       Ignore              `json:"ignore,omitempty"       yaml:"ignore,omitempty"`
+	Capabilities ast.Capabilities    `json:"capabilities,omitempty" yaml:"capabilities,omitempty"`
 }
 
 func (config *Config) UnmarshalYAML(value *yaml.Node) error {
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -83,11 +83,6 @@ func (config *Config) UnmarshalYAML(value *yaml.Node) error {
 		config.Capabilities = *capabilities
 	}
 
-	// by default, use the capabilities from the current OPA
-	if capabilitiesEngine == "" && capabilitiesFile == "" {
-		config.Capabilities = *ast.CapabilitiesForThisVersion()
-	}
-
 	// remove any builtins referenced in the minus config
 	for i, builtin := range config.Capabilities.Builtins {
 		for _, minusBuiltin := range result.Capabilities.Minus.Builtins {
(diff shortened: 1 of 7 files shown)


anderseknert:
Looks great @charlieegan3 👍 I think it makes sense to add a `file` option too while we do this, so that it's easy to bring your own if you have one already.

charlieegan3:
I've added this in cebcec16a00b716f060e477c3290421e8b5e5217, I think it makes sense to have an error when setting the engine and file.

pkg/config/config.go
```diff
@@ -9,12 +9,92 @@ import (
...
+		for _, minusBuiltin := range result.Capabilities.Minus.Builtins {
+			if minusBuiltin.Name == builtin.Name {
+				config.Capabilities.Builtins = append(config.Capabilities.Builtins[:i], config.Capabilities.Builtins[i+1:]...)
+			}
+		}
+	}
+
+	config.Capabilities.Builtins = append(config.Capabilities.Builtins, result.Capabilities.Plus.Builtins...)
+
+	return nil
```
anderseknert, line 97: If no capabilities have been provided, should we default to latest from OPA here?

pkg/config/config.go
```diff
@@ -9,12 +9,92 @@ import (
...
 )
 
+const capabilitiesEngineOPA = "opa"
+
 type Config struct {
-	Rules  map[string]Category `json:"rules"            yaml:"rules"`
-	Ignore Ignore              `json:"ignore,omitempty" yaml:"ignore,omitempty"`
+	Rules        map[string]Category `json:"rules"            yaml:"rules"`
+	Ignore       Ignore              `json:"ignore,omitempty" yaml:"ignore,omitempty"`
+	Capabilities ast.Capabilities    `json:"capabilities,omitempty"     yaml:"capabilities,omitempty"`
```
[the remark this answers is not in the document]
charlieegan3: Yeah, that doesn't look right...  I have fixed it manually in eeda01e979937165fa93c320fb0d090a667c6192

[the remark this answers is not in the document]
anderseknert: Thanks! I'll look into why the linter didn't flag this.

pkg/config/config.go
charlieegan3: I've had a go a finishing this off a little more in 1fff993

anderseknert: Awesome! Thanks 😃 Looks good to me.

The one thing I'm not sure about is passing the capabilities from the config to Rego's eval args. Would that not have the linter (i.e. Regal) itself be limited to whatever capabilities the target policy has? I mean, even if we're linting a Rego file with e.g. a pre-`contains` OPA version, surely _we_ should be able to use contains in Regal's linter policy? 

I think the way we'll want to use the capabilities at least initially, is basically as a datasource for what built-in functions we might encounter in the policy we're linting, and have linter rules be able to declare their dependencies... like how we should skip the [custom-has-key-construct](https://docs.styra.com/regal/rules/idiomatic/custom-has-key-construct) if `object.keys` is not in the capabilities, and so on. 

But if we eval with the provided capabilities, it would mean _we_ can't use `object.keys`, no?

charlieegan3: Ahh, I see. I think I got the wrong end of the stick! I can take this part out.

charlieegan3: I've taken out the changes in https://github.com/StyraInc/regal/pull/423/commits/5d79a7be38680ff08fa539d8593166ec274c4d4a and then added the defaulting back in https://github.com/StyraInc/regal/pull/423/commits/d292b58d6a3ab12ff136474dbef5f8e28ace0fb8

anderseknert: Thanks!

anderseknert (approved):

PR #435, opened 2023-11-01
`rule-length`: add `count-comments` configuration option

### LINKED ISSUE #432: `rule-length` should not count comments in body, unless configured to do so
The `rule-length` rule currently just counts the number of lines from the start of the rule body, to the closing curly brace. That's acceptable for a first version, but it would be good if the rule didn't punish commenting code within rules. I suggest we subtract lines containing only comments from the total, unless a configuration option to include them have been toggled.

anderseknert:
This is set to `false` by default, as we don't want to "punish" comments in rules.

Fixes #432

### DIFF
--- a/bundle/regal/config/provided/data.yaml
+++ b/bundle/regal/config/provided/data.yaml
@@ -93,6 +93,7 @@ rules:
       ignore-nesting-level: 2
       level: error
     rule-length:
+      count-comments: false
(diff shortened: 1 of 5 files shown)


charlieegan3 (approved):
Guess it's good to give the option, but I feel like generally I'd not want to include comments in these rules.

PR #437, opened 2023-11-01
docs: make logo smaller

charlieegan3:
For some reason, I don't think that the markdown preview is showing the correct version of the image, perhaps as it's updated as part of the same PR?


anderseknert (approved):
Thanks! Let's see what it'll look like :)

Issue #425, opened 2023-10-26
Add configurable exception case to `prefer-some-in-iteration`


anderseknert:
There are a few cases where `some .. in` can't _directly_ replace `[_]` iteration. In the [gatekeeper-library](https://github.com/open-policy-agent/gatekeeper-library) repo, they do a whole lot of:

```rego
input_containers[c] {
    c := input.review.object.spec.containers[_]
}
input_containers[c] {
    c := input.review.object.spec.initContainers[_]
}
input_containers[c] {
    c := input.review.object.spec.ephemeralContainers[_]
}
```

Which without [unconditional-assignment](https://docs.styra.com/regal/rules/style/unconditional-assignment) becomes:

```rego
input_containers[input.review.object.spec.containers[_]]

input_containers[input.review.object.spec.initContainers[_]]

input_containers[input.review.object.spec.ephemeralContainers[_]]
```

Using `some ... in` doesn't work in that location, so the alternative would be:

```rego
input_containers[c] {
    some c in input.review.object.spec.ephemeralContainers
}

# or

input_containers contains c if some c in input.review.object.spec.ephemeralContainers
```

But that's arguably an "unconditional assignment" — or at least an unconditional `if` — too. If one prefers the shorter `[_]` syntax in this case I think that should be allowed as an option.

PR #442, opened 2023-11-06
Rule: `unnecessary-some`

### LINKED ISSUE #440: Rule: `unnecessary-some` when some is not used for iteration
## Description

In my opinion, the following

```rego
some "a" in ["a", "b"]
```

is allowed, but looks weird and would be better written as

```rego
"a" in ["a", "b"]
```

I.e. the `some` keyword is in this case unnecessary, giving the name to this rule.

I first thought that this was a bug in the OPA syntax and I considered adding an issue there, but then I discovered the types given in the "Potential pitfalls" section that are kind of similar but are correct.

## Testing

I tested the following policy with Regal v0.11.0, and no complaints from existing lints were given

```rego
package play

import future.keywords.if
import future.keywords.in

a if {
        some "a" in ["a", "b"]
}
```

## Potential pitfalls

The following should not be flagged by the lint rule

```rego
some idx, "a" in ["a", "b"]
```

anderseknert:
Fixes #440


bundle/regal/ast.rego
```diff
@@ -402,6 +402,31 @@ is_chained_rule_body(rule, lines) if {
 	startswith(col_text, "{")
 }
 
+imports_keyword(imports, keyword) if {
+	some imp in imports
+
+	_has_keyword(imp, keyword)
+}
```
srenatus, line 409: [style] I wonder if Regal would flag this 🙃 

```suggestion
imports_keyword(imports, keyword) if _has_keyword(imports[_], keyword)
```

bundle/regal/ast.rego
```diff
@@ -402,6 +402,31 @@ is_chained_rule_body(rule, lines) if {
...
+	imp.path.value[0].value == "future"
+	imp.path.value[1].value == "keywords"
+	imp.path.value[2].value == keyword
+}
+
+_has_keyword(imp, _) if {
+	count(imp.path.value) == 2
+	imp.path.value[0].value == "rego"
+	imp.path.value[1].value == "v1"
+}
```
srenatus, line 428: hmm looking at these, I'd propose

```rego
imports_keyword(imports, keyword) if {
	some imp in imports
	_has_keyword(arr(imp), keyword)
}

arr(xs) := [ y.value | some y in x.path.value ]

_has_keyword(["future", "keywords"], _)

_has_keyword(["future", "keyword", kw), kw)

_has_keyword(["rego", "v1"], _)
```

srenatus (approved):
Nitpicks only

bundle/regal/ast.rego
anderseknert: It would, I think, but see #425 suggests allowing more exceptions to that.

bundle/regal/ast.rego
anderseknert: That's nice! I'll try that 👍

anderseknert: Done :) Thanks!

bundle/regal/ast.rego
```diff
@@ -402,6 +402,24 @@ is_chained_rule_body(rule, lines) if {
...
+# description: |
+#   returns whether a keyword is imported in the policy, either explicitly
+#   like "future.keywords.if" or implicitly like "future.keywords" or "rego.v1"
+imports_keyword(imports, keyword) if {
+	some imp in imports
+
+	_has_keyword(arr(imp), keyword)
+}
+
+arr(xs) := [y.value | some y in xs.path.value]
```
srenatus, line 415: hmm looking at this now, we could just inline `arr(.)`, or promote it to some utility method. But it's fine either way, let's keep it.

srenatus (approved):

kristiansvalland (approved):
Looks nice! 👏

bundle/regal/rules/style/unnecessary_some.rego
```diff
@@ -0,0 +1,43 @@
...
+
+import future.keywords.contains
+import future.keywords.if
+import future.keywords.in
+
+import data.regal.ast
+import data.regal.result
+
+report contains violation if {
+	# No need to traverse rules here if we're not importing `if`
```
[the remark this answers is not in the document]
anderseknert: Ah, good catch! I'll have that fixed while I try and figure out what changed in the formatter between v0.58 and 59, as apparently that's what is failing the build.

anderseknert:
Alright, I had to remove the "dogfooding" step where we pull down Regal via `setup-regal`. It's nice in theory, but we'll also be linting using an older version of Regal than what the PR provides, which is evidently problematic. Might want to revisit this later, but let's skip that for the time being.

PR #443, opened 2023-11-07
Allow rules to define capabilities requirements

### LINKED ISSUE #212: Support capabilities feature from OPA

anderseknert:
The goal of this is twofold:
- To allow some rules to only run when their required capabilities are met. This could mean either because a new capability makes an older rule obsolete (which is great!) or a missing capability from an older version prevents Regal from running a rule (which is not great!)
- To make Regal a helpful tool for OPA/Rego 1.0 migrations. This is extra useful now that Regal is integrated in the Rego Playground, and can start to suggest tips for 1.0 compatibility as early as from the next release.

Changes made to accomplish this:
- Pass merged config to Rego — don't merge in policy on each invocation as that'll be even more expensive now that the config object is quite big
- Use capabilities from config for rules that need details on e.g. arity. Previously we'd use a builtin_metadata.json file fetched at random occasions, but we don't really need that now, and code relating to that has been removed.
- Remove the code that would run Rego tests via `go test`. This was mostly an escape hatch before as we had some problems running the test command on Regal itself. Since that's been fixed, this duplication is no longer needed.
- Some unrelated minor fixes of new golangci-lint suggestions.
- Make use of capabilities in rules (like `object.keys`)
- Decomission old built-in "registry" fetching built-in metadata
- Report notices like unmet capabilities leading to some rules not being evaluated.

Fixes #212


bundle/regal/rules/custom/one_liner_rule.rego
```diff
@@ -12,6 +12,17 @@ import data.regal.result
...
+#   severity: warning
+notices contains result.notice(rego.metadata.chain()) if not has_if
+
+# if if if!
+has_if if "if" in config.capabilities.future_keywords
+
+has_if if "rego_v1_import" in config.capabilities.features
+
 # regal ignore:rule-length
 report contains violation if {
```
johanfylling, line 27: If `has_if` is `undefined`, is there a point in checking for violations?
Or we don't evaluate the `report` for a rule if it has `notices`?

Same question for all other rules.

bundle/regal/rules/testing/dubious_print_sprintf_test.rego
```diff
@@ -14,6 +14,7 @@ test_fail_print_sprintf if {
 	}`)
 
 	r := rule.report with input as module
+		with data.internal.combined_config as {"capabilities": data.regal.capabilities}
```
johanfylling, line 17: This is such a common statement it might be worthwhile adding a convenience rule/func for it.

bundle/regal/rules/custom/one_liner_rule.rego
johanfylling: Ok, we don't evaluate the `report` for a rule with `violations`.

johanfylling (approved):
LGTM 👍 
With the caveat that I'm not super familiar with this repo 😅 .

bundle/regal/rules/testing/dubious_print_sprintf_test.rego
anderseknert: Indeed — some test helper would be good. I have a few other ones to add, so will follow up in a future PR.

bundle/regal/config/config.rego
```diff
@@ -3,40 +3,16 @@ package regal.config
...
-# the level for this rule, so we should fall back on the provided level, i.e. that which was
-merged_rules[category] := rules if {
-	some category, _rules in config_union.rules
-	rules := {rule_name: empty_level_to_provided(category, rule_name, rule_conf) |
-		some rule_name, rule_conf in _rules
-	}
-}
-
-empty_level_to_provided(_, _, conf) := conf if conf.level != ""
+merged_config := data.internal.combined_config
```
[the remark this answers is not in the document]
anderseknert: Yeah, the latter pretty much. It's not a rule, just something I tend to follow without thinking about it.

PR #445, opened 2023-11-08
Rule: `yoda-condition`

### LINKED ISSUE #424: Rule: `yoda-condition`
While I don't think it's that common, a rule to flag "yoda conditions" — equality comparisons with the constant on the left hand side — would be a good addition to Regal. This is a style preference only, so the `style` category seems appropriate. 

**Avoid**
```rego
package policy

allow {
    "GET" == input.request.method
    "users" == input.request.path[0]
}
```

**Prefer**
```rego
package policy

allow {
    input.request.method == "GET"
    input.request.path[0] == "users"
}
```

https://en.wikipedia.org/wiki/Yoda_conditions

anderseknert:
Fixes #424

### DIFF
--- a/bundle/regal/rules/style/yoda_condition.rego
+++ b/bundle/regal/rules/style/yoda_condition.rego
@@ -1,0 +1,30 @@
+# METADATA
+# description: Yoda condition
+package regal.rules.style["yoda-condition"]
+
+import future.keywords.contains
+import future.keywords.if
+import future.keywords.in
+
+import data.regal.ast
+import data.regal.result
+
+report contains violation if {
+	walk(input.rules, [_, value])
+
+	value[0].type == "ref"
+	value[0].value[0].type == "var"
+	value[0].value[0].value in {"equal", "neq"} # perhaps add more operators here?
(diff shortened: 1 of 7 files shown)


bundle/regal/rules/style/yoda_condition.rego
```diff
@@ -0,0 +1,30 @@
...
+
+import data.regal.ast
+import data.regal.result
+
+report contains violation if {
+	walk(input.rules, [_, value])
+
+	value[0].type == "ref"
+	value[0].value[0].type == "var"
+	value[0].value[0].value == "equal"
```
johanfylling, line 17: Should we do the same for `neq`?

johanfylling (approved):
👍 
Just one question.

bundle/regal/rules/style/yoda_condition.rego
anderseknert: Hehe yeah I guess that's even more devilish :) I think all "operators" are reasonable to include here, but we can scope this to equality for now. I'll add `neq` too.

PR #448, opened 2023-11-08
Rule: `inconsistent-args`

### LINKED ISSUE #444: Rule: `inconsistent-args`
Custom functions should have consistently named args for each time they're repeated:

**Avoid**
```rego
foo(a, b) if a == b

foo(b, a) if b > a
```

**Prefer**
```rego
foo(a, b) if a == b

foo(a, b) if a > b
```

Obviously, this should only count vars and not scalars or wildcards.

anderseknert:
Warn on inconsistent argument names in repeated function declarations.

Fixes #444

### DIFF
--- a/bundle/regal/result.rego
+++ b/bundle/regal/result.rego
@@ -56,10 +56,10 @@ _category_title_from_path(path) := [category, title] if ["regal", "rules", categ
 _category_title_from_path(path) := [category, title] if ["custom", "regal", "rules", category, title] = path
 
 # Provided rules, i.e. regal.rules.category.title
-fail(chain, details) := violation if {
-	is_array(chain) # from rego.metadata.chain()
+fail(metadata, details) := violation if {
+	is_array(metadata) # from rego.metadata.chain()
 
-	some link in chain
+	some link in metadata
 	link.annotations.scope == "package"
 
 	some category, title
@@ -76,10 +76,10 @@ fail(chain, details) := violation if {
 }
 
 # Custom rules, i.e. custom.regal.rules.category.title
-fail(chain, details) := violation if {
-	is_array(chain) # from rego.metadata.chain()
+fail(metadata, details) := violation if {
+	is_array(metadata) # from rego.metadata.chain()
 
-	some link in chain
+	some link in metadata
 	link.annotations.scope == "package"
 
 	some category, title
--- a/bundle/regal/rules/bugs/inconsistent_args.rego
+++ b/bundle/regal/rules/bugs/inconsistent_args.rego
@@ -1,0 +1,63 @@
+# METADATA
+# description: Inconsistently named function arguments
+package regal.rules.bugs["inconsistent-args"]
+
+import future.keywords.contains
+import future.keywords.if
+import future.keywords.in
+
+import data.regal.ast
+import data.regal.result
+
+report contains violation if {
+	count(ast.functions) > 0
+
+	# Comprehension indexing — as made obvious here it would be great
+	# to have block level scoped ignore directives...
+	function_args_by_name := {name: args_list |
+		some i
+
+		# regal ignore:prefer-some-in-iteration
+		name := ast.ref_to_string(ast.functions[i].head.ref)
+		args_list := [args |
+			some j
+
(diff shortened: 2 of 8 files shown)


bundle/regal/result.rego
```diff
@@ -56,10 +56,10 @@ _category_title_from_path(path) := [category, title] if ["regal", "rules", categ
 _category_title_from_path(path) := [category, title] if ["custom", "regal", "rules", category, title] = path
 
 # Provided rules, i.e. regal.rules.category.title
-fail(chain, details) := violation if {
```
johanfylling, line 59: 🐶🍖

bundle/regal/rules/bugs/inconsistent_args.rego
```diff
@@ -0,0 +1,63 @@
...
+
+	# Comprehension indexing — as made obvious here it would be great
+	# to have block level scoped ignore directives...
+	function_args_by_name := {name: args_list |
+		some i
+
+		# regal ignore:prefer-some-in-iteration
+		name := ast.ref_to_string(ast.functions[i].head.ref)
+		args_list := [args |
+			some j
```
johanfylling, line 23: Probably won't make a dent on performance, but we could add a `i != j` here, right?

bundle/regal/rules/bugs/inconsistent_args.rego
```diff
@@ -0,0 +1,63 @@
...
+
+			# regal ignore:prefer-some-in-iteration
+			ast.ref_to_string(ast.functions[j].head.ref) == name
+
+			# regal ignore:prefer-some-in-iteration
+			args := ast.functions[j].head.args
+		]
+		count(args_list) > 1
```
johanfylling, line 31: Won't the arg list size always be > 1, as each rule is counted at least twice, once when iterating `i` and once for `j`?

bundle/regal/rules/bugs/inconsistent_args.rego
```diff
@@ -0,0 +1,63 @@
...
+	count(ast.functions) > 0
+
+	# Comprehension indexing — as made obvious here it would be great
+	# to have block level scoped ignore directives...
+	function_args_by_name := {name: args_list |
+		some i
+
+		# regal ignore:prefer-some-in-iteration
+		name := ast.ref_to_string(ast.functions[i].head.ref)
+		args_list := [args |
```
johanfylling, line 22: We'd possibly get a perf boost by making `args_list` a set.

johanfylling: If we didn't need to account for primitives, we could then simply just have made a `count(args_list) > 1` check 😄 .

bundle/regal/rules/bugs/inconsistent_args.rego
```diff
@@ -0,0 +1,63 @@
...
+		some arg in position
+		arg.type == "var"
+		not startswith(arg.value, "$")
+	}
+	count(named_vars) > 1
+}
+
+# Return the _second_ function found by name, as that
+# is reasonably the location the inconsistency is found
+find_function_by_name(name) := [fn |
```
johanfylling, line 60: Unless we can a pinpoint the first function with an actual argument name inconsistency, isn't it more clear to just point to the first function declaration.

johanfylling (approved):
👍

bundle/regal/rules/bugs/inconsistent_args.rego
anderseknert: But the args_list only iterates on `j`.

bundle/regal/rules/bugs/inconsistent_args.rego
anderseknert: We still need to iterate over it, so I don't think a set would perform better? 🤔

bundle/regal/rules/bugs/inconsistent_args.rego
anderseknert: Possibly, but the second function is at least arguably more likely to be inconsistent than the first :P

PR #449, opened 2023-11-08
Don't set capabilities on `regal new rule`

### LINKED ISSUE #446: The `regal new rule` command writes capabilities object to the provided config
This is mostly an annoyance, but looks like the `regal new rule` command is now writing the capabilities generated to the configuration file provided in the bundle as part of adding a new entry for the rule created. We should have it not do that.

anderseknert:
Tbh, we should probably reconsider if we should add capabilities as part of unmarshalling? But for now let's just fix the bug.

Fixes #446

### DIFF
--- a/cmd/new.go
+++ b/cmd/new.go
@@ -162,18 +162,15 @@ func addToDataYAML(params newRuleCommandParams) error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


johanfylling (approved):
👍

PR #455, opened 2023-11-10
Optimization: less use of walk

anderseknert:
The rules that walk refs may benefit from having the AST tree walked *once* with refs collected, and then simply iterate over that collection. Also, a few  micro-optimizations in the hotspots reported by `--profile`.

Eval time reduced by 4-5%, so nothing dramatic, but still good.


srenatus (approved):
LGTM

PR #456, opened 2023-11-10
e2e: fix custom caps assertions

srenatus:


anderseknert:
Quite some happy path testing! 😅 Thanks for spotting that 👍

anderseknert (approved):

PR #458, opened 2023-11-10
No more "file" in locations!

anderseknert:
As we now can exclude that: https://github.com/open-policy-agent/opa/pull/6398

About a 10% reduction in eval time linting the Kics repo 🚀

We will still add a "file" attribute to locations reported in violations, but we'll source that from `input.regal.file.name` instead, and won't have to worry about `walk` traversing thousands of redundant nodes.


srenatus (approved):
🥳

PR #461, opened 2023-11-11
Rule: `if-empty-body`

### LINKED ISSUE #451: Rule: `if-empty-object`
**Avoid**
```rego
package policy

import rego.v1

allow if {}
```
Using `if`, and soon enough we all will, what was previously an empty rule body error is now interpreted as an empty object, which doesn't render an error but rather has `allow` evaluate to `true`. This is obviously a mistake, and while hopefully caught by tests, we should flag it.

anderseknert:
Flag empty "body" like `{}` as that's no longer considered a body.

Also some light refactoring here, breaking out capabilities checks to a separate package. More to happen here soon!

Fixes #451


anderseknert:

PR #464, opened 2023-11-12
Concurrent parse in `InputFromPaths`

### LINKED ISSUE #462: Parallellize `InputFromPaths`
For most repos this doesn't matter that much (the `bundle` directory in Regal takes ~120 ms to parse), but in the largest policy repos I've pointed Regal at, the sequential parsing happening in [InputFromPaths](https://github.com/StyraInc/regal/blob/main/pkg/rules/rules.go#L59) accounts for as much as 2-4 seconds, and we're not doing anything meaningful in the meantime. This seems like a good target for parallellization, as ordering doesn't matter and we just want to get this out of the way and start linting.

It's tempting to move this to the already parallellized Rego eval step, but I think that would require some drastic changes to the API, and we'll likely want to provide readily parsed modules for Go rules as well.. so, let's just keep this function as it is but parse the files lined up concurrently.

anderseknert:
Good news is this reduced the total time to parse, as expected. The bad news is that this wasn't the metric I thought it was, and the real killer is `timer_rego_input_parse_ns`, which we can't fix as easily.

Oh well, a small win is still a win.

Fixes #462

### DIFF
--- a/pkg/rules/rules.go
+++ b/pkg/rules/rules.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"sort"
+	"sync"
 
 	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/loader"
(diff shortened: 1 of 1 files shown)


anderseknert:

PR #467, opened 2023-11-13
Various micro-optimizations

anderseknert:
Mostly farting around today while home sick, and didn't have the energy for more than these silly tricks.

Unsurprisingly, micro-optimizations render micro-improvements, but I'd like to think I learnt some things in the process too.

```
hyperfine --warmup 1 "regal-main lint bundle" "regal-tricks lint bundle"
```
```
Benchmark 1: regal-main lint bundle
  Time (mean ± σ):      1.640 s ±  0.027 s    [User: 11.746 s, System: 0.464 s]
  Range (min … max):    1.599 s …  1.679 s    10 runs

Benchmark 2: regal-tricks lint bundle
  Time (mean ± σ):      1.619 s ±  0.033 s    [User: 11.500 s, System: 0.458 s]
  Range (min … max):    1.562 s …  1.666 s    10 runs

Summary
  regal-tricks lint bundle ran
    1.01 ± 0.03 times faster than regal-main lint bundle
```

### DIFF
--- a/bundle/regal/rules/bugs/not_equals_in_loop.rego
+++ b/bundle/regal/rules/bugs/not_equals_in_loop.rego
@@ -9,8 +9,7 @@ import future.keywords.in
 import data.regal.result
 
 report contains violation if {
-	some rule in input.rules
-	some expr in rule.body
+	expr := input.rules[_].body[_]
 
 	expr.terms[0].type == "ref"
 	expr.terms[0].value[0].type == "var"
--- a/bundle/regal/rules/bugs/constant_condition.rego
+++ b/bundle/regal/rules/bugs/constant_condition.rego
@@ -21,8 +21,7 @@ _rules_with_bodies := [rule |
 ]
 
 report contains violation if {
-	some rule in _rules_with_bodies
-	some expr in rule.body
+	expr := _rules_with_bodies[_].body[_]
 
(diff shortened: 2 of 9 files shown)


bundle/regal/rules/bugs/not_equals_in_loop.rego
```diff
@@ -9,8 +9,7 @@ import future.keywords.in
 import data.regal.result
 
 report contains violation if {
-	some rule in input.rules
-	some expr in rule.body
+	expr := input.rules[_].body[_]
```
srenatus, line 13: Doesn't this lead to the same code once the compiler's stages ran over it...? 🤔

anderseknert: Almost:

```rego
report[__local5__] {
	__local2__ = input.rules[__local1__];
	__local5__ = __local2__.body[__local4__]
}
```
vs
```rego
report[__local0__] {
	__local0__ = input.rules[_].body[_]
}
```
Whether the difference is meaningful? 🤷

anderseknert: Also makes me curious about the whereabouts of `__local__3` :)

anderseknert:

PR #470, opened 2023-11-13
Rule: `use-if` and `use-contains`

### LINKED ISSUE #468: Rule: use-if and use-contains
It would be nice to have linter warnings for policies that are not using the `if` and `contains` keywords as these are planned to be required in 1.0. I looked at the docs and didn't see any rules that matched this description.

anderseknert:
Recommend the use of `if` and `contains` when available as capabilities.

Some fixes to tests where we didn't live up to this standard ourselves. Great!

(Note that this is targeting the `if-empty-object` branch as that contained a few improvements that I wanted here, so that should be merged before this PR)

Fixes #468


charlieegan3 (approved):
Looks good to me, do we need to make any changes to [implicit-future-keywords](https://github.com/StyraInc/regal/pull/470/files#implicit-future-keywords)?

anderseknert:
@charlieegan3 `import rego.v1` has some issues in current version of OPA and is not included in the capabilities. That'll be fixed in v0.59.0, and then we need to think about what to recommend and at what time. But for now these can be unconditionally recommended.

PR #471, opened 2023-11-14
Use `import rego.v1`

anderseknert:
Note that for now, this change only affects what we use internally in Regal, and not what we enforce.

This will be done too in due time, but given that `rego_v1_import` isn't a capability in any released version of OPA yet, we should hold off on recommending `import rego.v1` until there is, and the feature works without known bugs, which will be from the next release.


anderseknert:
Build failing due to that issue where OPA v0.58.0 thinks `default` rules should have an `if`... which is a good reminder why we shouldn't recommend this just yet. We'll use `edge` for the OPA version in CI in the meantime.

anderseknert:
Some test to be fixed, but I think I may hold off on this one until the release after OPA v0.59.0. I'm feeling uneasy about having a released Regal version depend on capabilities that aren't in a released version of OPA yet, and since the default capabilities of Regal is gotten from the OPA version we depend on, this could potentially be problematic. I'll think more about it.

anderseknert:
Having thought more about it, let's not hold off on the merge here as this will likely be hairy to try and keep up to date for the next two weeks. Since we'll likely want to push a release before OPA v0.59.0, we should however make sure that we set the capabilities for that release to v0.58.0 and not the development commit Regal currently depends on.

PR #475, opened 2023-11-16
Use capabilities from v0.58.0

anderseknert:
And not the later commit we're pointing at for rego.v1 functionality. We don't want to promote capabilities that users will need a dev version of OPA to get.

Also, don't run `use-contains` and `use-if` if `rego.v1` is imported, as that's enforced by OPA already.


charlieegan3 (approved):
LGTM

PR #488, opened 2023-11-22
Rule: `boolean-assignment`

### LINKED ISSUE #465: Rule: `boolean-assignment`
**Avoid**
```rego
is_snake_case(str) := str == lower(str)
```

**Prefer**
```rego
is_snake_case(str) if str == lower(str)
```

Controversial, perhaps? 😅 Just finding a few of these myself in Regal's codebase, and I'd like to be made aware of them, as the assignment is almost always redundant, and leaning in on undefined should be considered idiomatic.

The rule could also check for assignment _from_ boolean functions:

```rego
allow if {
	sw := startswith("f", "a")
	# ...
}
```
But perhaps that should be a different rule? 🤔

anderseknert:
Prefer boolean expression in rule body over assigning the result in the rule head.

Also did some consistency fixes in the docs "Related Resources" sections.

Fixes #465


srenatus (approved):
LGTM

PR #495, opened 2023-11-28
[build]: Add matrix PR build

charlieegan3:
This allows us to test the build on linux, mac and windows with the same set of steps. Previously, rq hadn't worked on windows and macOS is a new addition.


anderseknert (approved):
Very nice! Thanks <3

PR #499, opened 2023-11-28
`use-in-operator`: extend check to include static refs

### LINKED ISSUE #453: `use-in-operator` should work with refs as "item"
The [use-in-operator](https://docs.styra.com/regal/rules/idiomatic/use-in-operator) rule currently suggests using `in` for this case:

```rego
rule {
    "foo" == input.bar[_]
}
```

But not this case:

```rego
rule {
    input.foo == input.bar[_]
}
```
We should improve the rule to cover that case.

anderseknert:
Fixes #453

### DIFF
--- a/bundle/regal/rules/idiomatic/use_in_operator.rego
+++ b/bundle/regal/rules/idiomatic/use_in_operator.rego
@@ -4,34 +4,21 @@ package regal.rules.idiomatic["use-in-operator"]
 
 import rego.v1
 
+import data.regal.ast
 import data.regal.result
 
 report contains violation if {
 	some terms in eq_exprs_terms
 
-	terms[1].type in {"array", "boolean", "object", "null", "number", "set", "string", "var"}
-	terms[2].type == "ref"
+	nl_terms := non_loop_term(terms)
+	count(nl_terms) == 1
 
-	last := regal.last(terms[2].value)
+	nlt := nl_terms[0]
+	static_term(nlt.term)
 
-	last.type == "var"
-	startswith(last.value, "$")
-
-	violation := result.fail(rego.metadata.chain(), result.location(terms[2].value[0]))
-}
-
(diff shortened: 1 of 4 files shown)


bundle/regal/rules/idiomatic/use_in_operator.rego
```diff
@@ -9,35 +9,44 @@ import data.regal.result
...
+static_term(term) if term.type in {"array", "boolean", "object", "null", "number", "set", "string", "var"}
 
-	terms[0].type == "ref"
-	terms[0].value[0].type == "var"
-	terms[0].value[0].value == "equal"
+static_term(term) if {
+	term.type == "ref"
+
+	every t in array.slice(term.value, 1, count(term.value)) {
+		t.type == "string"
```
srenatus, line 50: 💭 What about non-string scalars?
```rego
p[x].foo = "bar" if x := true
p[x].foo = "bar" if x := 1
p[x].foo = "bar" if x := set()
```
⏬ 
```json
{
    "p": {
        "1": {
            "foo": "bar"
        },
        "[]": {
            "foo": "bar"
        },
        "true": {
            "foo": "bar"
        }
    }
}
```
not the most common thing, but still "static".

srenatus (approved):
One probably irrelevant nitpick, LGTM

bundle/regal/rules/idiomatic/use_in_operator.rego
anderseknert: That's true! So rather than checking for "string", we could do:
```rego
every t in array.slice(ref.value, 1, count(ref.value)) {
    t.type != "var"
}
```
?

srenatus: :tada: every-not-p is not-some-p, so I guess there are more potential places 😉

PR #500, opened 2023-11-29
go.mod: switch back to 1.20

srenatus:
and run `go mod tidy`.

Go 1.20.x and 1.21.x are both supported, let's not restrict Regal usage from Go without a strong reason.


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/styrainc/regal
 
-go 1.21
+go 1.20
```
anderseknert, line 3: It would be good to have a comment here explaining our "strategy", which I take it would be to align with OPA, right?

srenatus: Honestly, aligning with OPA is worse -- OPA supports the supported golang versions and one unsupported version. I'm not entirely convinced that this is a good idea, so this is a compromise.

srenatus: Added a comment ✔️

anderseknert (approved):

PR #502, opened 2023-11-29
pkg/version: remove spurious whitespace from String()

srenatus:
```
Version:    0.7.0
Go Version: go1.20.7
Platform:   darwin/amd64
Commit:        unknown
Timestamp:  2023-08-11T08:51:41Z
```
I found it slightly misaligned there... probably not intentionally so


anderseknert (approved):
Thanks!

PR #521, opened 2023-12-12
Rule: `default-over-not`

### LINKED ISSUE #517: Rule: `default-over-not`
Seeing this pattern quite frequently, i.e. one assignment for the "happy path" and another for the negated/**not** condition. Using `default` assignment is arguably much clearer, and should be considered the preferred option here.

**Avoid**
```rego
target_size := input.request.object.spec.replicas

target_size := 1 if not input.request.object.spec.replicas
```

**Prefer**
```rego
default target_size := 1

target_size := input.request.object.spec.replicas
```

anderseknert:
Also, start to break up ast.rego into more files, as we started to violate the `file-length` rule. We should continue on this work in follow-up PRs, but for now just make sure we aren't breaking our own rules.

Fixes #517

### DIFF
--- a/bundle/regal/ast/ast.rego
+++ b/bundle/regal/ast/ast.rego
@@ -326,6 +326,24 @@ static_ref(ref) if every t in array.slice(ref.value, 1, count(ref.value)) {
 	t.type != "var"
 }
 
+static_rule_ref(ref) if every t in array.slice(ref, 1, count(ref)) {
+	t.type != "var"
+}
+
+# METADATA
+# description: |
+#   return the name of a rule if, and only if it only has static parts with
+#   no vars. This could be "username", or "user.name", but not "user[name]"
+static_rule_name(rule) := rule.head.ref[0].value if count(rule.head.ref) == 1
+
+static_rule_name(rule) := concat(".", array.concat([rule.head.ref[0].value], [ref.value |
+	some i, ref in rule.head.ref
+	i > 0
+])) if {
+	count(rule.head.ref) > 1
+	static_rule_ref(rule.head.ref)
+}
+
 # METADATA
 # description: provides a set of all built-in function calls made in input policy
 builtin_functions_called contains name if {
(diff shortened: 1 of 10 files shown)


JAORMX (approved):

bundle/regal/ast/ast.rego
```diff
@@ -326,6 +326,24 @@ static_ref(ref) if every t in array.slice(ref.value, 1, count(ref.value)) {
 	t.type != "var"
 }
 
+static_rule_ref(ref) if every t in array.slice(ref, 1, count(ref)) {
+	t.type != "var"
+}
```
srenatus, line 331: 👏

bundle/regal/ast/ast.rego
```diff
@@ -326,6 +326,24 @@ static_ref(ref) if every t in array.slice(ref.value, 1, count(ref.value)) {
...
+
+# METADATA
+# description: |
+#   return the name of a rule if, and only if it only has static parts with
+#   no vars. This could be "username", or "user.name", but not "user[name]"
+static_rule_name(rule) := rule.head.ref[0].value if count(rule.head.ref) == 1
+
+static_rule_name(rule) := concat(".", array.concat([rule.head.ref[0].value], [ref.value |
```
srenatus, line 339: Would this yield the right thing for
```rego
p["just.like.that"].x := true
```
?

Haven't read to the end, I don't know if it matters here

docs/rules/style/default-over-not.md
```diff
@@ -0,0 +1,54 @@
...
+
+username := input.user.name
+```
+
+## Rationale
+
+While both forms are valid, using the `default` keyword to assign a constant value in the fallback case better
+communicates intent, avoids negation where it isn't needed, and requires less instructions to evaluate. Note that this
+rule only covers simple cases where one rule assigns the "happy" path, and another rule assigns on the same condition
+negated. This is by design, as using `not` and negation may very well be the right choice for more complex cases!
```
srenatus, line 32: This description is helpful! 💡 Especially with the deliberately simple approach

bundle/regal/ast/ast.rego
anderseknert: It wouldn't _print_ correctly, but it would work for the purpose of identifying other rules with the same name... which at least at this point will be good enough. But I can add a comment on that.

PR #525, opened 2023-12-18
Rule: `deprecated-builtin`

### LINKED ISSUE #509: Rule: `deprecated-builtin`
While this is covered by `opa check --strict`, it makes me sad to see people running Regal and still not know about this. Contrary to rules like "unused assignment", this will be trivial to add in Regal, so I suggest we do. The docs should link to OPA strict mode docs to help spread awareness on that.

anderseknert:
Fixes #509

### DIFF
--- a/bundle/regal/rules/bugs/deprecated_builtin.rego
+++ b/bundle/regal/rules/bugs/deprecated_builtin.rego
@@ -1,0 +1,39 @@
+# METADATA
+# description: Avoid using deprecated built-in functions
+package regal.rules.bugs["deprecated-builtin"]
+
+import rego.v1
+
+import data.regal.ast
+import data.regal.config
+import data.regal.result
+
+report contains violation if {
+	deprecated_builtins := {
+		"any", "all", "re_match", "net.cidr_overlap", "set_diff", "cast_array",
+		"cast_set", "cast_string", "cast_boolean", "cast_null", "cast_object",
+	}
+
+	# if none of the deprecated built-ins are in the
+	# capabilities for the target, bail out early
+	any_deprecated_builtin(object.keys(config.capabilities.builtins), deprecated_builtins)
+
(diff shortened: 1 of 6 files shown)


bundle/regal/rules/bugs/deprecated_builtin.rego
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	# if none of the deprecated built-ins are in the
+	# capabilities for the target, bail out early
+	any_deprecated_builtin(object.keys(config.capabilities.builtins), deprecated_builtins)
+
+	some ref in ast.all_refs
+
+	ref[0].value[0].type == "var"
+	not ref[0].value[0].value in {"input", "data"}
```
srenatus, line 24: Correct, but redundant, isn't it? This could never pass `name in deprecated_builtins` since that set contains nothing starting with input or data.

bundle/regal/rules/bugs/deprecated_builtin.rego
```diff
@@ -0,0 +1,39 @@
...
+
+	name := concat(".", [value |
+		some part in ref[0].value
```
srenatus, line 27: Interesting. We do depend on iteration ordering here, since it's not really _some_ part we care, but all parts in order. Not a problem, I think, but it reads a bit oddly

bundle/regal/rules/bugs/deprecated_builtin.rego
anderseknert: Ah, yes, copy-pasta 🍝  I'll fix

anderseknert: But OTOH, it'll avoid having to concat the parts in the first place on long `input`/`data` refs? So I think this is fine to keep.

PR #529, opened 2024-01-05
Allow ignore directive anywhere in comment

### LINKED ISSUE #527: Inline regal ignore comments don't work on the same line as a violating TODO comment
When attempting to add a `# regal ignore:  inline regal ignore: todo-comment` to TODO comments, it won't work unless they are on the line before: https://play.openpolicyagent.org/p/496I1Oy5tC

Won't work: (produces regal error)
```rego
# TODO: Add thing once other thing is in place. # regal ignore:todo-comment
```
Will work: (doesn't produce regal error)
```rego
# regal ignore:todo-comment
# TODO: Add thing once other thing is in place.
```

anderseknert:
Fixes #527

### DIFF
--- a/bundle/regal/main.rego
+++ b/bundle/regal/main.rego
@@ -156,12 +156,10 @@ ignore_directives[row] := rules if {
 	some comment in ast.comments_decoded
 	text := trim_space(comment.Text)
 
-	startswith(text, "regal")
-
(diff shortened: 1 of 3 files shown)


anderseknert:
No one else around today, and this is uncontroversial and trivial, so I'll merge without review. Feel free to take a look post merge!

PR #530, opened 2024-01-11
Rule: `duplicate-rule`

### LINKED ISSUE #427: Rule: `duplicate-rule`
Another policy library find was this in one of their tests:

```rego
input_review := {
  "object": {
    "metadata": {
      "name": "nginx"
    }
  }
}

input_review := {
  "object": {
    "metadata": {
      "name": "nginx"
    }
  }
}
```

This is harmless, but also useless, and might indicate a copy-paste mistake, so I think the `bugs` category is apt.

anderseknert:
As the name implies, flag duplicate rules when found. See the docs included in the PR for more details, including current limitations.

This rule should be an aggregate rule, but it was hairy enough to pull off with a single file scope, so I decided to hold off on that for now. This provides value already, so let's ship it and consider looking into making this aggregate later.

Fixes #427

### DIFF
--- a/bundle/regal/rules/bugs/duplicate_rule.rego
+++ b/bundle/regal/rules/bugs/duplicate_rule.rego
@@ -1,0 +1,69 @@
+# METADATA
+# description: Duplicate rule
+package regal.rules.bugs["duplicate-rule"]
+
+import rego.v1
+
+import data.regal.result
+import data.regal.util
+
+report contains violation if {
+	some indices in duplicates
+
+	first := indices[0]
+	rest := array.slice(indices, 1, count(indices))
+
+	dup_locations := [location |
+		some index in rest
+		location := input.rules[index].location
+	]
+
+	violation := result.fail(rego.metadata.chain(), object.union(
+		result.location(input.rules[first]),
+		{"description": message(dup_locations)},
+	))
+}
+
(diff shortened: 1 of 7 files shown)


bundle/regal/rules/bugs/duplicate_rule.rego
```diff
@@ -0,0 +1,69 @@
...
+	count(locations) > 1
+}
+
+rules_as_text := [base64.decode(rule.location.text) | some rule in input.rules]
+
+duplicates contains indices if {
+	# Remove whitespace from textual representation of rule and create a hash from the result.
+	# This provides a decent, and importantly *cheap*, approximation of duplicates. We can then
+	# parse the text of these suspected duplicate rules to get a more exact result.
+	rules_hashed := [crypto.md5(regex.replace(text, `\s`, "")) | some text in rules_as_text]
```
charlieegan3, line 45:
```suggestion
	rules_hashed := [crypto.md5(regex.replace(text, `\s+`, "")) | some text in rules_as_text]
```

Might be a tiny bit faster?

charlieegan3 (approved):
Looks great!

bundle/regal/rules/bugs/duplicate_rule.rego
anderseknert: Alright :)

PR #532, opened 2024-01-18
Regal Language Server & VS Code Client

charlieegan3:
This PR contains an initial implementation of a [Language Server](https://microsoft.github.io/language-server-protocol/) for Regal.

To test this pr:

Build and install Regal (the client depends on the regal binary having this functionality)

```
gh pr checkout 10850
go build
mv regal $(which regal)
```

Open `clients/language-server/visual-studio-code` in VS Code.

In the terminal for VS Code, run `npm install`.

Now start an instance of VS Code with the client running:

Screenshot 2024-01-18 at 11 15 21

You can restart the test client with cmd+r.

You can view the server output in the output pane of the spawned test client.

Screenshot 2024-01-18 at 11 17 10


internal/lsp/server.go
```diff
@@ -0,0 +1,522 @@
+package lsp
```
anderseknert, line 1: Wow! Amazing attention to detail in this file 👏

pkg/linter/linter.go
```diff
@@ -755,6 +755,10 @@ func (l Linter) lintWithRegoAggregateRules(
 		return report.Report{}, fmt.Errorf("failed to convert result set to report: %w", err)
 	}
 
+	for i := range result.Violations {
+		result.Violations[i].IsAggregate = true
```
anderseknert, line 759: I don't think I understand this. Why is all violations marked as aggregate?

anderseknert (commented):
This is AWESOME! ⭐ Fantastic work on this, and I can't wait to have this integrated in editors.

I've done a first review pass but I'll likely need another one tomorrow, as it was a lot to unpack. 

Seems like some tests failing,so that'd be good to sort out.

Also, could we move the `clients` directory under `build`? I rather not have a new top level directory if we can avoid it. I know `clients` may not be entirely related to "build", but it's not entirely unrelated either :P

cmd/languageserver.go
```diff
@@ -0,0 +1,59 @@
...
+
+func init() {
+	languageServerCommand := &cobra.Command{
+		Use:   "language-server",
+		Short: "Run the Regal Language Server",
+		Long:  `Start the Regal Language Server and listen on stdin/stdout for client editor messages.`,
+
+		RunE: wrapProfiling(func(args []string) error {
+
+			ctx, cancel := context.WithCancel(context.Background())
```
srenatus, line 24: [nit] It's a little uncommon to not "just" do `defer cancel()`. I see that you deliberately call `cancel()` in the two select cases below; but is there a reason not to call it on the happy path, too? If so, `defer cancel()` should cover all our exits, I think.

internal/lsp/server.go
```diff
@@ -0,0 +1,522 @@
...
+
+	return nil, &jsonrpc2.Error{Code: jsonrpc2.CodeMethodNotFound, Message: fmt.Sprintf("method not supported: %s", req.Method)}
+}
+
+func (l *LanguageServer) SetConn(conn *jsonrpc2.Conn) {
+	l.conn = conn
+}
+
+func (l *LanguageServer) StartDiagnosticsWorker(ctx context.Context) {
+	go func() {
```
srenatus, line 94: What's the benefit of having a go-func-only method here, as opposed to calling
```go
go l.StartDiagnosticsWorker(ctx)
```
wherever we call `l.StartDiagnosticsWorker(ctx)` now? 🤔

internal/lsp/server.go
```diff
@@ -0,0 +1,522 @@
...
+}
+
+func (l *LanguageServer) handleInitialized(ctx context.Context, conn *jsonrpc2.Conn, req *jsonrpc2.Request) (result interface{}, err error) {
+
+	l.diagnosticRequestWorkspace <- "initialized"
+
+	return nil, nil
+}
+
+func (l *LanguageServer) handleTextDocumentDiagnostic(ctx context.Context, conn *jsonrpc2.Conn, req *jsonrpc2.Request) (result interface{}, err error) {
```
srenatus, line 464: [nit]
```suggestion
func (*LanguageServer) handleTextDocumentDiagnostic(context.Context, *jsonrpc2.Conn, *jsonrpc2.Request) (any, error) {
```

internal/lsp/stdio.go
```diff
@@ -0,0 +1,20 @@
...
+
+import "os"
+
+type StdOutReadWriteCloser struct{}
+
+func (StdOutReadWriteCloser) Read(p []byte) (int, error) {
+	return os.Stdin.Read(p)
+}
+
+func (c StdOutReadWriteCloser) Write(p []byte) (int, error) {
```
srenatus, line 11:
```suggestion
func (StdOutReadWriteCloser) Write(p []byte) (int, error) {
```

internal/lsp/stdio.go
```diff
@@ -0,0 +1,20 @@
...
+	return os.Stdout.Write(p)
+}
+
+func (c StdOutReadWriteCloser) Close() error {
```
srenatus, line 15:
```suggestion
func (StdOutReadWriteCloser) Close() error {
```

internal/lsp/stdio.go
```diff
@@ -0,0 +1,20 @@
...
+	if err := os.Stdin.Close(); err != nil {
+		return err
+	}
+	return os.Stdout.Close()
+}
```
srenatus, line 20: When is it required to _Close()_ os.Stdin and os.Stdout? Genuinely curious 😅

srenatus (commented):
Tremendous! I've got a couple of nitpicks on the Go code that feel silly compared to this huge effort 👏

internal/lsp/server.go
anderseknert: Hmm, I wonder if the linter could be configured for this preference?

srenatus: https://github.com/StyraInc/regal/pull/534 Yup!

pkg/linter/linter.go
charlieegan3: Since it's in `lintWithRegoAggregateRules`, and rule with violations here is an Aggregate rule. I use this for some optimisations/categorizations in the language server on when to re-run a full lint so knowing if the violation is of this type is useful. I guess the other option would be to label aggregate rules in their definitions.

internal/lsp/server.go
charlieegan3: Thanks, this should be all fixed now!

internal/lsp/stdio.go
charlieegan3: Hmm, this was based on another LSP implementation. Upon further research it seems like it's not a good idea to close stdout or in, so I have removed this and made the function a nop.

cmd/languageserver.go
charlieegan3: hmm, that's a good spot. I have removed.

internal/lsp/messages.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+type WorkspaceDiagnosticReport struct {
+	Items []WorkspaceFullDocumentDiagnosticReport `json:"items"`
+}
+
+type WorkspaceFullDocumentDiagnosticReport struct {
+	URI     string       `json:"uri"`
+	Version *uint        `json:"version"`
+	Kind    string       `json:"kind"` // Full
```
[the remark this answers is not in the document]
charlieegan3: I am clueless! It's gone now.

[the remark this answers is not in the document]
charlieegan3: I have added a short note in 53c29644d8fb50ab8a3a2b13274f6bc794fb4f82

cmd/languageserver.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+func init() {
+	languageServerCommand := &cobra.Command{
+		Use:   "language-server",
+		Short: "Run the Regal Language Server",
+		Long:  `Start the Regal Language Server and listen on stdin/stdout for client editor messages.`,
+
+		RunE: wrapProfiling(func(args []string) error {
+
```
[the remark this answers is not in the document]
charlieegan3: I hadn't looked at any of the linter errors, this was one of them. Sorry to get you all to be my human liners on this one. It's fixed now.

clients/language-server/visual-studio-code/tsconfig.json
```diff
@@ -0,0 +1,21 @@
+{
```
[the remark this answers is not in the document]
charlieegan3: I think since we plan to get this into the existing OPA client, we can remove the client all together.

clients/language-server/visual-studio-code/package.json
```diff
@@ -0,0 +1,60 @@
...
+		"url": "https://github.com/StyraInc/regal"
+	},
+	"engines": {
+		"vscode": "^1.75.0"
+	},
+	"activationEvents": [
+		"onLanguage:rego"
+	],
+	"main": "./client/out/extension",
+	"contributes": {
```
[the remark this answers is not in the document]
charlieegan3: same as https://github.com/StyraInc/regal/pull/532#discussion_r1466579402

charlieegan3:
Hi both, sorry to take so long to get back to the reviews here. I have made all the updates now and am happier with how I works. I think that next steps are to:

* remove the client implementation here and see how it'd fit into the OPA vscode ext
* consider merging this in an alpha state somehow, or hiding the command. I guess the other option is to leave this open, but it'd be nice not to have it waiting while we work on the client side of this.

Open to ideas!

clients/language-server/visual-studio-code/tsconfig.json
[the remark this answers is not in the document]
anderseknert: Right — I'm suggesting we remove the server parts, not the client. We're not building a server as part of this.

clients/language-server/visual-studio-code/package.json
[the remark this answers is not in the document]
anderseknert: I meant, are we using the things we declare in the "contributes" object?

clients/language-server/visual-studio-code/tsconfig.json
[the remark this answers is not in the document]
anderseknert: I guess what I was trying to get at was that we have an almost identical file in the `client` dir. Perhaps we wouldn't need to have one file coordinating the build of both a client and a server if we're only building the client in JS/TS?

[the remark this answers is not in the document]
charlieegan3: Sure, but I think that most of this will end up going into the OPA extension instead right? Since it's a generic language server, I hope that we can extend the OPA ext with generic use of $some_lang_server, including Rego? Or is that not what you were expecting?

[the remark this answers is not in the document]
charlieegan3: 👍 I'll drop the client from this PR and we can catch up next week!

charlieegan3:
Thanks Anders!

PR #536, opened 2024-02-01
OPA v0.61.0

anderseknert:


srenatus (approved):
Where's dependabot? 👀

srenatus:
image

There's nothing to be seen about the warning, though, when clicking the link 🤷

anderseknert:
Hmm, I had not considered the absence of dependabot 🤔 It should be configured correctly as we haven't touched that since we started out. I guess it doesn't matter too much in this repo, but still. Annoying.

PR #541, opened 2024-02-07
docs: add workflow to auto update docs

charlieegan3:

### DIFF
--- a/.github/workflows/update-docs.yaml
+++ b/.github/workflows/update-docs.yaml
@@ -31,7 +31,7 @@ jobs:
             -H 'Authorization: Bearer ${{ steps.generate_token.outputs.token }}' \
             -H 'X-Github-Api-Version: 2022-11-28' \
             -H 'Content-Type: application/json; charset=utf-8' \
-            -d $'{
+            -d '{
               "ref": "main",
               "inputs": {
                 "version": "latest"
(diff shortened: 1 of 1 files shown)


.github/workflows/update-docs.yaml
```diff
@@ -0,0 +1,39 @@
...
+          private_key: ${{ secrets.PLATFORM_AUTOMATION_GH_APP_PEM_KEY }}
+
+      - name: Trigger Update in docs repo
+        run: |
+          curl -X "POST" "https://api.github.com/repos/StyraInc/docs/actions/workflows/update-regal.yaml/dispatches" \
+            -H 'Accept: application/vnd.github+json' \
+            -H 'Authorization: Bearer ${{ steps.generate_token.outputs.token }}' \
+            -H 'X-Github-Api-Version: 2022-11-28' \
+            -H 'Content-Type: application/json; charset=utf-8' \
+            -d $'{
```
srenatus, line 34: What does the `$` do here? 🤔

charlieegan3: I actually think in this case it's not doing anything, but if there were `\n`s in there, they'd be handled as new lines. I think, just a bash thing as I understand it.

anderseknert (approved):

.github/workflows/update-docs.yaml
charlieegan3: I have removed it, it's not needed. It's just set in the generated commands from my REST gui thingy.

srenatus: Interesting! I'll have to read up on that.

PR #554, opened 2024-02-13
Fix golangci-lint violations

anderseknert:
Not sure if these are from a local upgrade or we're not running this as we think we are in CI. Some rules here are definitely new, but not all of them. I guess existing linters could have seen fixes as well. Will look into it.

### DIFF
--- a/cmd/languageserver.go
+++ b/cmd/languageserver.go
@@ -19,7 +19,7 @@ func init() {
 		Short: "Run the Regal Language Server",
 		Long:  `Start the Regal Language Server and listen on stdin/stdout for client editor messages.`,
 
-		RunE: wrapProfiling(func(args []string) error {
+		RunE: wrapProfiling(func([]string) error {
 			ctx, cancel := context.WithCancel(context.Background())
 			defer cancel()
 
--- a/cmd/new.go
+++ b/cmd/new.go
@@ -65,31 +65,31 @@ Example:
 
 regal new rule --type custom --category naming --name camel-case`,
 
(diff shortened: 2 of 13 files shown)


cmd/languageserver.go
```diff
@@ -19,7 +19,7 @@ func init() {
 		Short: "Run the Regal Language Server",
 		Long:  `Start the Regal Language Server and listen on stdin/stdout for client editor messages.`,
 
-		RunE: wrapProfiling(func(args []string) error {
+		RunE: wrapProfiling(func(_args []string) error {
```
srenatus, line 22: I'd just do `func([]string)` in this and the similar cases...

anderseknert: Maybe it's obvious enough in this case? But I did that first and it just looked like something I'd wonder about 3 months later.

cmd/new.go
```diff
@@ -65,31 +65,31 @@ Example:
 
 regal new rule --type custom --category naming --name camel-case`,
 
-		PreRunE: func(cmd *cobra.Command, args []string) error {
+		PreRunE: func(_ *cobra.Command, _args []string) error {
```
srenatus, line 68:
```suggestion
		PreRunE: func(*cobra.Command, []string) error {
```

srenatus (approved):
LGTM. I'd kill the _arg names, but if you think they're helpful, fine with me 🙃

cmd/languageserver.go
anderseknert: Alright, I caved in to your demands :)

PR #555, opened 2024-02-14
Rule: `with-outside-test-context`

### LINKED ISSUE #548: Rule: `with-outside-test`
Using `with` outside tests has some valid use cases, and we even do that in Regal occasionally. But I think it falls under the "people who know what they're doing" category, and arguably, Regal should help those who don't yet entirely know what they're doing to enlightenment. For those who know what they're doing, disabling this rule either entirely or per location is easy enough.

Using `with` as a _generic way to rewrite input or data_ is what we'd want to avoid here, as there are often better ways of doing the same thing locally (like using functions), and `with` comes with some serious performance implications as things like rule result caching goes out the window.

I'm filing this in the `custom` category for now as it's not clear we'll want to have this enabled by default. If we don't however, there's less of a chance that it's useful to new developers as they might not discover it... so we'll have to weigh the pros and cons on this before we include it.

anderseknert:
See docs in PR for description.

This also introduces a new `performance` category.

Fixes #548

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,74 +188,75 @@ The following rules are currently available:
 
 <!-- RULES_TABLE_START -->
 
-| Category  |                                                Title                                                |                        Description                        |
-|-----------|-----------------------------------------------------------------------------------------------------|-----------------------------------------------------------|
-| bugs      | [constant-condition](https://docs.styra.com/regal/rules/bugs/constant-condition)                    | Constant condition                                        |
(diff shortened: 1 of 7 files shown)


charlieegan3 (approved):
I didn't know about this!

anderseknert:
@charlieegan3 I knew with wasn’t very performant, but it was an example from @robhafner that made me realize just how problematic it can be. Good that get that knowledge into Regal!

PR #557, opened 2024-02-14
rules: Add double negative rule

charlieegan3:
@anderseknert this is 'your' rule so feel free to rebase this branch and take the commit.


anderseknert (commented):
LGTM! But thinking about it I think this is better kept in the `custom` category and not enabled by default. It's a bit pedantic even though I personally like it. WDYT?

charlieegan3:
Maybe... I think it's enough of a smell to leave it in though. For other custom rules we generally have some clear examples where it'd be a bad idea to turn on but I can't think of them here. not not_foo is just always bad as I see it. Maybe we'll have some community members with some good counter examples, but for now I think it's good enough to have in the default list.

anderseknert:
Alright, perhaps you're right. I'll have some time to ponder this before the next release :)

anderseknert (approved):

PR #562, opened 2024-02-20
Docs: Make it obvious how to ignore rules and files

### LINKED ISSUE #561: Clarify the `[rule].ignore .files` option in the documentation
It is not clear from the documentation that the you can add [rule].ignore.files to ignore files for a specifc rule. It is only clear that you can ignore all rules for specific files and that you can ignore files in opa-fmt. https://docs.styra.com/regal#configuration

anderseknert:
As reported by @bdumpp

Fixes #561

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,8 +318,8 @@ rules:
       # not needed as error is the default, but
       # being explicit won't hurt
       level: error
-      # Files can be ignored.
-      # In this example, test files are ignored
+      # files can be ignored for any individual rule
+      # in this example, test files are ignored
       ignore:
         files:
           - "*_test.rego"
@@ -332,6 +332,7 @@ rules:
         - pattern: '^acmecorp\.[a-z_\.]+$|^system\.[a-z_\.]+$'
           targets:
             - package
+
 capabilities:
   from:
     # optionally configure Regal to target a specific version of OPA
@@ -342,6 +343,7 @@ capabilities:
     # version of OPA available at the time of the Regal release
     engine: opa
     version: v0.58.0
+
 ignore:
   # files can be excluded from all lint rules according to glob-patterns
   files:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -356,6 +358,96 @@ reached. If no configuration file is found, Regal will use the default configura
...
+
+### Ignoring a Rule Entirely
+
+If you want to ignore a rule entirely, set its level to `ignore` in the configuration file:
+
+```yaml
+rules:
+  style:
+    prefer-snake-case:
+      # Use camel case to comply with our naming conventions
```
charlieegan3, line 375: idea: I might use an example company name here, e.g.

```suggestion
      # At example.com, we use camel case to comply with our naming conventions
```

I feel this might make 'our' clearer, but feel free to ignore this.

README.md
```diff
@@ -356,6 +358,96 @@ reached. If no configuration file is found, Regal will use the default configura
...
+
+If you'd like to ignore a specific violation in a file, you can add an ignore directive above the line in question, or
+alternatively on the same line to the right of the expression:
+
+```rego
+package policy
+
+import rego.v1
+
+# regal ignore:prefer-snake-case
```
charlieegan3, line 420: Following the suggestion above ('providing a comment on ignored rules is a good way to communicate why the decision was made.'), perhaps that'd make sense here too.


```suggestion
# At example.com, this variable needs to be camelCase because we like camels.
# regal ignore:prefer-snake-case
```

Or similar.

README.md
```diff
@@ -356,6 +358,96 @@ reached. If no configuration file is found, Regal will use the default configura
...
+Note that at this point in time, Regal only considers the same line or the line following the ignore directive, i.e. it
+does not apply to entire blocks of code (like rules, functions or even packages). See [configuration](#configuration)
+if you want to ignore certain rules altogether.
+
+### Ignoring Rules via CLI Flags
+
+For development, rules, or classes of rules, may quickly be enabled or disabled using the relevant CLI flags for the
+`regal lint` command:
+
+- `--disable-all` disables **all** rules
```
charlieegan3, line 441: just an idea, but might make it clear why this flag exists.

```suggestion
- `--disable-all` disables **all** rules to allow specific use of `--enable`
```


I might consider moving this to the end, as it makes more sense _after_ reading about `--enable` etc... 

Or maybe they all only make sense when you know all the options... 🤔

README.md
```diff
@@ -356,6 +358,96 @@ reached. If no configuration file is found, Regal will use the default configura
...
+- `--disable-category` disables all rules in a category, overriding `--enable-all` (may be repeated)
+- `--disable` disables a specific rule, overriding `--enable-all` and `--enable-category` (may be repeated)
+- `--enable-all` enables **all** rules
+- `--enable-category` enables all rules in a category, overriding `--disable-all` (may be repeated)
+- `--enable` enables a specific rule, overriding `--disable-all` and `--disable-category` (may be repeated)
+- `--ignore-files` ignores files using glob patterns, overriding `ignore` in the config file (may be repeated)
+
+All CLI flags override configuration provided in file.
```
charlieegan3, line 449:
```suggestion
**Note** All CLI flags override configuration provided in file.
```

README.md
```diff
@@ -356,6 +358,96 @@ reached. If no configuration file is found, Regal will use the default configura
...
+
```
charlieegan3, line 450: Would a worked example showing how `--disable-category` and `--enable` might work together be helpful here? 🤔 Just to help show how these flags must be composed.

README.md
```diff
@@ -356,6 +358,96 @@ reached. If no configuration file is found, Regal will use the default configura
...
+```
+
+The format of an ignore directive is `regal ignore:<rule-name>,<rule-name>...`, where `<rule-name>` is the name of the
+rule to ignore. Multiple rules may be added to the same ignore directive, separated by commas.
+
+Note that at this point in time, Regal only considers the same line or the line following the ignore directive, i.e. it
+does not apply to entire blocks of code (like rules, functions or even packages). See [configuration](#configuration)
+if you want to ignore certain rules altogether.
+
+### Ignoring Rules via CLI Flags
```
charlieegan3, line 436:
```suggestion
### Ignoring Rules via CLI Flags for Testing & Development
```

Just an idea, feel free to ignore.

charlieegan3 (approved):
Just some ideas, please feel free to skip over any you don't think make sense.

README.md
anderseknert: 👍

README.md
anderseknert: 👍

anderseknert:
Thanks for the suggestions! I applied a few of them but left the CLI args be as we explain the purpose in the first paragraph and I think that should be enough.

PR #564, opened 2024-02-20
Fix false positive in `prefer-some-in-iteration`

### LINKED ISSUE #553: `prefer-some-in-iteration` false positive in set entries list
Doing something like the below should likely be a linter issue in its own right :) But Regal should definitely not recommend using `some ... in` style iteration here:

```rego
package p

my_set := {
    false,
    input.arr[_] == "bar"
}
```

(Obviously not limited to set items, but any similar context)

anderseknert:
Don't flag iteration inside of arrays, sets or objects as using `some` in that context isn't possible.

Fixes #553

### DIFF
--- a/bundle/regal/rules/style/prefer_some_in_iteration.rego
+++ b/bundle/regal/rules/style/prefer_some_in_iteration.rego
@@ -11,11 +11,11 @@ import data.regal.result
 cfg := config.for_rule("style", "prefer-some-in-iteration")
 
 report contains violation if {
-	some rule in input.rules
+	some i, rule in input.rules
 
 	node := filter_top_level_ref(rule)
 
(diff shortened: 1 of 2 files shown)


charlieegan3 (approved):
This looks good to me, but just wanted to check that in comprehensions like this:

```
package p

import rego.v1

my_set := {e|
  e := input.arr[_]
  e == 2
}
```

Would/should still be a violation right? (and that this change won't exclude these)

anderseknert:
Yes, that'll still work, as the node type would be `setcomprehension` rather than `set`. I tested manually to confirm but will add a unit test in the next PR as I'm fixing https://github.com/StyraInc/regal/issues/425 next (which was very much related).

PR #569, opened 2024-02-21
lsp: Correctly handle URIs and paths on Windows

charlieegan3:
I have been doing some testing on Windows and found that the URI formats representing files and identifying them work in different ways. This PR contains changes to the language server implementation that allow Regal LS to work with a VS code client on windows.

Screenshot 2024-02-21 at 10 59 49

Namely, this PR adds a new package with functions to perform the required translations on URI values needed to operate in the context of each supported client - currently VS code only.

### DIFF
--- a/internal/lsp/cache.go
+++ b/internal/lsp/cache.go
@@ -2,7 +2,6 @@ package lsp
 
 import (
 	"fmt"
-	"net/url"
 	"os"
 	"sync"
 
@@ -196,17 +195,8 @@ func (c *Cache) Delete(uri string) {
 	c.diagnosticsParseMu.Unlock()
 }
 
-func updateCacheForURIFromDisk(cache *Cache, uri string) (string, error) {
-	parsedURI, err := url.Parse(uri)
-	if err != nil {
-		return "", fmt.Errorf("failed to parse URI: %w", err)
-	}
-
-	if parsedURI.Scheme != "file" {
-		return "", fmt.Errorf("only file:// URIs are supported, got %q", parsedURI.String())
-	}
-
-	content, err := os.ReadFile(parsedURI.Path)
+func updateCacheForURIFromDisk(cache *Cache, uri, path string) (string, error) {
+	content, err := os.ReadFile(path)
 	if err != nil {
 		return "", fmt.Errorf("failed to read file: %w", err)
 	}
--- a/internal/lsp/messages.go
+++ b/internal/lsp/messages.go
@@ -120,6 +120,15 @@ type TextDocumentIdentifier struct {
 	URI string `json:"uri"`
 }
 
+type TextDocumentDidChangeParams struct {
+	TextDocument   TextDocumentIdentifier           `json:"textDocument"`
+	ContentChanges []TextDocumentContentChangeEvent `json:"contentChanges"`
+}
+
+type TextDocumentContentChangeEvent struct {
+	Text string `json:"text"`
+}
+
 type Diagnostic struct {
 	Range    Range          `json:"range"`
 	Message  string         `json:"message"`
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -421,11 +421,10 @@ func (l *LanguageServer) handleInitialize(
 	l.clientRootURI = params.RootURI
 	l.clientIdentifier = clients.DetermineClientIdentifier(params.ClientInfo.Name)
 
-	if l.clientIdentifier == clients.IdentifierUnknown {
-		l.logError(fmt.Errorf(
-			"unable to match client identifier for initializing client: %s",
-			params.ClientInfo.Name,
-		))
+	if l.clientIdentifier == clients.IdentifierGeneric {
+		l.log(
+			"Unable to match client identifier for initializing client, using generic functionality: " + params.ClientInfo.Name,
+		)
 	}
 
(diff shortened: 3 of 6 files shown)


internal/lsp/cache.go
```diff
@@ -196,17 +195,8 @@ func (c *Cache) Delete(uri string) {
...
-	if err != nil {
-		return "", fmt.Errorf("failed to parse URI: %w", err)
-	}
-
-	if parsedURI.Scheme != "file" {
-		return "", fmt.Errorf("only file:// URIs are supported, got %q", parsedURI.String())
-	}
-
-	content, err := os.ReadFile(parsedURI.Path)
+func updateCacheForURIFromDisk(cache *Cache, uri, path string) (string, error) {
```
charlieegan3, line 198: updateCacheForURIFromDisk has been spared knowledge of the URI path translations and so the path must now be provided by the caller.

internal/lsp/messages.go
```diff
@@ -120,6 +120,15 @@ type TextDocumentIdentifier struct {
 	URI string `json:"uri"`
 }
 
+type TextDocumentDidChangeParams struct {
```
charlieegan3, line 123: This type represents a messages that is sent in the protocol and so should have been here to begin with. it has been moved.

internal/lsp/server.go
```diff
@@ -51,7 +52,16 @@ type LanguageServer struct {
...
 	diagnosticRequestWorkspace chan string
 
-	clientRootURI string
+	clientRootURI    string
+	clientIdentifier clients.Identifier
+}
+
+// fileDiagnosticRequiredEvent is sent to the diagnosticRequestFile channel when
+// diagnostics are required for a file.
+type fileDiagnosticRequiredEvent struct {
```
charlieegan3, line 61: This type has just been moved up in the file.

internal/lsp/uri/uri_test.go
```diff
@@ -0,0 +1,93 @@
...
+			want: "file:///c%3A/foo/bar",
+		},
+	}
+
+	for label, tc := range testCases {
+		tt := tc
+
+		t.Run(label, func(t *testing.T) {
+			t.Parallel()
+			got := FromPath(clients.IdentifierVSCode, tt.path)
```
charlieegan3, line 39: This will need to change or be adjusted as we add more clients...

charlieegan3:
When this is done, we can get this in https://github.com/open-policy-agent/vscode-opa/pull/98 and release a new version of the extension too.

internal/lsp/server.go
```diff
@@ -418,6 +419,14 @@ func (l *LanguageServer) handleInitialize(
 	}
 
 	l.clientRootURI = params.RootURI
+	l.clientIdentifier = clients.DetermineClientIdentifier(params.ClientInfo.Name)
+
+	if l.clientIdentifier == clients.IdentifierUnknown {
+		l.logError(fmt.Errorf(
```
anderseknert, line 425: This doesn't seem like an error to me in a supposedly open protocol :) Can this just be a notice?

internal/lsp/server.go
```diff
@@ -453,36 +462,36 @@ func (l *LanguageServer) handleInitialize(
 		},
 	}
 
-	folderURI, err := url.Parse(l.clientRootURI)
-	if err != nil {
-		return nil, fmt.Errorf("failed to parse URI: %w", err)
-	}
+	workspaceRootPath := uri.ToPath(l.clientIdentifier, l.clientRootURI)
+	l.logError(fmt.Errorf("workspaceRoot: %s", workspaceRootPath))
```
anderseknert, line 466: Or wait, you're logging at error level here too. Is that all we got?

anderseknert (approved):
Good catch! Attention to detail 💯 It'll still work for those trying to connect to Regal from Neovim or whatnot, right? Having hard coded client types feels ugly, but I understand it might be needed to deal with cases like this. I hope we'll get to "normalize" some of this once we try another client implementation later 🙂

internal/lsp/server.go
charlieegan3: Ahh, that should be an error. fixed in 1d516dcaacc3aecb3cf1640c555d1c11fc2c8ca9

internal/lsp/server.go
charlieegan3: we have log too, it goes to std err, since std in is reserved for the JSON RPC data.

charlieegan3: This isn't needed so have removed in 1d516dcaacc3aecb3cf1640c555d1c11fc2c8ca9

charlieegan3:
> It'll still work for those trying to connect to Regal from Neovim or whatnot, right?

I have added some tests now to show the default behaviour makes sense too. However, I don't think that we'll know what idiosyncrasies we'll see from other clients until we implement them. I am confident that we can make it all work though since the differences are likely to be minimal.

anderseknert:
All good! Feel free to merge when you want 👍

charlieegan3:
Once this is in a release, we can update https://github.com/open-policy-agent/vscode-opa/pull/98/files#diff-1944111a462efb504e2d39ee24a9cb532c03164ada37caada104fecc06d874bdR23 and then I think the ext is ready to go too.

charlieegan3:
Amazing, that sounds great!

PR #574, opened 2024-02-22
Fix `prefer-some-in-iteration` false positive on contains check

### LINKED ISSUE #566: `prefer-some-in-iteration`: false positive on contains check
This one is similar to other issues addressed recently. While we're generally too forgiving on the use of `=`, as unification is tricky to deal with statically, this one should not present that much of a problem, I think:

```rego
package p

import rego.v1

r if {
    "x" = input[_]
}
```

Using `some .. in` should not be recommended in this case, at least not if there's a constant on one side. For the case of variables we can keep the recommendation as we might not be sure whether the var is or will be bound later, or at least not without significant effort, which I don't think this warrants at the moment.

anderseknert:
When scalar values are on one side of e.g. `5 == input[_]` we should not recommending using `some`, but `in`.

Fixes #566

### DIFF
--- a/bundle/regal/rules/style/prefer_some_in_iteration.rego
+++ b/bundle/regal/rules/style/prefer_some_in_iteration.rego
@@ -73,3 +73,14 @@ invalid_some_context(rule, path) if {
 impossible_some(node) if node.type in {"array", "object", "set"}
 
 impossible_some(node) if node.key
+
+# technically this is not an _impossible_ some, as we could replace e.g. `"x" == input[_]`
+# with `some "x" in input`, but that'd be an `unnecessary-some` violation as `"x" in input`
+# would be the correct way to express that
+impossible_some(node) if {
+	node.terms[0].value[0].type == "var"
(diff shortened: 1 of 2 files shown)


bundle/regal/rules/style/prefer_some_in_iteration.rego
```diff
@@ -73,3 +73,14 @@ invalid_some_context(rule, path) if {
 impossible_some(node) if node.type in {"array", "object", "set"}
 
 impossible_some(node) if node.key
+
+# technically this is not an _impossible_ some, as we could replace e.g. `x == input[_]`
```
srenatus, line 77:
```suggestion
# technically this is not an _impossible_ some, as we could replace e.g. `"x" == input[_]`
```

srenatus (approved):

bundle/regal/rules/style/prefer_some_in_iteration.rego
anderseknert: Good catch!

PR #577, opened 2024-02-23
Rule: `ignored-import`

### LINKED ISSUE #540: Rule: `ignored-import`
```rego
package p

import rego.v1

import data.x.y.z

z := data.x.y.z
```
There are probably better names, and certainly better examples, but essentially — flag whenever an import has been declared and the same ref is used explicitly somewhere instead of using the import.

anderseknert:
Flagging any reference in a policy that does not make use of an import where applicable.

Fixes #540

### DIFF
--- a/bundle/regal/config/config_test.rego
+++ b/bundle/regal/config/config_test.rego
@@ -161,7 +161,7 @@ test_all_rules_are_in_provided_configuration if {
 		some category, title
 		data.regal.rules[category][title]
 		not endswith(title, "_test")
-		not data.regal.config.provided.rules[category][title]
+		not config.provided.rules[category][title]
 	}
 
 	count(missing_config) == 0
@@ -172,7 +172,7 @@ test_all_configured_rules_exist if {
 
 	missing_rules := {title |
 		some category, title
-		data.regal.config.provided.rules[category][title]
+		config.provided.rules[category][title]
 		not data.regal.rules[category][title]
 	}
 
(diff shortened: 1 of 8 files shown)


bundle/regal/config/config_test.rego
```diff
@@ -161,7 +161,7 @@ test_all_rules_are_in_provided_configuration if {
 		some category, title
 		data.regal.rules[category][title]
 		not endswith(title, "_test")
-		not data.regal.config.provided.rules[category][title]
+		not config.provided.rules[category][title]
```
anderseknert, line 164: Incredibly satisfying when new rules find issues or improvement opportunities in Regal 😍

bundle/regal/rules/imports/ignored_import.rego
```diff
@@ -0,0 +1,34 @@
...
+	some ref in ast.all_rules_refs
+
+	ref.value[0].type == "var"
+	ref.value[0].value in {"data", "input"}
+
+	most_specific_match := regal.last(sort([ip |
+		ref_path := [p.value | some p in ref.value]
+
+		some ip in import_paths
+		array.slice(ref_path, 0, count(ip)) == ip
```
srenatus, line 27: Took me a few blinks to see what this means. But you want to check if `ref_path` has a prefix that's in `import_paths` ✔️

srenatus (approved):

Issue #528, opened 2024-01-02
Not considering `some` used in the comprehension body when checking the comprehension assignment


nevumx:
`idiomatic/use-some-for-output-vars` and `style/prefer-some-in-iteration` are violated in this playground, even though they shouldn't be: https://play.openpolicyagent.org/p/SQNTZtJ0be

e.g.
```rego
build_obj(params) := result if {
	param_paths := {
		"principal": ["principal"],
		"action": ["action"],
		"entity": ["entity"],
	}
	param_objects := [build_object_path(param_paths[param_key], param_value) | some param_key, param_value in params]
	result := object.union_n(param_objects)
}
```

PR #579, opened 2024-02-23
Fix: `prefer-some-in-iteration` false positive iteration in fn args

anderseknert:
This fixes one of the errors reported in #528


anderseknert:
Meh, checking back at that code snippet from #528 and this did in fact not solve that 😒 Keep digging, I guess, but this is an improvement nonetheless.

charlieegan3 (approved):

anderseknert:
Fixed so now it *does* fix the some .. in false positive as reported in the above mentioned issue.

Issue #578, opened 2024-02-23
Error launching LSP server in Neovim with string type initializationOptions


folliehiyuki:
According to the LSP specs (https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize), `initializationOptions` can be of any types. Regal currently forces the value to be string, leading to a json unmarshall error in Neovim.

https://github.com/StyraInc/regal/blob/3205762160dab68f5766d58446cde74617ce9b9d/internal/lsp/messages.go#L24

Neovim sends the request with `initializationOptions` being an empty object if not explicitly specified, hence the error. An example client request:
```bash
{
  "capabilities": {
    "textDocument": {
      "callHierarchy": {
        "dynamicRegistration": false
      },
      "codeAction": {
        "codeActionLiteralSupport": {
          "codeActionKind": {
            "valueSet": [
              "",
              "quickfix",
              "refactor",
              "refactor.extract",
              "refactor.inline",
              "refactor.rewrite",
              "source",
              "source.organizeImports"
            ]
          }
        },
        "dataSupport": true,
        "dynamicRegistration": false,
        "isPreferredSupport": true,
        "resolveSupport": {
          "properties": [
            "edit"
          ]
        }
      },
      "completion": {
        "completionItem": {
          "commitCharactersSupport": true,
          "deprecatedSupport": true,
          "documentationFormat": [
            "markdown",
            "plaintext"
          ],
          "insertReplaceSupport": true,
          "insertTextModeSupport": {
            "valueSet": [
              1,
              2
            ]
          },
          "labelDetailsSupport": true,
          "preselectSupport": true,
          "resolveSupport": {
            "properties": [
              "documentation",
              "detail",
              "additionalTextEdits",
              "sortText",
              "filterText",
              "insertText",
              "textEdit",
              "insertTextFormat",
              "insertTextMode"
            ]
          },
          "snippetSupport": true,
          "tagSupport": {
            "valueSet": [
              1
            ]
          }
        },
        "completionItemKind": {
          "valueSet": [
            1,
            2,
            3,
            4,
            5,
            6,
            7,
            8,
            9,
            10,
            11,
            12,
            13,
            14,
            15,
            16,
            17,
            18,
            19,
            20,
            21,
            22,
            23,
            24,
            25
          ]
        },
        "completionList": {
          "itemDefaults": [
            "commitCharacters",
            "editRange",
            "insertTextFormat",
            "insertTextMode",
            "data"
          ]
        },
[...]

PR #580, opened 2024-02-26
Language server initialisation adjustments

charlieegan3:
This PR makes two changes:

- https://github.com/StyraInc/regal/issues/578 raised an issue with the unmarshalling of the initializationOptions field. We don't use this field so it has been dropped.
- We an issue when there is no root URI set in the initialize request. This PR disables workspace functions such as aggregate violations and linting of unopened files when this is the case (when a single file has been opened).

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -54,6 +54,10 @@ type LanguageServer struct {
 
 	clientRootURI    string
 	clientIdentifier clients.Identifier
+
+	// workspaceMode is set to true when the ls is initialized with
+	// a clientRootURI.
+	workspaceMode bool
 }
 
 // fileDiagnosticRequiredEvent is sent to the diagnosticRequestFile channel when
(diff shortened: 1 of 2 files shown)


internal/lsp/server.go
```diff
@@ -461,10 +473,31 @@ func (l *LanguageServer) handleInitialize(
...
+		l.log("running in workspace mode")
+		l.workspaceMode = true
+
+		err = l.loadWorkspaceContents()
+		if err != nil {
+			return nil, fmt.Errorf("failed to load workspace contents: %w", err)
+		}
+
+		// attempt to load the config as it is found on disk
+		file, err := config.FindConfig(strings.TrimPrefix(l.clientRootURI, "file://"))
```
anderseknert, line 486: Was file:// not an issue on windows? Or it’s not applicable here?

anderseknert (approved):
Thanks! Just one question.

internal/lsp/server.go
charlieegan3: Hmm, need to double check this. Will save a TODO for this.

PR #581, opened 2024-02-27
Rename `unused-return-value` -> `unassigned-return-value`

### LINKED ISSUE #297: Rename `unused-return-value`
Rename rule `unassigned-return-value` as that better reflects its purpose

For the rename, ensure that we can redirect traffic from the old doc URL to the new.

anderseknert:
Fixes #297


charlieegan3 (approved):
I guess we might also want to put a redirect in for this on the docs site?

anderseknert:
@charlieegan3 totally! Just not sure how to remember that after the release, but let's try :P

Issue #153, opened 2023-06-08
Allow enable/disable all via config file


anderseknert:
Since some users may want to only run a few rules in their repo, having to manually disable all the rules they don't care for is a bit of a chore. It'd be good if we allowed enabling/disabling all, or per category from the config as well.

anderseknert:
Expanding on the scope a bit here as this was requested today, or at least something that is close enough. It'd be good if there was a way to freeze configuration of rules at any given point, so that updates wouldn't necessarily mean new rules automatically enabled. Some prefer to have them enabled manually, and that's obviously fine. I imagine this would look like:

1. Add a command to print the de-facto configuration used, i.e. the built-in default configuration merged with the user's custom config file.
2. Add a mode to Regal (settable via cia CLI args or config file) that have rules not mentioned in config ignored.
3. Log any rules ignored above so that it's easy for a user to see which rules they are missing from having frozen their rules config.

Definitely open to alternative approaches if anyone has suggestions!

anderseknert:
I've been thinking some more about this, and I think a better design would be to mirror the `regal lint` flags for disabling/enabling all rules, or all per category, and allow more specific overrides, but via configuration rather than CLI args. I imagine it might look something like this:

```yaml
rules:
  '*':
    level: ignore
  bugs:
    '*':
      level: error
    constant-condition:
      level: ignore
  testing:
    print-or-trace-call:
      level: error
```

Disabling (i.e. setting level to `ignore` for) all rules at the top level, and then allow more specific overrides to enable (setting level to `warning` or `error` for) categories or rules would provide a way to "freeze" configuration for those who don't want new Regal versions to automatically add new rules, and it would also be a pretty convenient way to easily enable/disable all, categories and individual rules.

I'm not too fond of using `'*'` to mean "all" as the required quoting makes it rather ugly... but I'm not sure what would be a better option. We could call it "all" or "default", but it could easily be mistaken for a category/rule name. `'*'` while ugly at least clearly stands out from regular names. So might be that I go with that, but if anyone has better ideas, I'm listening 🙂

anderseknert:
Fixed in #587

PR #587, opened 2024-03-06
config: Support the defaulting of rules in config

charlieegan3:
This PR implements an approach to rule defaulting described in: https://github.com/StyraInc/regal/issues/153#issuecomment-1946882444


pkg/linter/linter_test.go
```diff
@@ -138,6 +139,69 @@ or := 1
...
+					Categories: map[string]config.Default{
+						"style": {Level: "error"},
+					},
+				},
+				Rules: map[string]config.Category{},
+			},
+			filename:      "p.rego",
+			expViolations: []string{"opa-fmt", "top-level-iteration", "rule-shadows-builtin"},
+			expLevels:     []string{"error", "warning", "warning"},
+		},
```
anderseknert, line 204: Test cases are good! I wonder if we should add something (e2e) for the case of "category disabled via config file, enabled via CLI" or similar... wouldn't need to be extensive but just something to ensure CLI args have precedence.

anderseknert (commented):
Looks great to me! I guess we'll probably undo some of this if we move things over to Rego later, but that goes for a lot of code in our initialization routine, so... :)

One thing I noticed is how there is no documentation to cover this. Adding a section in the configuration section of the README explaining how this works would be good.

anderseknert:
I've done some manual testing and things work just as I'd expect 👍

pkg/linter/linter_test.go
charlieegan3: Yeah I can look into doing that since we don't have that covered elsewhere

pkg/config/config.go
```diff
@@ -15,16 +15,51 @@ import (
...
+)
 
 type Config struct {
 	Rules        map[string]Category `json:"rules"                  yaml:"rules"`
 	Ignore       Ignore              `json:"ignore,omitempty"       yaml:"ignore,omitempty"`
 	Capabilities *Capabilities       `json:"capabilities,omitempty" yaml:"capabilities,omitempty"`
+
+	// Defaults state is loaded from configuration under rules and so is not (un)marshalled
+	// in the same way.
+	Defaults Defaults `json:"-" yaml:"-"`
```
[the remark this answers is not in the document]
charlieegan3: 🤓 all this stuff still feels painful. Let's try a rego based POC of this soon.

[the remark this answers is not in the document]
anderseknert: 100%

README.md
```diff
@@ -361,13 +361,34 @@ provided will be used to override the default configuration.
 
 ## Ignoring Rules
 
-If some rule doesn't align with your team's preferences, don't worry! Regal is not meant to be the law, and some rules
-may not make sense for your project, or parts of it. Regal provides several different ways to ignore rules, either
-entirely, or with more granularity.
+If one of Regal's rules doesn't align with your team's preferences, don't worry! Regal is not meant to be the law,
+and some rules may not make sense for your project, or parts of it.
+Regal provides several different methods to ignore rules with varying precedence.
+The available methods are (ranked High to Lowest precedence):
```
anderseknert, line 367: Mostly just curious about the use of capitalized High and Lowest here! Also, why is it "High" and not "Highest"?

e2e/cli_test.go
```diff
@@ -577,6 +577,80 @@ func TestMergeRuleConfigWithoutLevel(t *testing.T) {
...
+	// this error should not be ignored
+	if !strings.Contains(stdout.String(), "top-level-iteration") {
+		t.Errorf("expected stdout to contain top-level-iteration")
+		t.Log("stdout:\n", stdout.String())
+	}
+
+	expectExitCode(t, err, 3, &stdout, &stderr)
+}
+
+func TestConfigDefaultingWithEnableDirective(t *testing.T) {
```
anderseknert, line 617: This is great!

anderseknert (approved):
LGTM! I just had a question, but it's not blocking, so feel free to merge when you want :)

Thanks! 👏

README.md
charlieegan3: Ha, oops. That went from `High to Low` to `High to Lowest precedence` to `ranked High to Lowest precedence` and I just was too careless to change the case... not that it even needed to be Title Case in the first place! 😂

README.md
```diff
@@ -361,13 +361,34 @@ provided will be used to override the default configuration.
...
 
-### Ignoring a Rule Entirely
+* [Inline Ignore Directives](#inline-ignore-directives) cannot be overridden by any other method.
```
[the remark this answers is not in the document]
charlieegan3: Caught! I am on team `*` but will get that patched up.

PR #592, opened 2024-03-11
Fix false positive in rules not counting imports in scope

### LINKED ISSUE #590: False positive in `top-level-iteration` when a ref var is imported
```rego
package p

import data.x

y := input[x]
```

`x` is not an output var in this case, therefore there's no iteration. But looks like the `top-level-iteration` rule missed that. Not likely a common case but I found one in the wild today :)

anderseknert:
Fixes #590

That issue only mentions `top-level-iteration`, but I found the same behavior in `use-some-for-output-vars`, so this fixes that too.


charlieegan3 (approved):
LGTM!

Issue #585, opened 2024-03-02
Invalid code and severity values in LSP server's response


folliehiyuki:
* In a valid document, the server responds with `code` field being a JSON object containing `value` and `target` keys. The official LSP specs only allows int or string.
* In an invalid document, the server responds with `severity = 0` (default empty int value, I suppose), which is not in the allowed range [1, 4]. The diagnostic message is still displayed fine though.

Ref: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnostic

More details can be seen in the [original comment](https://github.com/StyraInc/regal/issues/578#issuecomment-1973783653).

charlieegan3:
Thanks for opening a new issue for this @folliehiyuki, I'll take a look at this on Tuesday. Have a great weekend.

charlieegan3:
Hi @folliehiyuki, sorry for the radio silence last week. I have had a go at some fixes in https://github.com/StyraInc/regal/pull/594. It seems to work ok in VS Code ok, would you be able to do some testing with a build from this branch to check is works for you use case too?

folliehiyuki:
I don't have access to my development computer now, so let me report back this weekend.

Thanks for working on it!

PR #594, opened 2024-03-12
lsp: Make server messages compliant

charlieegan3:
Fixes https://github.com/StyraInc/regal/issues/585

Changes:
- make sure that errors are 1 not 0s in severity codes
- Use code desc for Href to lint docs

TODO:
- [ ] wait for some confirmation from reporter of #585  that this works ok

### DIFF
--- a/internal/lsp/lint.go
+++ b/internal/lsp/lint.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/open-policy-agent/opa/ast"
 
@@ -53,7 +54,7 @@ func updateParse(cache *Cache, uri string) (bool, error) {
 			}
 
 			diags = append(diags, Diagnostic{
-				Severity: 0, // parse errors are always errors
+				Severity: 1, // parse errors are the only error Diagnostic the server sends
 				Range: Range{
 					Start: Position{
 						Line:      uint(line),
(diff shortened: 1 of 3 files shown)


anderseknert (approved):
👍

srenatus (approved):

anderseknert:
Cutting a release tomorrow, so let's get this included. @folliehiyuki do let us know if this did not resolve the issue and we'll get back to it immediately.

folliehiyuki:
I can confirm the diagnostic messages seem correct now. Here is a few screenshots with my setup.

On a file with found issues:



with syntax error or being an empty file:

folliehiyuki:
Also, I've opened https://github.com/neovim/nvim-lspconfig/pull/3071 to include basic `regal` setup into the common LSP configuration of Neovim.

charlieegan3:
@folliehiyuki thanks so much for checking this over and adding it to the nvim-lspconfig repo too. Much appreciated! 👏 

`nvim-lspconfig` is the default way to use Language Servers from Neovim? I'm just trying to work out the best message for the notes here: https://docs.styra.com/regal/editor-support

folliehiyuki:
> `nvim-lspconfig` is the default way to use Language Servers from Neovim?

Yes, it's a Neovim plugin that most (if not all) people use as a base to configure their LSP servers (and other LSP-related settings).

This is a snippet to enable Regal LSP with default settings, judging that you already have `regal` and the plugin installed:
```lua
require('lspconfig').regal.setup()
```

charlieegan3:
That's great, I think we can get a note in the docs for that! Thanks so much for this info as we're not neovim users it's hard to be sure on some of these points!

charlieegan3:
PR for the docs: https://github.com/StyraInc/regal/pull/604

PR #600, opened 2024-03-14
Add support for editing Regal in VS Code

anderseknert:
Options like `checkOnSave` would previously not work when hacking on Regal, as a custom capabilities file would need to be provided. Using the latest version of the OPA extension, this is now supported via the `opa.env` option.

Note that actually setting `checkOnSave` or similar options is left up to each developer.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,5 +4,3 @@ dist/
 
 /regal
 /regal.exe
-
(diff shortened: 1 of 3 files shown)


charlieegan3 (approved):
👏

PR #609, opened 2024-03-26
Publish binaries from build workflow

anderseknert:
Might be some things to work out, but since we only run the build workflow on PRs or pushes to main, here we are.


anderseknert:
Looks like the build hanged in the matrix conf somehow, while still successful. Will check if a new PR comes with the same issue.

PR #610, opened 2024-03-26
Publish binaries from build workflow #2

### LINKED ISSUE #608: Publish binaries as part of each CI build
It's occasionally useful to be able to quickly download an artifact from a build for testing, or to instruct someone else to do so. Our regular PR builds currently don't produce any artifacts. It would be good if they did so, at least for some time.

anderseknert:
No Windows binaries as of yet since `build/do.rq pull_request` silently fails, or does nothing, on Windows. At least when called from Powershell. The big discovery here however — we likely never actually built on Windows in these worflows before, or even on MacOS! This since the `runs-on` attribute didn't make use of the matrix OS value, but was instead hard-coded to use `ubuntu-latest` 🙃

We'll have to get back to the Windows build at some point, but for now we'll settle on only building on that platform for releases via goreleaser, just as we unknowingly seem to have done up until this point anyway.

Fixes #608

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -14,13 +14,19 @@ on:
 jobs:
   build-matrix:
     name: Matrix
-    runs-on: ubuntu-latest
     strategy:
       matrix:
         os:
-        - ubuntu-latest
-        - windows-latest
-        - macos-latest
+        - runner: ubuntu-latest
+          name: linux
+          static: true
+        - runner: macos-latest
+          name: macos-intel
+          static: false
+        - runner: macos-14
+          name: macos-arm64
+          static: true
+    runs-on: ${{ matrix.os.runner }}
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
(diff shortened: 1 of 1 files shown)


anderseknert:
Also identified and fixed the cause of the hangs where some jobs get stuck in "Expected". I know I've seen that in some repos before, so I'll get back to those to see if that's still a problem now that I know how to fix it.

anderseknert:
Running `build/do.rq pull_request` on the unix runners vs. windows

Screenshot 2024-03-26 at 22 40 25
Screenshot 2024-03-26 at 22 40 10

charlieegan3 (approved):
LGTM, we might also want to set: https://docs.github.com/en/organizations/managing-organization-settings/configuring-the-retention-period-for-github-actions-artifacts-and-logs-in-your-organization

anderseknert:
Yeah, that's a good point. I'll defer it to later though, as I don't really know what to set there. Honestly I'd probably be fine with having anything removed once it's merged into main. Can't really see the point of keeping branch logs / artifacts around after the branch is merged and deleted, and one could always get the same from `main` post merge. Doesn't seem like that's an option though.

PR #617, opened 2024-04-02
lsp: `textDocument/hover` implementation

anderseknert:
This PR adds support for the `textDocument/hover` feature of the language server protocol. Currently, only the built-in functions of OPA are supported, which likely satisfies the most common requirement. In the future, we could extend this to cover also user defined functions or rules, where annotations are used to provide data for the hover response.

Screenshot from VS Code with Regal running from this branch.

Screenshot 2024-04-02 at 13 38 56

### DIFF
--- a/internal/lsp/cache.go
+++ b/internal/lsp/cache.go
@@ -30,6 +30,9 @@ type Cache struct {
 	// diagnosticsParseErrors is a map of file URI to parse errors for that file
 	diagnosticsParseErrors map[string][]Diagnostic
 	diagnosticsParseMu     sync.Mutex
+
+	builtinPositionsFile map[string]map[uint][]BuiltinPosition
+	builtinPositionsMu   sync.Mutex
 }
 
 func NewCache() *Cache {
@@ -40,6 +43,8 @@ func NewCache() *Cache {
 		diagnosticsFile:        make(map[string][]Diagnostic),
 		diagnosticsAggregate:   make(map[string][]Diagnostic),
 		diagnosticsParseErrors: make(map[string][]Diagnostic),
+
+		builtinPositionsFile: make(map[string]map[uint][]BuiltinPosition),
 	}
 }
 
@@ -172,6 +177,22 @@ func (c *Cache) SetParseErrors(uri string, diags []Diagnostic) {
 	c.diagnosticsParseErrors[uri] = diags
 }
 
+func (c *Cache) GetBuiltinPositions(uri string) (map[uint][]BuiltinPosition, bool) {
+	c.builtinPositionsMu.Lock()
+	defer c.builtinPositionsMu.Unlock()
+
+	val, ok := c.builtinPositionsFile[uri]
+
+	return val, ok
+}
+
+func (c *Cache) SetBuiltinPositions(uri string, positions map[uint][]BuiltinPosition) {
+	c.builtinPositionsMu.Lock()
+	defer c.builtinPositionsMu.Unlock()
+
+	c.builtinPositionsFile[uri] = positions
+}
+
 // Delete removes all cached data for a given URI.
 func (c *Cache) Delete(uri string) {
 	c.fileContentsMu.Lock()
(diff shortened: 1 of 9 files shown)


internal/lsp/cache.go
```diff
@@ -30,6 +30,9 @@ type Cache struct {
 	// diagnosticsParseErrors is a map of file URI to parse errors for that file
 	diagnosticsParseErrors map[string][]Diagnostic
 	diagnosticsParseMu     sync.Mutex
+
+	builtinPositionsFile map[string]map[uint][]BuiltinPosition
```
charlieegan3, line 34: 'File' seemed unexpected in this name to me since the index is by URI, and the contents isn't files but rather by line. idk, minor one so feel free to ignore!

internal/lsp/hover.go
```diff
@@ -0,0 +1,174 @@
+package lsp
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/olekukonko/tablewriter"
+
```
charlieegan3, line 8: Is there a reason to break these groups up? If so, we might want to do the same in the server.go too for consistency.

internal/lsp/hover.go
```diff
@@ -0,0 +1,174 @@
...
+func builtinCategory(builtin *ast.Builtin) (category string) {
+	if len(builtin.Categories) == 0 {
+		if s := strings.Split(builtin.Name, "."); len(s) > 1 {
+			category = s[0]
+		} else {
+			category = builtin.Name
+		}
+	} else {
+		category = builtin.Categories[0]
```
charlieegan3, line 39: What are we losing here, presumably the rest of the categories are not useful?

internal/lsp/hover.go
```diff
@@ -0,0 +1,174 @@
...
+			sb.WriteString(n.Name)
+		} else {
+			sb.WriteString(arg.String())
+		}
+	}
+
+	sb.WriteString(")\n```")
+}
+
+func createHoverContent(builtin *ast.Builtin) string {
```
charlieegan3, line 75: It might be nice to have a simple test case for this function to make for easier refactoring. E.g. we might want to make use of a template file or similar in future.

charlieegan3 (approved):
LGTM

It might be simple enough to extend the `TestLanguageServerSingleFileWithConfig` case with a few extra messages for Hover functionality? They can always come in another PR if you'd rather.

internal/lsp/cache.go
anderseknert: Right, but the URI points to a file, no?

internal/lsp/hover.go
anderseknert: We do that in server.go too, don't we? :)

Imports are grouped according to this rule: https://github.com/StyraInc/regal/blob/main/docs/development.md#linting

internal/lsp/hover.go
anderseknert: This is to try and mimic the way categories are placed under headings in the OPA reference docs. The "other side" of this is where it's being generated, i.e. [here](https://github.com/open-policy-agent/opa/blob/main/internal/cmd/genbuiltinmetadata/main.go#L153).

internal/lsp/hover.go
anderseknert: Sounds good. I'll look into adding a couple of tests.

anderseknert:
> It might be simple enough to extend the TestLanguageServerSingleFileWithConfig case with a few extra messages for Hover functionality? They can always come in another PR if you'd rather.

Oof, that's a 300 LOC test 😄 I did add tests for hover content creation, but I think I'll have to defer this one  until later, or I'd spend the whole afternoon on that. But I definitely want to come back to tests later.

internal/lsp/hover.go
charlieegan3: Ahh ok, so OPA is different from other external reps, gotcha.

charlieegan3:
Yeah the tests for the lip functionality are kinda intense. That's on me really but better than nothing.

internal/lsp/cache.go
charlieegan3: Sometimes the URI is a folder, but that doesn't apply to hover/built ins clearly.

anderseknert:
For sure! We can try and break it up later.

Issue #618, opened 2024-04-03
Handle missing module gracefully in hover handler


anderseknert:
If we fail to parse the module, the hover handler will fall back on the latest successfully parsed module, which is good. There's however an edge case I had not considered, which is when a user opens a broken Rego file to begin with.

```rego
package p 

r := http.send({})

# broken, won't parse
foo.
```
This will cause an error message to be displayed on hover, as there is no parsed module available. We should just fail gracefully here and return an empty response.

PR #619, opened 2024-04-03
lsp: Missing built-in data is not a hard error

charlieegan3:
Now when built-in state is not known for a file,
the hover response is empty. This is commonly the case when the file that has just been opened has failed to parse and never parsed in the time that the server has been running.

Fixes https://github.com/StyraInc/regal/issues/618


anderseknert (approved):
Thanks a lot!

Issue #589, opened 2024-03-10
Wildcard expressions in global ignore config not respected


anderseknert:
Given a workspace like this:

```shell
.
├── .regal
│   └── config.yaml
└── foo
    └── bar.rego
```

Where `.regal/config.yaml` looks like this:

```yaml
ignore:
  files:
    - foo/*
```

Running `regal lint` will have the linter ignore any violations in `foo/bar.rego`. The language server however doesn't seem to do that. Matching on exact names works, just not `**` or `*`.

PR #620, opened 2024-04-03
lsp: Support file ignore config

charlieegan3:
Fixes: https://github.com/StyraInc/regal/issues/589

This PR makes the filtering logic support a new rootDir prefix. When set, this will be stripped from file paths before testing their ignore status. 

https://github.com/StyraInc/regal/assets/1774239/e2234823-6c27-4df7-b546-062ee13afc60

### DIFF
--- a/pkg/config/filter.go
+++ b/pkg/config/filter.go
@@ -12,7 +12,12 @@ import (
 	"github.com/open-policy-agent/opa/bundle"
 )
 
-func FilterIgnoredPaths(paths, ignore []string, checkFileExists bool) ([]string, error) {
+func FilterIgnoredPaths(paths, ignore []string, checkFileExists bool, rootDir string) ([]string, error) {
+	// if set, rootDir is normalized to end with a platform appropriate separator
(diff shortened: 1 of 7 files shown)


pkg/config/filter.go
```diff
@@ -12,7 +12,12 @@ import (
 	"github.com/open-policy-agent/opa/bundle"
 )
 
-func FilterIgnoredPaths(paths, ignore []string, checkFileExists bool) ([]string, error) {
+func FilterIgnoredPaths(paths, ignore []string, checkFileExists bool, rootDir string) ([]string, error) {
```
anderseknert, line 15: This is a breaking change, I guess, but hopefully not one anyone has come to rely on 😬 And if they did, it should be easy to fix.

anderseknert (approved):
Very nice!

pkg/config/filter.go
charlieegan3: Discussed on slack, after being unable to find any public integrations based on this we have opted to merge and leave a note in the release.

PR #621, opened 2024-04-03
lsp: add inlay hint support

anderseknert:
This PR adds support for inlay hints of function arguments. Similar to the hover feature, only built-in functions are supported at this point.

Also did some minor refactoring in the hover function to have it use the same method for collecting built-in functions from a module as the inlay hint system does.


charlieegan3 (approved):
This looks great to me!

srenatus (approved):
This is exactly what we had hoped "some day" when introducing the NamedType stuff. 🆒 ✨

anderseknert:
Thank you both! 😃

PR #630, opened 2024-04-10
Code action for OPA fmt

charlieegan3:
https://github.com/StyraInc/regal/assets/1774239/88b8b6ab-8e23-48b5-b78a-2ebc4aa84700

### DIFF
--- a/internal/lsp/command.go
+++ b/internal/lsp/command.go
@@ -1,0 +1,19 @@
+package lsp
+
+func toAnySlice(a []string) []any {
+	b := make([]any, len(a))
+	for i := range a {
+		b[i] = a[i]
+	}
+
+	return b
+}
+
+func FmtCommand(args []string) Command {
+	return Command{
+		Title:     "Format using opa-fmt",
+		Command:   "regal.fmt",
+		Tooltip:   "Format using opa-fmt",
+		Arguments: toAnySlice(args),
+	}
+}
--- a/internal/lsp/format.go
+++ b/internal/lsp/format.go
@@ -11,16 +11,17 @@ import (
 func Format(path, contents string, opts format.Opts) (string, error) {
 	formatted, err := format.SourceWithOpts(filepath.Base(path), []byte(contents), opts)
 	if err != nil {
-		return "", fmt.Errorf("failed to format Rego source file: %v", err)
+		return "", fmt.Errorf("failed to format Rego source file: %w", err)
 	}
 
 	return string(formatted), nil
 }
 
-// ComputeEdits computes diff edits from 2 string inputs
+// ComputeEdits computes diff edits from 2 string inputs.
 func ComputeEdits(before, after string) []TextEdit {
 	ops := operations(splitLines(before), splitLines(after))
 	edits := make([]TextEdit, 0, len(ops))
+
 	for _, op := range ops {
 		switch op.Kind {
 		case Delete:
@@ -43,5 +44,6 @@ func ComputeEdits(before, after string) []TextEdit {
 		case Equal:
 		}
 	}
+
 	return edits
 }
--- a/internal/lsp/format.go
+++ b/internal/lsp/format.go
@@ -1,0 +1,47 @@
+package lsp
+
+import (
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	"github.com/open-policy-agent/opa/format"
+)
+
+func Format(path, contents string, opts format.Opts) (string, error) {
+	formatted, err := format.SourceWithOpts(filepath.Base(path), []byte(contents), opts)
+	if err != nil {
+		return "", fmt.Errorf("failed to format Rego source file: %v", err)
+	}
+
+	return string(formatted), nil
+}
+
+// ComputeEdits computes diff edits from 2 string inputs
+func ComputeEdits(before, after string) []TextEdit {
+	ops := operations(splitLines(before), splitLines(after))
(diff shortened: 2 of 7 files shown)


internal/lsp/command.go
```diff
@@ -0,0 +1,19 @@
...
+		b[i] = a[i]
+	}
+
+	return b
+}
+
+func FmtCommand(args []string) Command {
+	return Command{
+		Title:     "Format using opa-fmt",
+		Command:   "regal.fmt",
```
srenatus, line 15: Would you happen to know if there's a well-known command name or some other mechanism at play when an IDE/editor does "format on save" via LSP? What I'm wondering is what we'd need to do to have that just work. If it's command-name-based, we'd need to adjust this. But I know way too little about LSP myself 😅

internal/lsp/server.go
```diff
@@ -121,6 +134,8 @@ func (l *LanguageServer) Handle(
 		return struct{}{}, nil
 	}
 
+	fmt.Fprintf(l.errorLog, "%s\n", string(util.MustMarshalJSON(req)))
```
srenatus, line 137:
```suggestion
	fmt.Fprintln(l.errorLog, string(util.MustMarshalJSON(req)))
```
(if it was desired to keep this line)

internal/lsp/server.go
```diff
@@ -357,7 +440,58 @@ func (l *LanguageServer) handleHover(
...
+	if err := json.Unmarshal(*req.Params, &params); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal params: %w", err)
+	}
+
+	actions := make([]CodeAction, 0)
+
+	for _, diag := range params.Context.Diagnostics {
+		if diag.Code == "opa-fmt" {
+			actions = append(actions, CodeAction{
+				Title:       "Format using opa fmt",
```
srenatus, line 458: This is cool ✨

srenatus (approved):
I'm not the most competent reviewer, but I've reviewed it anyways 🙃 (except for the copied-over diff parts)

internal/lsp/command.go
srenatus: https://zed.dev/docs/configuring-zed#lsp this makes me think there's no convention...

anderseknert: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_formatting

And yeah, that'll be trivial to add with this code in place as the response is the same.

internal/lsp/server.go
charlieegan3: Thanks, I don't think we need this tbh. Will remove, I have some ideas to improve the logging in the server, but will do in another PR.

internal/lsp/server.go
```diff
@@ -204,6 +219,74 @@ func (l *LanguageServer) StartHoverWorker(ctx context.Context) {
...
+						},
+					},
+				}
+
+				l.logOutboundMessage(methodWorkspaceApplyEdit, editParams)
+
+				// note, here conn.Call is used as the workspace/applyEdit message is a request, not a notification
+				// as per the spec. In order to be 'routed' to the correct handler on the client it must have an ID
+				// receive responses too.
+				err = l.conn.Call(
```
anderseknert, line 276: Was this the only change from before? Because I'm like 99% sure I tried this too, but it didn't make a difference 😄

internal/lsp/format.go
```diff
@@ -0,0 +1,49 @@
...
+	formatted, err := format.SourceWithOpts(filepath.Base(path), []byte(contents), opts)
+	if err != nil {
+		return "", fmt.Errorf("failed to format Rego source file: %w", err)
+	}
+
+	return string(formatted), nil
+}
+
+// ComputeEdits computes diff edits from 2 string inputs.
+func ComputeEdits(before, after string) []TextEdit {
```
anderseknert, line 21: This was borrowed from regols, so we should give some credit (and a MIT header, I guess) here.

anderseknert (commented):
Awesome! Left a few comments, but nothing big.

internal/lsp/server.go
charlieegan3: The main diff is that this is A) Call and not notify, and B) does not happen while processing the request. This needs to be sent async.

internal/lsp/messages.go
```diff
@@ -137,6 +198,9 @@ type TextDocumentSyncOptions struct {
 
 type TextDocumentIdentifier struct {
 	URI string `json:"uri"`
+	// Version is optional (i.e. it can be null), but it cannot be undefined when used in some requests
```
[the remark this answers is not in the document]
charlieegan3: Yeah perhaps we should... I will do that.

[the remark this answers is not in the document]
charlieegan3: 7d0946da5b5d5a3d6c0a9f2d9b1a4926959ccf4d

internal/lsp/format.go
charlieegan3: Like this: c1a079681e395948f70f9123583d47a70a8321b2

anderseknert (approved):

internal/lsp/server.go
anderseknert: I guess it was the latter then... but how does the client know it's sent asynchronously, lol

anderseknert: That doesn't sound very asynchronous to me, haha, but you're clearly correct 👍

PR #632, opened 2024-04-10
Refactor LSP logging

charlieegan3:
While working on https://github.com/StyraInc/regal/pull/630, I needed to be able to see requests, responses and notifications to the server from the client and from the server to the client.

I learned that there is functionality in the jsonrpc2 lib that we use to do this nicely. This PR implements a new logging configuration to make use of this using an adapted version of the LogMessage function that is found in the library and using the OnRecv and OnSend functions.

Because of this, I have removed the logging responsibility from the server and moved it to the connection. The server now should log errors only and that should be the default mode of operation.

Enabling verbose logging will show you all messages being sent on the connection. ExcludeMethods and IncludeMethods can be used to filter the stream of messages when you are interested in monitoring or ignoring certain flows of messages. I know that this will be useful for me and I hope that this is a sensible enough way to do this overall.

### DIFF
--- a/cmd/languageserver.go
+++ b/cmd/languageserver.go
@@ -30,18 +30,12 @@ func init() {
 
 			ls := lsp.NewLanguageServer(opts)
 
-			loggingCfg := lsp.ConnectionLoggingConfig{
-				Writer:      os.Stderr,
-				LogInbound:  false,
-				LogOutbound: false,
-			}
-			if verboseLogging {
-				loggingCfg.LogInbound = true
-				loggingCfg.LogOutbound = true
-			}
-
 			conn := lsp.NewConnectionFromLanguageServer(ctx, ls.Handle, &lsp.ConnectionOptions{
-				LoggingConfig: loggingCfg,
+				LoggingConfig: lsp.ConnectionLoggingConfig{
+					Writer:      os.Stderr,
+					LogInbound:  verboseLogging,
+					LogOutbound: verboseLogging,
+				},
 			})
 
 			ls.SetConn(conn)
--- a/cmd/languageserver.go
+++ b/cmd/languageserver.go
@@ -7,7 +7,6 @@ import (
 	"os/signal"
 	"syscall"
 
-	"github.com/sourcegraph/jsonrpc2"
 	"github.com/spf13/cobra"
 
 	"github.com/styrainc/regal/internal/lsp"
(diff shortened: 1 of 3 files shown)


cmd/languageserver.go
```diff
@@ -26,19 +25,27 @@ func init() {
...
-			)
+			loggingCfg := lsp.ConnectionLoggingConfig{
+				Writer:      os.Stderr,
+				LogInbound:  false,
+				LogOutbound: false,
+			}
+			if verboseLogging {
+				loggingCfg.LogInbound = true
+				loggingCfg.LogOutbound = true
+			}
```
srenatus, line 41: [golf]
```suggestion
			loggingCfg := lsp.ConnectionLoggingConfig{
				Writer:      os.Stderr,
				LogInbound:  verboseLogging,
				LogOutbound: verboseLogging,
			}
```

😉 Like this, we could even inline it below.

internal/lsp/server.go
```diff
@@ -28,7 +27,9 @@ const (
 )
 
 type LanguageServerOptions struct {
-	ErrorLog       *os.File
+	ErrorLog *os.File
+
+	// Deprecated: VerboseLogging is deprecated and will be removed in a future release.
```
srenatus, line 32: If it's internal, do we care?

internal/lsp/server.go
```diff
@@ -692,8 +617,9 @@ func (l *LanguageServer) handleInitialize(
 	l.clientIdentifier = clients.DetermineClientIdentifier(params.ClientInfo.Name)
 
 	if l.clientIdentifier == clients.IdentifierGeneric {
-		l.log(
-			"Unable to match client identifier for initializing client, using generic functionality: " + params.ClientInfo.Name,
+		l.logError(
+			fmt.Errorf("unable to match client identifier for initializing client, using generic functionality: %s",
+				params.ClientInfo.Name),
```
srenatus, line 622: So it's an error, but it's not so bad that we cannot go on with some default? Is it an error, then? 🤔

srenatus (approved):
Just nitpicks/questions

cmd/languageserver.go
charlieegan3: Ahh, now that looks better! Done in 5541657aa7c8252a1ddd3ce3bfe8932250fdc095

internal/lsp/server.go
charlieegan3: Probably not, I'll just remove it.

charlieegan3:
Thanks for the review Stephan!

internal/lsp/connection.go
```diff
@@ -0,0 +1,224 @@
...
+			if !cfg.ShouldLog(method) {
+				return
+			}
+
+			switch {
+			case resp.Result != nil:
+				result, _ := json.Marshal(resp.Result)
+				logger.Printf("--> response #%s: %s: %s\n", resp.ID, method, result)
+			case resp.Error != nil:
+				err, _ := json.Marshal(resp.Error)
```
anderseknert, line 175: Heh, using `err` for anything but an `Error` sure is something 😅

anderseknert (approved):
Awesome! I can't wait to use this in future development 🤤

internal/lsp/connection.go
charlieegan3: Yeah... I have altered that now... 9892c609e05ea8ad441f5ff987ef8a38e645b2eb

charlieegan3:
Thanks! I think this should be really handy to tune into the messages you need.

charlieegan3:
https://github.com/open-policy-agent/vscode-opa/pull/147 is the sibling PR in the vscode ext repo

Issue #624, opened 2024-04-08
LSP: `opa-fmt` rule should be reported on entire package line, not just the first character


anderseknert:
It's easy to miss this one as it's only reported on the first character of the package line. Other rules, like `use-rego-v1` will use the whole line, and this one should do the same.

<img width="182" alt="Screenshot 2024-04-08 at 21 31 33" src="https://github.com/StyraInc/regal/assets/510711/2f6983e2-0f10-4a89-8a58-1b2fa3fdddc9">

PR #633, opened 2024-04-10
rules: OPA fmt rule location fix

charlieegan3:
Fixes https://github.com/StyraInc/regal/issues/624


pkg/rules/fmt.go
```diff
@@ -60,7 +66,10 @@ func (f *OpaFmtRule) Run(ctx context.Context, input Input) (*report.Report, erro
 						Reference:   f.Documentation(),
 					}},
 					Location: report.Location{
-						File: filename,
+						File:   filename,
+						Row:    1,
+						Column: 1,
```
anderseknert, line 71: Instead of hard coding the first line, do you think we should use the location (or at least row) from `module.Package`? Could be a (metadata) comment on the first line.

anderseknert (approved):
You're on 🔥 👏 Just a question / suggestion, but up to you

pkg/rules/fmt.go
charlieegan3: Ahh that's a good thought. I will do that tomorrow.

charlieegan3: I've updated this in 3697869752616f1dd7f1097f9a5e7f73bde1f0cf

Issue #625, opened 2024-04-09
Provide "Show documentation for <Regal rule>" as a code action


anderseknert:
The built-in Typescript extension does this, and while we show the link in the error message already, I think it would be a nice touch to have it in a code action too.

<img width="658" alt="Screenshot 2024-04-09 at 08 56 08" src="https://github.com/StyraInc/regal/assets/510711/97694f65-602f-4276-8e0e-5028bdfd779f">

I'm not 100% sure how they do this, but I'm sure we can figure it out.

PR #635, opened 2024-04-10
Open docs code action

charlieegan3:
This adds support for opening of docs as a code action for vscode.

Fixes: https://github.com/StyraInc/regal/issues/625

We might want to consider another implementation that uses sensible browser or https://pkg.go.dev/github.com/pkg/browser for non vscode users.

https://github.com/StyraInc/regal/assets/1774239/7ede9dab-ef94-43f3-9eb4-335a56ce2c97


anderseknert (approved):
Very nice!

PR #642, opened 2024-04-11
lsp: Accept but ignore cancelled requests

charlieegan3:
I have not been able to work out why these are sent but they don't seem to impact the running functionality of the server and supporting them seems to be optional.

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#dollarRequests

For now, this updates the server to do nothing when receiving such a request. This avoids some log lines that we don't do anything about.

I would like to come back to this though as it seems like there's more here that we might be missing. So it has been labelled as TODO.

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -125,6 +125,11 @@ func (l *LanguageServer) Handle(
 		}
 
 		return struct{}{}, nil
+	case "$/cancelRequest":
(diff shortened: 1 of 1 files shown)


charlieegan3:
logTrace and setTrace are similar, but I am more optimistic that we can add support for them quickly.

anderseknert (approved):

Issue #638, opened 2024-04-10
Implement `textDocument.formatting`


anderseknert:
The hard part of this is already done in #630, so this should just be a matter of implementing the handler.

anderseknert:
I should have known better than to claim this was "just a matter of ...". Obvious faux pas. 

While it was as simple as I had imagined implementing this in Regal, it turns out the UX for having multiple formatting providers in the _client_ was... well, you be the judge:

<img width="287" alt="Screenshot 2024-04-10 at 22 24 19" src="https://github.com/StyraInc/regal/assets/510711/94ba2485-e5a5-488a-8381-230939b5451d">
<img width="637" alt="Screenshot 2024-04-10 at 22 24 40" src="https://github.com/StyraInc/regal/assets/510711/76a29a36-66d6-4b89-b70d-aba6bb87468d">

So while I now have this "done" in Regal, and I could submit a PR just now — we'd need to update the VS Code extension to handle this better as soon as this lands in the server. Options I can think of would be:

1. Make the "native" client formatter conditional on the Regal LSP client not being started. This would however only work for those who run the latest version of the client... which is hopefully most users, but certainly not all.
2. Disable this server-side _if_ the client ID equals VS Code — we already have a client-side formatter doing the same thing there.
3. Remove the client side formatter and have the VS Code extension rely entirely on Regal for formatting. This would be the cleanest option as it would harmonize client implementations towards LSP features, but it would also mean that users that for whatever reason declined using Regal would lose an existing feature. That's... not good.

Thoughts?

charlieegan3:
Gosh, tough call! 😅 I don't like any of the options really.

When we activate the plugin, we run `registerDocumentFormattingEditProvider` to enable the client-side formatter, I wonder if we can later disable this if regal is enabled.

Some voodoo magic here: https://github.com/Microsoft/vscode/issues/70313 but it looks pretty promising.

I think that step 1 is merge: https://github.com/StyraInc/regal/tree/formatting with the vscode toggle off. Then we update vscode-opa to disable the OPA formatter when Regal is on following the example above, and at the same time undo the vscode toggle from step one in Regal.

anderseknert:
Sounds good to me! I'll disable it and will submit the PR 👍

charlieegan3:
This should be addressed with https://github.com/StyraInc/regal/pull/643 and https://github.com/open-policy-agent/vscode-opa/pull/148

PR #643, opened 2024-04-11
lsp: Enable formatting functionality

charlieegan3:
In https://github.com/StyraInc/regal/pull/641 support was added to handle textDocument/formatting requests from clients. However, the functionality was disabled for fear that users would see a confusing message:

https://github.com/StyraInc/regal/issues/638#issuecomment-2048459623

In https://github.com/open-policy-agent/vscode-opa/pull/148 the regal formatter will be used if the user has enabled regal and opa fmt will be used if not. This means that there is only ever one formatter available for rego and the user will see no message.


charlieegan3:
Fixes https://github.com/StyraInc/regal/issues/638

internal/lsp/messages.go
```diff
@@ -160,7 +160,7 @@ type FormattingOptions struct {
 	InsertSpaces           bool `json:"insertSpaces"`
 	TrimTrailingWhitespace bool `json:"trimTrailingWhitespace"`
 	InsertFinalNewline     bool `json:"insertFinalNewline"`
-	TrimFinalNewlines      bool `json:"trimFinalNewline"`
+	TrimFinalNewlines      bool `json:"trimFinalNewlines"`
```
anderseknert, line 163: 🦅👁️

anderseknert (approved):

Issue #626, opened 2024-04-09
Possible to specify path to Regal config file in the OPA VSCode Extention


bdjgs:
A setting for the OPA VSCode Extention that allows to specify a path to a Regal config file. 

Currently, it is required to have a .regal/config.yaml at the root of the every project. It would be great to be able to specify a path to a global Regal config that is used in all projects, if a local is not present.

charlieegan3:
Hmm, good point. Regal itself is actually designed with this in mind (see here: https://docs.styra.com/regal#configuration). However, this functionality isn't the same when Regal is running as a language server.

Let's fix that! 

One thing I'm not sure about is what to do when the global config changes while the server is running. At the moment, the client sends information about the updated config file to the server and it updates. However, then the config file loaded is outside the workspace of client, then we won't get updates in the same way. I think we'd need to implement some fsnotify based watching of the higher level config file. Doable, but less simple.

I guess the other option is that we allow the user to set a global config in the VSCode config, however, I like this less as it deviates from the Regal CLI behaviour, and still needs to the fsnotified...

anderseknert:
Thank you both!

@bdjgs may correct me if I'm wrong, but I would not expect a global Regal configuration shared between many projects to see that frequent updates. Considering the intricacies of fsnotify and portability, perhaps it'd be fine considering config outside of the workspace as static? At least for a first implementation. 

> I guess the other option is that we allow the user to set a global config in the VSCode config, however, I like this less as it deviates from the Regal CLI behaviour

Wouldn't that be equivalent to point out a config file via the CLI's `--config-file`?

bdjgs:
I dont know much about VSCode extentions so maybe the following is not possible, but I think it would be fine if the configfile is only loaded when a new workspace is loaded in VSCode. I do also think that it would be static for the most time and then when a change needs to be made, the user just have to reopen the VSCode workspace to reload the configfile. 

The feature request comes from the fact that if there is a central team managing the Regal configuration for a bunch of teams. If each team want to use this Regal linting locally, they would need to have the configfile, so they dont get errors on stuff that is ignored by the central team. And instead of each team having to have a configfile in each of their Rego projects, they can just have a global configfile on their PC and then update it whenever the central team annonces that changes has been made.

anderseknert:
Yep, certainly a valid scenario. I wonder if the VS Code client deals gracefully with symlinks?

[...]

PR #650, opened 2024-04-11
lsp: Load config from parent dirs

charlieegan3:
Related https://github.com/StyraInc/regal/issues/626

https://github.com/StyraInc/regal/assets/1774239/d8108b47-568c-4ce9-a0fc-5fd608960b8e

### DIFF
--- a/internal/lsp/config/watcher.go
+++ b/internal/lsp/config/watcher.go
@@ -41,9 +41,7 @@ func NewWatcher(opts *WatcherOpts) *Watcher {
 }
 
 func (w *Watcher) Start(ctx context.Context) error {
-	var err error
-
-	err = w.Stop()
+	err := w.Stop()
 	if err != nil {
 		return fmt.Errorf("failed to stop existing watcher: %w", err)
 	}
@@ -61,19 +59,17 @@ func (w *Watcher) Start(ctx context.Context) error {
 }
 
 func (w *Watcher) loop(ctx context.Context) {
-	var err error
-
 	for {
 		select {
 		case path := <-w.pathUpdates:
 			if w.path != "" {
-				err = w.fsWatcher.Remove(w.path)
+				err := w.fsWatcher.Remove(w.path)
 				if err != nil {
 					fmt.Fprintf(w.errorWriter, "failed to remove existing watch: %v\n", err)
 				}
 			}
 
-			err = w.fsWatcher.Add(path)
+			err := w.fsWatcher.Add(path)
 			if err != nil {
 				fmt.Fprintf(w.errorWriter, "failed to add watch: %v\n", err)
 			}
@@ -97,10 +93,10 @@ func (w *Watcher) loop(ctx context.Context) {
 				w.path = ""
 				w.Drop <- struct{}{}
 			}
-		case err = <-w.fsWatcher.Errors:
+		case err := <-w.fsWatcher.Errors:
 			fmt.Fprintf(w.errorWriter, "config watcher error: %v\n", err)
 		case <-ctx.Done():
-			err = w.Stop()
+			err := w.Stop()
 			if err != nil {
 				fmt.Fprintf(w.errorWriter, "failed to stop watcher: %v\n", err)
 			}
--- a/internal/lsp/config/watcher.go
+++ b/internal/lsp/config/watcher.go
@@ -1,0 +1,136 @@
+package config
+
+import (
+	"context"
+	"fmt"
+	"io"
+
+	"github.com/fsnotify/fsnotify"
+)
+
+type Watcher struct {
+	Reload chan string
+	Drop   chan struct{}
+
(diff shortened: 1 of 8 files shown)


charlieegan3:
Huh, there seems to be something wrong with a flake. I'll need to look into that next week 😞

internal/lsp/config/watcher_test.go
```diff
@@ -0,0 +1,71 @@
...
+		if err != nil {
+			t.Errorf("failed to start watcher: %v", err)
+		}
+	}()
+
+	watcher.Watch(configFilePath)
+
+	select {
+	case <-watcher.Reload:
+	case <-time.After(100 * time.Second):
```
anderseknert, line 42: Can we cap this at like 5-10 seconds or something? Anything above that would seem like something we'd want to look into anyway.

internal/lsp/server.go
```diff
@@ -434,7 +488,7 @@ func (l *LanguageServer) handleTextDocumentCodeAction(
 
 	for _, diag := range params.Context.Diagnostics {
 		switch diag.Code {
-		case "opa-fmt":
+		case ruleNameOPAFmt:
```
anderseknert, line 491: We should do this for the v1 rule too, but I guess it's not rebase here?

anderseknert (approved):
A couple of questions but looks good! 

Do I understand correctly that this will scan for a config file in any directory above the workspace, but it's not possible to point out the config file explcitly, like from `arbitrary/dir/.regal/config.yaml`?

internal/lsp/config/watcher.go
```diff
@@ -0,0 +1,136 @@
...
+		w.path = opts.Path
+	}
+
+	return w
+}
+
+func (w *Watcher) Start(ctx context.Context) error {
+	var err error
+
+	err = w.Stop()
```
srenatus, line 46: [nit] 
```suggestion
	err := w.Stop()
```

internal/lsp/config/watcher.go
```diff
@@ -0,0 +1,136 @@
...
+
+	go func() {
+		w.loop(ctx)
+	}()
+
+	return nil
+}
+
+func (w *Watcher) loop(ctx context.Context) {
+	var err error
```
srenatus, line 64: [style] Hmm there seems to be a pattern here.. there are a few situations where declaring this upfront is required, like
```go
foo := "y"
if x {
    foo, err = f() // := would introduce a new `foo` in this branch's scope, not what we want
}
```
but outside of those, I think I'd rather go with the `:=`..

srenatus (commented):
Just a nitpick, sorry 🔍 😅

internal/lsp/config/watcher.go
charlieegan3: Thanks, that's fair. I have fixed in b4efd89

internal/lsp/config/watcher.go
charlieegan3: Also fixed in b4efd89

internal/lsp/server.go
```diff
@@ -241,6 +246,55 @@ func (l *LanguageServer) formatToEdits(params ExecuteCommandParams, opts format.
...
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case path := <-l.configWatcher.Reload:
+			configFile, err := os.Open(path)
+			if err != nil {
+				l.logError(fmt.Errorf("failed to open config file: %w", err))
```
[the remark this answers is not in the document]
charlieegan3: I'll leave it for now 👍

internal/lsp/config/watcher_test.go
charlieegan3: Oops, that's a mistake. Fixed in d5bee69

internal/lsp/server.go
charlieegan3: I am lazy, golangci lint didn't flag that one... added in d5bee69

internal/lsp/config/watcher_test.go
charlieegan3: Made it 100ms like the others in bf48f11

anderseknert: Thanks!

charlieegan3:
I have made some adjustments to the server tests again to make them more robust to the order of messages. Feel free to take another look if you'd like or otherwise I think we can merge.

anderseknert:
@charlieegan3 good to merge!

Just this question:

> Do I understand correctly that this will scan for a config file in any directory above the workspace, but it's not possible to point out the config file explcitly, like from `arbitrary/dir/.regal/config.yaml`?

But if this solves the problem as reported, I think that could be a future improvement if requested.

charlieegan3:
I'll get back to the reporter and see if they can confirm 😊

Issue #92, opened 2023-04-13
Remediation


anderseknert:
Testing Regal on some policy libraries found online, there are a **lot** of violations reported for style-related rules, like preferring `:=` over `=` for top-level assignment. While we should be adamant about good style, there's also a clear risk that a user seeing hundreds of such violations either disables those rules, or worse — misses more severe violations as they drown in the amount of warnings reported.

For simple style rules like the one above, it would be interesting to explore if we could provide a remediation option, were we rewrote "bad" constructs to better ones without the user having to do the heavy lifting. Something similar to `opa fmt`, but limited to linter rules.

anderseknert:
Two linter rules now have remediations available via the Regal language server: [opa-fmt](https://docs.styra.com/regal/rules/style/opa-fmt) and [use-rego-v1](https://docs.styra.com/regal/rules/imports/use-rego-v1). More will likely follow soon. It would be good if all these remediations were made available via the CLI too as to avoid having drift in features supported by the CLI vs LSP.

PR #653, opened 2024-04-15
fixer: Regal fix command 

charlieegan3:
At the moment it works like this:

```
regal $ cat test.rego
package wow

#comment

allow = 1 {
        input.foo == true
}
regal $ ./regal fix test.rego
3 fixes applied:
test.rego:
- no-whitespace-comment
- use-assignment-operator
- use-rego-v1
regal $ cat test.rego
package wow

import rego.v1

# comment

allow := 1 if {
        input.foo == true
}
```

Related to https://github.com/StyraInc/regal/issues/92


pkg/fixer/fixer.go
```diff
@@ -0,0 +1,117 @@
...
+	"io"
+
+	"github.com/open-policy-agent/opa/ast"
+	"github.com/open-policy-agent/opa/format"
+
+	"github.com/styrainc/regal/pkg/fixer/fixes"
+	"github.com/styrainc/regal/pkg/report"
+)
+
+func Fix(rep *report.Report, readers map[string]io.Reader, _ fixes.Options) (map[string][]byte, error) {
```
charlieegan3, line 14: the idea with the readers here was that it made the impl not dependent on files. However, I'm not sure it's all that clean because of it.

pkg/fixer/fixer.go
```diff
@@ -0,0 +1,117 @@
...
+		"opa-fmt":     {},
+		"use-rego-v1": {},
+	}
+
+	filesToFix, err := computeFilesToFix(rep, readers, fixableViolations)
+	if err != nil {
+		return nil, fmt.Errorf("failed to determine files to fix: %w", err)
+	}
+
+	fixResults := make(map[string][]byte)
```
charlieegan3, line 25: Only the contents of the files that need to be fixed are loaded into memory.

cmd/lint.go
```diff
@@ -307,6 +312,13 @@ func lint(args []string, params lintCommandParams) (report.Report, error) {
 		return report.Report{}, fmt.Errorf("error(s) encountered while linting: %w", err)
 	}
 
+	if params.fix {
+		err = fixReport(&result)
```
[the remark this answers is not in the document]
charlieegan3: It also feels incorrect to assume that fixes are correct and that the report results without fix will be the same the next time around 🤔 

It feels like if there are many fixes, then it'd be hard to know for sure that `--fix` means the report's violations should be removed without re-linting.

pkg/fixer/fixer.go
```diff
@@ -0,0 +1,117 @@
...
+
+	"github.com/styrainc/regal/pkg/fixer/fixes"
+	"github.com/styrainc/regal/pkg/report"
+)
+
+func Fix(rep *report.Report, readers map[string]io.Reader, _ fixes.Options) (map[string][]byte, error) {
+	fixableViolations := map[string]struct{}{
+		"opa-fmt":     {},
+		"use-rego-v1": {},
+	}
```
charlieegan3, line 18: I think that something like what we have in OPA for builtins would be ideal here.

e.g.

```
var Builtins []*Builtin

// RegisterBuiltin adds a new built-in function to the registry.
func RegisterBuiltin(b *Builtin) {
	Builtins = append(Builtins, b)
	BuiltinMap[b.Name] = b
	if len(b.Infix) > 0 {
		BuiltinMap[b.Infix] = b
	}
}
```

then the fixes init functions can call RegisterFix or something?

pkg/fixer/fixer.go
```diff
@@ -0,0 +1,117 @@
...
+				fixed = false
+			}
+
+			if fixed {
+				fixedViolations = append(fixedViolations, i)
+			}
+		}
+	}
+
+	for i := len(fixedViolations) - 1; i >= 0; i-- {
```
charlieegan3, line 74: grim

pkg/fixer/fixes/options.go
```diff
@@ -0,0 +1,12 @@
+package fixes
+
+import "github.com/open-policy-agent/opa/format"
+
+type Options struct {
+	Fmt *FmtOptions
+}
+
+type FmtOptions struct {
```
charlieegan3, line 9: FmtOptions was meant to be an options struct for all the known fixes but I think that `regal lint --fix` won't have any options. and the LSP will just call  fixes on their own so this feels wrong...

pkg/fixer/fixes/fmt.go
```diff
@@ -0,0 +1,29 @@
+package fixes
+
+import (
+	"fmt"
+	"path/filepath"
+
+	"github.com/open-policy-agent/opa/format"
+)
+
+func Fmt(in []byte, opts *FmtOptions) (bool, []byte, error) {
```
charlieegan3, line 10: `Fmt` is the first and currently only implementation of a 'fix', next would be `:=` but I think that this interface is approx correct. where a fix has a byte slice input and some options where fix config can be set.

For example, if a single line or set of lines were to be fixed only, then that could be set in options if the fix supported it.

charlieegan3:
Also, interested to get thoughts on `--fix` vs `regal fix`. Having a dedicated command gives us more options for flags etc. but `--fix` is what might feel more familiar.

pkg/fixer/fixer.go
anderseknert: I guess it could just be a map of name -> func too?

pkg/fixer/fixes/fmt.go
anderseknert: Yeah, that makes sense to me! Just thinking out loud now but I suppose another option would be that the functions all returned `TextEdit` arrays just like the LSP client expects, but we would also implement a "remediator" in Regal that would take that an transform the input text using that. It's basically just a list of locations and changes, so I don't think it'll be hard to do... that way both the LSP and "native" commands would look the same, just passed onwards to different remediators.

But it was just an idea, I haven't considered if it's a good one or not, lol

pkg/fixer/fixes/options.go
anderseknert: If we used `regal fix` I guess the options could be to provide `--rules` similarly to the `lint` command. But yeah, I don't imagine we'll want configuration options for individual remediations.

anderseknert (commented):
Nice work! I'm fine with the direction taken here 👍 Although I'm now leaning more and more towards providing a `regal fix` command for a cleaner separation of linting and transformations. I've left some comments here and there for further discussion.

cmd/lint.go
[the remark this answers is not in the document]
charlieegan3: I've reworked this all today to be `regal fix`, lmk what you think!

pkg/fixer/fixer.go
charlieegan3: This is how it works now:

```go
	f := fixer.Fixer{}
	f.RegisterFixes(fixer.NewDefaultFixes()...)
```

I think this is ok too?

pkg/fixer/fixer.go
charlieegan3: sorted, this isn't what I we do now in `regal fix`

pkg/fixer/fixes/options.go
charlieegan3: This has all gone now. no-whitespace-comment has been added too.

Violations will only be fixed where there is an instance in the lint report. So this `take existing ruleconfiguration into account` should happen by default.

charlieegan3:
@anderseknert this is ready for review again. I am more happy with the implementation here now. If you agree on the direction & interfaces etc, then I think we can get this in.

We might want to add code actions for the whitespace comment and assignment operator too, but these could happen as a follow on PR since this is already kinda large.

pkg/fixer/fixer.go
```diff
@@ -0,0 +1,236 @@
...
+	if !ok {
+		return nil, false
+	}
+
+	return fixInstance, true
+}
+
+// OrderedFixes returns the fixes in the order they should be applied.
+// Fixes that are marked as WholeFile are applied last since they
+// can add new lines that would affect the fixes for line based violations.
```
anderseknert, line 131: Hmm, I see your point. But also — if run last, it's possible that the formatter itself introduces some changes that makes it non-compliant with the fixes that got applied before it. As an example:

```rego
allow {
	true
}
```
will be rewritten to be:
```rego
allow = true
```
I don't think there are many examples of this, but it's obviously a problem if running `regal fix` still leaves some files with violations 🤔

pkg/fixer/fixes/nowhitespacecomment.go
```diff
@@ -0,0 +1,43 @@
...
+		return false, nil, nil
+	}
+
+	for _, loc := range opts.Locations {
+		if loc.Row > len(lines) {
+			return false, nil, nil
+		}
+
+		if loc.Col != 1 {
+			// current impl only understands the first column
```
anderseknert, line 31: Interesting. Just a shortcut or tricky to make it work if the comment is later on the line?

pkg/fixer/fixes/nowhitespacecomment_test.go
```diff
@@ -0,0 +1,120 @@
...
+			afterFix: []byte(`package test\n
+
+# this is a comment
+`),
+			fixExpected: true,
+			runtimeOptions: &RuntimeOptions{
+				Locations: []ast.Location{
+					{
+						Row: 3,
+						Col: 1, // this is what the rule outputs at the moment
```
anderseknert, line 58: Is it? AFAICS I get the col where the comment starts
```
Rule:         	no-whitespace-comment
Description:  	Comment should start with whitespace
Category:     	style
Location:     	p.rego:13:3
Text:         	} #doo
Documentation:	https://docs.styra.com/regal/rules/style/no-whitespace-comment
```

pkg/fixer/fixes/useassignmentoperator.go
```diff
@@ -0,0 +1,41 @@
...
+		return false, nil, nil
+	}
+
+	for _, loc := range opts.Locations {
+		if loc.Row > len(lines) {
+			return false, nil, nil
+		}
+
+		if loc.Col != 1 {
+			// current impl only understands the first column
```
anderseknert, line 31: OK, there's probably something I don't understand here 😅

pkg/fixer/reporter.go
```diff
@@ -0,0 +1,31 @@
+package fixer
+
+import (
+	"fmt"
+	"io"
+)
+
+// Reporter writes a summary of the fixes applied to the output writer.
+func Reporter(outputWriter io.Writer, fixReport *Report) {
```
anderseknert, line 9: I think It's fine to only support a single output format for the first implementation, but I think we should provide the `--format` flag from the start, with only a single option available, and this implementation here prepared for more formats getting added later.

anderseknert (commented):
Amazing work on this! The code is really clean, and I had no problems whatsoever following along exactly in what it does. I did have a few thoughts and questions, but this is definitely getting there 😍 Next release is gonna be 💥

cmd/fix.go
```diff
@@ -0,0 +1,269 @@
...
+
+		PreRunE: func(_ *cobra.Command, args []string) error {
+			if len(args) == 0 {
+				return errors.New("at least one file or directory must be provided for fixing")
+			}
+
+			return nil
+		},
+
+		RunE: func(_ *cobra.Command, args []string) error {
```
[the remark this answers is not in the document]
charlieegan3: I opted to drop this since it didn't seem to add profiling information that would be different from the lint output?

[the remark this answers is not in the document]
charlieegan3: I can add it back in though since it does have all the go metrics etc.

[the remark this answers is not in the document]
charlieegan3: Added back in 95bd4b0b743f2c7cd87844c2df5ae979775da07f

cmd/fix.go
```diff
@@ -0,0 +1,269 @@
...
+	fixer.Reporter(outputWriter, fixReport)
+
+	return nil
+}
+
+func fixLintReport(rep *report.Report) (*fixer.Report, error) {
+	fileReaders := make(map[string]io.Reader)
+
+	for _, v := range rep.Violations {
+		f, err := os.Open(v.Location.File)
```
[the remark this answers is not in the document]
charlieegan3: Ahh yeah. That was the whole idea here but I forgot that detail... fixed in 03fac1230eb71416b2478174351af40ff87d0c50

cmd/fix.go
```diff
@@ -0,0 +1,269 @@
...
+	f := fixer.Fixer{}
+	f.RegisterFixes(fixer.NewDefaultFixes()...)
+
+	fixReport, err := f.Fix(rep, fileReaders)
+	if err != nil {
+		return fixReport, fmt.Errorf("error encountered while fixing: %w", err)
+	}
+
+	for file, content := range fixReport.FileContents() {
+		f, err := os.OpenFile(file, os.O_RDWR|os.O_TRUNC, 0o755)
```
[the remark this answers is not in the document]
charlieegan3: Ahh yeah that makes sense. Fixed in b885fa196031faf42258accc725e745aac4cbec1

pkg/fixer/fixes/nowhitespacecomment_test.go
charlieegan3: Huh, I could have sworn this was only for the first char... fixed in fa62e1865e9484ca50014b86833b01b845fefa61.

I think that opa fmt must have trimmed it on save and it was moved to the first char in my test file...

pkg/fixer/fixes/nowhitespacecomment.go
charlieegan3: See note below, fixed in https://github.com/StyraInc/regal/commit/fa62e1865e9484ca50014b86833b01b845fefa61

pkg/fixer/fixes/useassignmentoperator.go
charlieegan3: So I think for this one I am correct in that it shows as char one: 

```
regal $ cat test.rego
package wow

allow = true #thjings
regal $ cat test.rego
package wow
allow = true
regal $ regal lint test.rego --disable-all --enable=use-assignment-operator
Rule:           use-assignment-operator                                                 
Description:    Prefer := over = for assignment                                         
Category:       style                                                                   
Location:       test.rego:2:1                                                           
Text:           allow = true                                                            
Documentation:  https://docs.styra.com/regal/rules/style/use-assignment-operator        

1 file linted. 1 violation found.

```

pkg/fixer/fixer.go
charlieegan3: Ahh I hadn't used that as an example, where only opa-fmt is needed and not rego.v1.

```
package wow

import rego.v1

allow {
    true
}
```

Will leave violations now...

I wonder if for the fixing, we could only run the RegoV0CompatV1 formatter?

pkg/fixer/fixes/useassignmentoperator.go
anderseknert: Ah, right, since the `=` is nowhere to be found in the AST. I'll take a stab at fixing this in the rule instead, as that'll get us more accurate reporting in the CLI output as well.

charlieegan3: I can change the rule with that in mind.

charlieegan3: ca8e0a2e6b2130299667c2123fee7016351fff51 makes this change

pkg/fixer/fixer.go
charlieegan3: Changes from today impl this now. Namely: https://github.com/StyraInc/regal/pull/653/commits/5f42a5ceb9a3f5208c98995bc094942b6bf71cb8

charlieegan3:
This is ready for another review. I think we are getting there. Main changes:

- The fixer will now run repeatedly until there are no fixes applied https://github.com/StyraInc/regal/pull/653#discussion_r1567822052
- Interfaces for fixing are now more generic and should better support other fix operations such as renaming.

pkg/fixer/fixer.go
```diff
@@ -0,0 +1,128 @@
...
+	for {
+		fixMadeInIteration := false
+
+		in, err := fp.ToInput()
+		if err != nil {
+			return nil, fmt.Errorf("failed to generate linter input: %w", err)
+		}
+
+		fixLinter := l.WithDisableAll(true).
+			WithEnabledRules(fixableEnabledRules...).
```
anderseknert, line 71: Nice

pkg/fixer/fixes/fixes.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"github.com/open-policy-agent/opa/ast"
+	"github.com/open-policy-agent/opa/format"
+)
+
+// NewDefaultFixes returns a list of default fixes that are applied by the fix command.
+// When a new fix is added, it should be added to this list.
+func NewDefaultFixes() []Fix {
+	return []Fix{
+		&Fmt{},
```
anderseknert, line 12: Not that it matters, but do we even need to include this when the next fix is also calling the formatter?

pkg/fixer/fixes/fmt.go
```diff
@@ -0,0 +1,55 @@
...
+	// this allows this fix config to also be registered under another key so that different
+	// configurations can be registered under other linter rule keys.
+	if f.KeyOverride != "" {
+		return f.KeyOverride
+	}
+
+	return "opa-fmt"
+}
+
+func (*Fmt) WholeFile() bool {
```
anderseknert, line 30: Do we need this any longer? From what I can tell these are never called.

anderseknert (approved):
Terrific! Only some nits, but none of them blocking, so go on and merge when you feel done! 👏 

This will be an awesome addition to Regal, and will save a lot of time both for us and others.

pkg/fixer/fixes/fixes.go
charlieegan3: If this were removed, the below case would not be covered:

```
regal $ cat test.rego
package badformatv1

import rego.v1

  allow := true
regal $ go run main.go lint test.rego 
Rule:           opa-fmt                                                 
Description:    File should be formatted with `opa fmt`                 
Category:       style                                                   
Location:       test.rego:1:1                                           
Text:           package badformatv1                                     
Documentation:  https://docs.styra.com/regal/rules/style/opa-fmt        

1 file linted. 1 violation found.
exit status 3
[1]regal $ go run main.go fix test.rego 
No fixes applied.
```

Each fix in the list of NewDefaultFixes has a `Key()`, the key corresponds to the name of a linter error that the fix addresses and triggers the running of the fix.

So if a file doesn't have `use-rego-v1` but is badly formatted, then it would only have a `opa-fmt` violation. And if this were removed, then the file would not be fixed by formatting.

I wonder if it makes sense for `Key()` to be `Name()` to match what we have on rules...

pkg/fixer/fixes/fmt.go
charlieegan3: Ahh, oops. Fixed in f05782e1cc5cc1416ed6672da5d298baa9ec9353

pkg/fixer/fixes/fixes.go
charlieegan3: Renamed in 8b9e6285662b11afb0a404982b533d6b07350235

charlieegan3: https://github.com/StyraInc/regal/pull/653/commits/01a38884ebbdcc36a2ea42b01a6c153682a75391

anderseknert: I was thinking the default fixes would all be applied, but it's later filtered by actual linter errors is what you're saying? Yeah, that makes total sense then 👍

charlieegan3: Yeah exactly, this is the default in that it's the 'defaultly enabled list' of fixes to be run when there are linter violations.

PR #658, opened 2024-04-17
Rule: `unresolved-import`

### LINKED ISSUE #300: Rule: `unresolved-import`
This is requested from time to time in OPA, even as far back as [2017](https://github.com/open-policy-agent/opa/issues/491)! 

How imports are handled is unlikely to change in OPA, but we could certainly be more stringent in Regal. Once #282 has landed, it would be simple to aggregate the imports of all policies, and then have an aggregate rule that "resolves" all imports — i..e checks that there is a corresponding module to map to.

Obviously, we'll need to allow exceptions for dynamic / data imports, which would be easy to do either via configuration:

```yaml
rules:
  custom:
    unresolved-import:
      level: error
      except-imports:
        - data.some.dynamic.stuff
``` 

or an ignore directive, like:

```rego
# regal ignore:require-imports
import data.some.dynamic.stuff
```

Making this a custom rule seems like the best way to me, as I don't think we'll want this enabled by default, and the rule likely will need custom configuration in most repos.

anderseknert:
Requested since 2017! This aggregate rule will try to resolve any import provided in policy, while allowing the user to configure paths that should be resolved at runtime, like data refs.

See the docs included in the PR for more details.

Fixes #300

Screenshot 2024-04-18 at 09 00 24

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,6 +232,7 @@ The following rules are currently available:
 | imports     | [prefer-package-imports](https://docs.styra.com/regal/rules/imports/prefer-package-imports)           | Prefer importing packages over rules                      |
 | imports     | [redundant-alias](https://docs.styra.com/regal/rules/imports/redundant-alias)                         | Redundant alias                                           |
 | imports     | [redundant-data-import](https://docs.styra.com/regal/rules/imports/redundant-data-import)             | Redundant import of data                                  |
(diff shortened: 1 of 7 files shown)


charlieegan3 (approved):
This looks really good to me, no comments. Great docs and options too

PR #660, opened 2024-04-18
lsp: fix flaky test

charlieegan3:
This test was relying on the correct sequence of messages to be sent to the client. Now we just wait for a condition to be true and continue when it is met.

This was missed as part of the test updates in #650 testRequestDataCodes is now used for these examples too.


anderseknert (approved):
Thanks!

Issue #659, opened 2024-04-18
Create code action for `no-whitespace-comment`


anderseknert:
The backend for this is now in place with #653 , so we'll just need to expose this in the language server.

Issue #649, opened 2024-04-11
Create code action for `use-assignment-operator`


anderseknert:
Having a code action to convert `=` to `:=` when reported by the [use-assignment-operator](https://docs.styra.com/regal/rules/style/use-assignment-operator) rule would quite convenient. Since it could be tedious to do this for all violations in some legacy policies, perhaps we can have two actions — one to do it at the location of the violation, and one to do it across the whole file.

PR #661, opened 2024-04-18
lsp: Implement code actions for new fixes

charlieegan3:
https://github.com/StyraInc/regal/pull/653 added fixes for no-whitespace-comment and use-assignment-operator. This PR adds code actions for these in the regal lsp.

Fixes https://github.com/StyraInc/regal/issues/659

Fixes https://github.com/StyraInc/regal/issues/649

### DIFF
--- a/internal/lsp/commands/parse_test.go
+++ b/internal/lsp/commands/parse_test.go
@@ -1,0 +1,110 @@
+package commands
+
+import (
+	"testing"
+
+	"github.com/open-policy-agent/opa/ast"
+
+	"github.com/styrainc/regal/internal/lsp/types"
+)
+
+func TestParse(t *testing.T) {
+	t.Parallel()
+
+	testCases := map[string]struct {
+		ExecuteCommandParams types.ExecuteCommandParams
+		ParseOptions         ParseOptions
+		ExpectedTarget       string
+		ExpectedLocation     *ast.Location
+	}{
+		"extract target only": {
+			ExecuteCommandParams: types.ExecuteCommandParams{
+				Command:   "example",
+				Arguments: []interface{}{"target"},
+			},
+			ParseOptions:     ParseOptions{TargetArgIndex: 0},
+			ExpectedTarget:   "target",
+			ExpectedLocation: nil,
+		},
+		"extract target and location": {
+			ExecuteCommandParams: types.ExecuteCommandParams{
+				Command:   "example",
+				Arguments: []interface{}{"target", "1", 2}, // different types for testing, but should be strings
+			},
+			ParseOptions:     ParseOptions{TargetArgIndex: 0, RowArgIndex: 1, ColArgIndex: 2},
+			ExpectedTarget:   "target",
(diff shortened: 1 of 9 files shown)


internal/lsp/types/types.go
```diff
@@ -0,0 +1,6 @@
+package types
+
+type ExecuteCommandParams struct {
```
charlieegan3, line 3: For now, I have only moved this type here. In a follow on PR, I'll move them all here.

internal/lsp/commands/parse_test.go
```diff
@@ -0,0 +1,110 @@
...
+				Arguments: []interface{}{"target"},
+			},
+			ParseOptions:     ParseOptions{TargetArgIndex: 0},
+			ExpectedTarget:   "target",
+			ExpectedLocation: nil,
+		},
+		"extract target and location": {
+			ExecuteCommandParams: types.ExecuteCommandParams{
+				Command:   "example",
+				Arguments: []interface{}{"target", "1", 2}, // different types for testing, but should be strings
```
anderseknert, line 32: I suppose we could have standardized on args just being a single object for all commands, which we could serialize into a struct. But this might be more aligned with what the LSP "expects".

internal/lsp/server.go
```diff
@@ -307,73 +269,60 @@ func (l *LanguageServer) StartConfigWorker(ctx context.Context) {
...
-				edits, target, err := l.formatToEdits(params, format.Opts{RegoVersion: ast.RegoV0CompatV1})
-				if err != nil {
-					l.logError(err)
+				fixed, editParams, err = l.fixEditParams(
+					"Format for Rego v1 using opa-fmt",
+					&fixes.Fmt{OPAFmtOpts: format.Opts{RegoVersion: ast.RegoV0CompatV1}},
+					commands.ParseOptions{TargetArgIndex: 0},
+					params,
+				)
+			case "regal.use-assignment-operator":
```
anderseknert, line 302: For consistency, perhaps we should rename `regal.fmt` -> `regal.opa-fmt` and `regal.fmt.v1` -> `regal.use-rego-v1` above?

anderseknert: Or perhaps `regal.fix.opa-fmt` and so on, in case we want to add commands later that aren't about editing.

anderseknert (approved):
Looks great! Just some questions.

internal/lsp/server.go
charlieegan3: Yeah, I think that makes sense.

internal/lsp/commands/parse_test.go
charlieegan3: Yeah I wasn't so sure about this but figured that we'd go with this until we ran into some issues. It's all internal so seemed ok to go with this for now.

internal/lsp/server.go
charlieegan3: Fixed in 3a1d447bd0ef6ab6305e73408880565b445d9501

PR #662, opened 2024-04-18
lsp: refactor types

charlieegan3:
Having all these types in the lsp package makes it harder to create subpackages in the LSP without cyclic import issues.

This PR moves them all under a types subpackage.


anderseknert (approved):
Very nice! Once things have stabilized, I could imagine even making this a separate lib.

charlieegan3:
Yeah I can imagine further breaking these down too 👍

PR #663, opened 2024-04-19
lsp: Support folding ranges

### LINKED ISSUE #648: Implement "folding range"
As described by the spec: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_foldingRange

Note that VS Code already does this, likely via [this file](https://github.com/open-policy-agent/vscode-opa/blob/master/language-configuration.json), but IntelliJ and other editors don't. I don't think there'll be a conflict in VS Code when both a language definition and a language server is provided, but that should be tested.

anderseknert:
Adding [folding range](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_foldingRange) support for:
* Imports
* Comments (when grouped on more than one line)
* Blocks from tokens like `{}`, `[]`, `()`

Scanner stolen from OPA so feel free to review that, but fixes should go into OPA ;)

Fixes #648

### DIFF
--- a/internal/lsp/foldingrange.go
+++ b/internal/lsp/foldingrange.go
@@ -1,0 +1,155 @@
+package lsp
+
+import (
+	"strings"
+
+	"github.com/open-policy-agent/opa/ast"
+
+	"github.com/styrainc/regal/internal/lsp/opa/scanner"
+	"github.com/styrainc/regal/internal/lsp/opa/tokens"
+	"github.com/styrainc/regal/internal/lsp/types"
+)
+
+type stack []scanner.Position
+
+func (s stack) Push(p scanner.Position) stack {
+	return append(s, p)
+}
+
+func (s stack) Pop() (stack, scanner.Position) {
+	l := len(s)
+
+	return s[:l-1], s[l-1]
+}
+
+func TokenFoldingRanges(policy string) []types.FoldingRange {
+	scn, err := scanner.New(strings.NewReader(policy))
+	if err != nil {
+		panic(err)
+	}
+
+	var lastPosition scanner.Position
+
+	curlyBraceStack := stack{}
+	bracketStack := stack{}
+	parensStack := stack{}
+
+	foldingRanges := make([]types.FoldingRange, 0)
+
+	for {
+		token, position, _, errors := scn.Scan()
+
+		if token == tokens.EOF || len(errors) > 0 {
+			break
+		}
+
+		switch {
+		case token == tokens.LBrace:
+			curlyBraceStack = curlyBraceStack.Push(position)
+		case token == tokens.RBrace && len(curlyBraceStack) > 0:
+			curlyBraceStack, lastPosition = curlyBraceStack.Pop()
+
+			startChar := uint(lastPosition.Col)
+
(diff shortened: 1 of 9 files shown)


internal/lsp/foldingrange.go
```diff
@@ -0,0 +1,176 @@
...
+
+	curlyBraceStack := stack{}
+	bracketStack := stack{}
+	parensStack := stack{}
+
+	foldingRanges := make([]types.FoldingRange, 0)
+
+	for {
+		token, position, _, errors := scn.Scan()
+
```
srenatus, line 41: [style] With all these,

```go
if something {
    // ...
    continue
}
if somethingElse {
    // ...
    continue
}
```

we could also use a conditionless switch
```go
for {
  token, position, _, errors := scn.Scan()
  switch {
    case token == tokens.EOF || len(errors) > 0:
        break
    case token= tokens.LBRACE:
        curlyBraceStack = curlyBraceStack.Push(position)
    case token == tokens.RBrace && len(curlyBraceStack) > 0:
        // ...
  }
}
```

internal/lsp/opa/scanner/scanner.go
```diff
@@ -0,0 +1,444 @@
+// Copyright 2020 The OPA Authors.  All rights reserved.
+// Use of this source code is governed by an Apache2
+// license that can be found in the LICENSE file.
```
srenatus, line 3: [nit] Lets reference where it came from, it'll make it easier to adjust this in the future

srenatus (approved):

internal/lsp/foldingrange.go
anderseknert: Ah, yeah, that's better! I'll make it so, thanks 👍

internal/lsp/opa/scanner/scanner.go
anderseknert: There aren't that many scanner.go files in OPA :) But sure, I'll add a link.

srenatus: Commit message would be fine, too!

internal/lsp/foldingrange.go
```diff
@@ -0,0 +1,155 @@
...
+		if token == tokens.EOF || len(errors) > 0 {
+			break
+		}
```
srenatus, line 44: This could also go into the switch, can't it?

internal/lsp/foldingrange.go
anderseknert: It didn't work with the `break` condition as it won't break out of the loop but the switch. So I moved that on top and kept the rest.

internal/lsp/foldingrange.go
anderseknert: That went into and endless loop, presumably because the `break` breaks out of the `switch` and not the loop.

internal/lsp/foldingrange.go
srenatus: ah ok 😅 could have used labels, but this is fine, too!

srenatus (approved):
code looks good, but I haven't tried this locally

PR #664, opened 2024-04-20
lsp: handle `textDocument/definition` requests

### LINKED ISSUE #645: Implement "go to definition"
As described in the spec https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_definition

OPA already provides some means to do this, and the VS Code extension currently calls the "opa oracle" command. A first implementation would likely use the same API from Regal.

As this too is covered by the VS Code client already, we'll want to guard this similarly to how @charlieegan3 implemented it for formatting https://github.com/open-policy-agent/vscode-opa/pull/148

anderseknert:
This was rather simple, as this was already handled by the `opa oracle find-definition` command. The API for that was however internal, so I've copied the single file handling that from OPA.

PR marked WIP as two things warrants some discussion:
* Similarly to formatting, this LSP feature seems to conflict with client side implementations, and you'll see duplicate "definitions" in VS Code when this is enabled both in the client and from the LSP server. We should probably make this conditional on Regal in VS Code.
* The `GetAllModules` function that exists on the cache does not seem to take into account files that have been marked as ignored via `.regal/config.yaml`. This lead to an annoying situation when this was used in the Regal repo, as some files we've ignored did not compile, and that caused go to definition to fail.

Fixes #645

Screenshot 2024-04-19 at 22 45 28

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -21,6 +21,7 @@ import (
 	"github.com/styrainc/regal/internal/lsp/clients"
 	"github.com/styrainc/regal/internal/lsp/commands"
 	lsconfig "github.com/styrainc/regal/internal/lsp/config"
+	"github.com/styrainc/regal/internal/lsp/opa/oracle"
 	"github.com/styrainc/regal/internal/lsp/types"
 	"github.com/styrainc/regal/internal/lsp/uri"
 	"github.com/styrainc/regal/pkg/config"
@@ -102,6 +103,8 @@ func (l *LanguageServer) Handle(
 		return l.handleInitialized(ctx, conn, req)
 	case "textDocument/codeAction":
 		return l.handleTextDocumentCodeAction(ctx, conn, req)
+	case "textDocument/definition":
+		return l.handleTextDocumentDefinition(ctx, conn, req)
 	case "textDocument/diagnostic":
 		return l.handleTextDocumentDiagnostic(ctx, conn, req)
 	case "textDocument/didOpen":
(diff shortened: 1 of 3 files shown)


charlieegan3:
> The GetAllModules function that exists on the cache does not seem to take into account

So now that we know this was a misleading error relating to inlay hints, fixed in #669. We can have a think about:

> Similarly to formatting, this LSP feature seems to conflict with client side implementations

I think it makes sense to just disable the provider in the client when regal is on. I did the same for formatting so I'll do that now.

charlieegan3 (approved):
This all looks good to me, safe to merge but needs some client updates to make useful.

internal/lsp/server.go
```diff
@@ -608,6 +611,46 @@ func (l *LanguageServer) handleTextDocumentInlayHint(
...
+		Filename: uri.ToPath(l.clientIdentifier, params.TextDocument.URI),
+		Pos:      positionToOffset(contents, params.Position),
+		Modules:  l.cache.GetAllModules(),
+		Buffer:   []byte(contents),
+	}
+
+	definition, err := orc.FindDefinition(query)
+	if err != nil {
+		// fail silently — the user could have clicked anywhere
+		return nil, nil //nolint:nilerr,nilnil
```
charlieegan3, line 640: I thought we could return `struct{}{}` here in such cases?

anderseknert: I had some problems before returning an empty array, but yeah that seems to work. I'll update.

PR #668, opened 2024-04-22
lsp: handle `textDocument/documentSymbol`

anderseknert:
More weekend fun! This provides a basic implementation of `textDocument/documentSymbol`. I say basic because we currently scan for nested symbols, but only package, rules and functions, with some nice touches like providing details on what kind of rule it is, and the type it returns, and so on. I think this covers what most actually need, but we can certainly refine this later and add things like variables inside of rule bodies, and so on.

Screenshot 2024-04-22 at 12 21 07

### DIFF
--- a/internal/lsp/documentsymbol.go
+++ b/internal/lsp/documentsymbol.go
@@ -1,0 +1,269 @@
+package lsp
+
+import (
+	"bytes"
+	"fmt"
+	"strings"
+
+	"github.com/open-policy-agent/opa/ast"
+
+	"github.com/styrainc/regal/internal/lsp/types"
+	"github.com/styrainc/regal/internal/lsp/types/symbols"
+)
+
+//nolint:nestif
+func documentSymbols(
+	contents string,
+	module *ast.Module,
+) []types.DocumentSymbol {
+	// Only pkgSymbols would likely suffice, but we're keeping docSymbols around in case
+	// we ever want to add more top-level symbols than the package.
+	docSymbols := make([]types.DocumentSymbol, 0)
+	pkgSymbols := make([]types.DocumentSymbol, 0)
+
+	lines := strings.Split(contents, "\n")
+
+	pkgRange := types.Range{
+		Start: types.Position{Line: 0, Character: 0},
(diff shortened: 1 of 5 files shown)


internal/lsp/documentsymbol.go
```diff
@@ -0,0 +1,267 @@
...
+
+			pkgSymbols = append(pkgSymbols, groupSymbol)
+		}
+	}
+
+	if len(pkgSymbols) > 0 {
+		pkg.Children = &pkgSymbols
+	}
+
+	docSymbols = append(docSymbols, pkg)
```
charlieegan3, line 123: is `docSymbols` doing something I'm missing? It looks like it'll only ever be the pkgSymbols right?

anderseknert: No, you're right. I was probably too cautious, and wanted to prepare for the case where two "packages" was in the same document, like when general ref heads rules are used. But I guess even in those cases one will be a child of the other.

I don't think it hurts having a top-level array of symbols though, should we later want to add anything else there. But I can remove it if you think it's redundant.

charlieegan3: nah it's fine!

charlieegan3: Perhaps a comment for the intended purpose of docSymbols would be enough? / helpful

anderseknert: Sure, I can add a comment 👍

internal/lsp/documentsymbol.go
```diff
@@ -0,0 +1,267 @@
...
+func getRuleDetail(rule *ast.Rule) string {
+	if rule.Head.Args != nil {
+		return "function" + rule.Head.Args.String()
+	}
+
+	if rule.Head.Key != nil && rule.Head.Value == nil {
+		return "multi-value rule"
+	}
+
+	if rule.Head.Value != nil {
```
charlieegan3, line 184: could exit early here for `rule.Head.Value == nil` and then avoid a level of nesting in the latter half? Just an idea, feel free to ignore. If `getRuleDetail` is going to be extended for lots of other cases it might not be a good idea.

charlieegan3 (approved):
This looks really nice! great feature and the implementation was easy to follow too! 🏁

PR #673, opened 2024-04-23
lsp: implement `workspace/symbol`

anderseknert:
This was rather trivial with document symbols supported. It's now made possible to search for symbols across the whole workspace, which is rather nice!

Screenshot 2024-04-23 at 07 26 20


charlieegan3 (approved):
This looks great to me!

PR #680, opened 2024-04-24
Fix false positive in `unresolved-import` and map rules

anderseknert:
Thanks @nevumx for reporting this


nevumx (approved):
fellowship-of-the-ring-fellowship

johanfylling (approved):
LGTM 👍

Issue #682, opened 2024-04-25
Installing OPA - extension server exited


scoop96:
Hi, we are starting to test OPA-Rego at work, and i found in the repo a similar closed issue with regal two weeks ago.
OS is a windows 10.

When I add the opa VSCODE extension it installs regal and opa, i added them to the path (both account and system environmental variables)

here is the VSCODE Output from regal

```
panic: runtime error: index out of range [0] with length 0

goroutine 19 [running]:
github.com/styrainc/regal/pkg/config.FindRegalDirectory({0xc0002b47c1, 0x1d})
	/home/runner/work/regal/regal/pkg/config/config.go:138 +0x465
github.com/styrainc/regal/pkg/config.FindConfig({0xc0002b47c1?, 0xc0002d81e0?})
	/home/runner/work/regal/regal/pkg/config/config.go:147 +0x1c
github.com/styrainc/regal/internal/lsp.(*LanguageServer).handleInitialize(0xc000096880, {0x140?, 0x15b2220?}, 0xc0002b21e0?, 0xc0002b0ba0)
	/home/runner/work/regal/regal/internal/lsp/server.go:1024 +0x517
github.com/styrainc/regal/internal/lsp.(*LanguageServer).Handle(0xc000096880, {0x1b7f1e8, 0xc0000afc20}, 0xc00008ca20, 0xc0002b0ba0)
	/home/runner/work/regal/regal/internal/lsp/server.go:102 +0x191
github.com/sourcegraph/jsonrpc2.(*HandlerWithErrorConfigurer).Handle(0xc00008aeb0, {0x1b7f1e8, 0xc0000afc20}, 0xc00008ca20, 0xc0002b0ba0)
	/home/runner/go/pkg/mod/github.com/sourcegraph/jsonrpc2@v0.2.0/handler_with_error.go:21 +0x57
github.com/sourcegraph/jsonrpc2.(*Conn).readMessages(0xc00008ca20, {0x1b7f1e8, 0xc0000afc20})
	/home/runner/go/pkg/mod/github.com/sourcegraph/jsonrpc2@v0.2.0/conn.go:205 +0x2dd
created by github.com/sourcegraph/jsonrpc2.NewConn in goroutine 1
	/home/runner/go/pkg/mod/github.com/sourcegraph/jsonrpc2@v0.2.0/conn.go:62 +0x1e6
[Error - 10:05:40 AM] Server initialization failed.
  Message: Pending response rejected since connection got disposed
  Code: -32097 
[...]

PR #684, opened 2024-04-25
Avoid panic when searching for `.regal/config.yaml`

anderseknert:
As reported in #682, the server panics when the dir for whatever reason is "". I don't yet know how to trigger this, and the result of this fix is likely that the config file won't be found when this happens. But as a temporary measure I prefer to have Regal run with default configuration compared to crashing.

### DIFF
--- a/cmd/lint.go
+++ b/cmd/lint.go
@@ -299,7 +299,7 @@ func lint(args []string, params *lintCommandParams) (report.Report, error) {
 		}
 
 		regal = regal.WithUserConfig(userConfig)
-	case err != nil && params.configFile != "":
+	case params.configFile != "":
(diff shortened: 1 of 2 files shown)


cmd/lint.go
```diff
@@ -299,7 +299,7 @@ func lint(args []string, params *lintCommandParams) (report.Report, error) {
 		}
 
 		regal = regal.WithUserConfig(userConfig)
-	case err != nil && params.configFile != "":
+	case params.configFile != "":
```
anderseknert, line 302: This is unrelated. Just fixing a redundant check as the condition will always be true.

srenatus (approved):

PR #685, opened 2024-04-26
fix: make sure deleted items aren't added back to the cache

### LINKED ISSUE #679: Reported parse error from empty module persists even after file is deleted
This is an edge case, but perhaps it applies to other errors too when a file has been deleted. Either way, there seems to be some state that isn't cleared after a file has been removed. In order to reproduce:

1. Create an empty policy file (`foo.rego`)
2. See that the server corrrectly reports a parse error
3. Delete `foo.rego`
4. See that the server (or at least the client) still reports a parse error for that file

Screenshot 2024-04-24 at 19 56 48

anderseknert:
Fixes #679

unrelated: disabled the gocognit linter as it's annoying

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -84,6 +84,7 @@ type LanguageServer struct {
 type fileUpdateEvent struct {
 	Reason  string
 	URI     string
+	OldURI  string
 	Content string
 }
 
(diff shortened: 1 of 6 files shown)


internal/lsp/server.go
```diff
@@ -873,7 +884,6 @@ func (l *LanguageServer) handleWorkspaceDidDeleteFiles(
 		}
 
 		l.diagnosticRequestFile <- evt
-		l.builtinsPositionFile <- evt
```
srenatus, line 876: I'm not super familiar with the code base, so please enlighten me: why do we no longer need to inform the builtins-positions thing?

internal/lsp/server.go
```diff
@@ -429,25 +439,26 @@ func (l *LanguageServer) processTextContentUpdate(
...
-func (l *LanguageServer) processBuiltinsUpdate(
-	_ context.Context,
-	uri string,
-	content string,
-) (bool, error) {
+func (l *LanguageServer) processBuiltinsUpdate(_ context.Context, uri string, content string) error {
+	if _, ok := l.cache.GetFileContents(uri); !ok {
+		// If the file is not in the cache, exit early or else
+		// we might accidentally put it in the cache after it's been
+		// deleted: https://github.com/StyraInc/regal/issues/679
```
srenatus, line 446: Reading that issue, I have to wonder: does this change also take care of renamed files?

srenatus (approved):
Code looks good, just some meta questions

internal/lsp/server.go
anderseknert: There are no builtin positions in a deleted file :)

internal/lsp/server.go
anderseknert: Hmm, no, I'll look into it. Thanks!

internal/lsp/server.go
srenatus: ....and the builtins-positions-thingy doesn't have a cache that needs updating?

srenatus: Ah OK. Thanks!

internal/lsp/server.go
anderseknert: That's now handled 👍

PR #686, opened 2024-04-26
fix: Avoid panic in config finder

anderseknert:
Not sure if it's worth having two different exit conditions, but I guess it doesn't hurt with some granularity?

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -139,6 +139,10 @@ func FindRegalDirectory(path string) (*os.File, error) {
 			return nil, errors.New("stopping as dir is empty string")
 		}
 
+		if len(parts) < 2 {
+			return nil, errors.New("stopping as dir is root directory")
+		}
+
 		parts = parts[:len(parts)-1]
 
 		if parts[0] == volume {


pkg/config/config.go
```diff
@@ -139,6 +139,10 @@ func FindRegalDirectory(path string) (*os.File, error) {
 			return nil, errors.New("stopping as dir is empty string")
 		}
 
+		if len(parts) < 2 {
+			return nil, errors.New("stopping as dir is root directory")
+		}
```
srenatus, line 144: Is "root directory" well defined for windows? How about `C:\` or `D:\` or `\\`?

srenatus (approved):
I think I'd use more `path/filepath` here (like https://pkg.go.dev/path/filepath#Dir), but since a minimal change is good, LGTM.

pkg/config/config.go
anderseknert: Yeah, no idea. I hope this code isn’t permanent, as I’ll want to find out what’s going on here… but first to get a windows box somehow.

PR #698, opened 2024-05-07
Rule: `impossible-not`

### LINKED ISSUE #603: Rule: `impossible-not`
This is a mistake I see devs do every now and then — it just never struck me as something we could try and prevent. 

```rego
deny contains something if {
    # ...
}

test_deny if {
    not deny with # ...
}
```

We should flag when `not` is used on anything but boolean rules / vars. This might not be possible to determine in all cases based on the AST alone, but for the cases where we can, we should.

anderseknert:
See docs in PR (and issue) for further details.

Fixes #603

### DIFF
--- a/docs/rules/bugs/impossible-not.md
+++ b/docs/rules/bugs/impossible-not.md
@@ -1,0 +1,85 @@
+# impossible-not
+
+**Summary**: Impossible `not` condition
+
+**Category**: Bugs
+
+**Type**: Aggregate - only runs when more than one file is provided for linting
+
+**Avoid**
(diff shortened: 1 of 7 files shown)


docs/rules/bugs/impossible-not.md
```diff
@@ -0,0 +1,85 @@
+# constant-condition
```
charlieegan3, line 1:
```suggestion
# impossible-not
```

charlieegan3 (approved):
lgtm, really great that we can catch this one

Issue #696, opened 2024-05-07
Unparseable document freezes inlay hints


anderseknert:
As long as the working document parses, inlay hints are recalculated as they should. But if there are parser errors, it looks like we use the last good location. This is probably the right choice for some cases, but it can also lead to situtations where inlay hints obstruct the text being worked on:

<img width="716" alt="Screenshot 2024-05-06 at 21 58 34" src="https://github.com/StyraInc/regal/assets/510711/9fee12c0-3306-40ba-87ef-454e38acf0ca">

Not sure what the best solution is. Clearing all inlay hints on parse errors is obviously the easy choice, but perhaps we could be smarter about it an clear only those around where the parse error is reported.

PR #701, opened 2024-05-08
lsp: Fix issue with parse error inlay hints

charlieegan3:
Fixes https://github.com/StyraInc/regal/issues/696

This change updates the server to attempt a partial parse instead when there are parse errors to still, if possible, show the hints for sections of the file before the first parse error.

This is a best effort adjustment, and if the parse still fails, all inlay hints are cleared until one or all of the parse errors are addressed.


anderseknert (approved):
Looks good to me!

PR #703, opened 2024-05-08
lsp: URI to path conversions respect os.Separator

charlieegan3:
This was causing an issue in the loading of config files on windows https://github.com/StyraInc/regal/issues/682


anderseknert (approved):
Thanks!

PR #704, opened 2024-05-08
lsp: Rename lastValidLine variable

charlieegan3:
This is really the first line that contains a parse error, and so a better name would be clearer.


anderseknert (approved):
Thanks!

Issue #646, opened 2024-04-11
Implement "completion proposals"


anderseknert:
As described in the spec: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion

anderseknert:
I say we close this one as done with #709 

Let's create new issues for specific things we want to have completions for, like references, or `input`.

PR #709, opened 2024-05-09
lsp/completions: Implement a minimal completions provider

charlieegan3:
Related to https://github.com/StyraInc/regal/issues/646, but more work to do here so not to close.

here is a little demo of what is done:

https://github.com/StyraInc/regal/assets/1774239/b7ebb65e-d9ed-43a6-a23a-0f27ff3c612c


internal/lsp/completions/providers/regov1.go
```diff
@@ -0,0 +1,64 @@
...
+
+	if !strings.HasPrefix(line, "import ") { // if in rule body
+		return nil, nil
+	}
+
+	words := strings.Split(line, " ")
+	lastWord := words[len(words)-1]
+
+	//nolint:gocritic
+	if !strings.HasPrefix("rego.v1", lastWord) {
```
anderseknert, line 40: I'm not sure I understand this 😅 But it works so I must be missing something. But we exit if the last word doesn't start with "rego.v1"? 🤔

internal/lsp/completions/providers/shared.go
```diff
@@ -0,0 +1,3 @@
+package providers
+
+const fileURI = "file:///foo/bar/file.rego"
```
anderseknert, line 3: What does this do?

internal/lsp/completions/providers/package.go
```diff
@@ -0,0 +1,64 @@
...
+	fileContents, ok := c.GetFileContents(fileURI)
+	if !ok {
+		// if the file contents is missing then we can't provide completions
+		return []types.CompletionItem{}, nil
+	}
+
+	lines := strings.Split(fileContents, "\n")
+	if len(lines) < 1 {
+		return []types.CompletionItem{}, nil
+	}
```
anderseknert, line 25: Looks like pretty much all of the `Run` implementations share the first 10 lines, which makes sense as they'll all need the contents split by line. I wonder if it would make sense to have the manager do this instead and have it passed to the `Run` functions. Not so much as an optimization (but it won't hurt) but just to avoid cluttering each of the `Run` implementations with the same setup code.

anderseknert (approved):
This is an awesome addition, and the refactoring was icing on the cake 👍 Just had a few comments, but I'd be happy to have it merged as is.

Thank you! 👏

internal/lsp/completions/providers/shared.go
charlieegan3: I dislike this but https://golangci-lint.run/usage/linters/#goconst is on and there doesn't seem to be a way to ignore test files...

I have made it clearer in 166faa0da9c44324714a3bc28c2d3bcc5d611d80, but I'd rather just disable goconst if we can for such things. I'd rather have it repeated as I feel like it makes the tests clearer.

internal/lsp/completions/providers/regov1.go
charlieegan3: I have added more notes for this in 37e711cbe1289749707b0c6fc75ef0c037bccdf3, does that make it more clear?

internal/lsp/completions/providers/packagename.go
```diff
@@ -0,0 +1,70 @@
...
+		}
+	}
+
+	// if we can't confirm that the user has package statement on the line then we don't provide completions
+	if len(lines)+1 < int(params.Position.Line) {
+		return nil, nil
+	}
+
+	path := uri.ToPath(clients.IdentifierGeneric, fileURI)
+	dir := filepath.Base(filepath.Dir(path))
```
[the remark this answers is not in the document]
charlieegan3: One of the challenges here is that we'd need to know where to stop, this would require being told the workspace root. should we always suggest the longest path, or all three? e/g/ bugs, rules.bugs, and regal.rules.bugs?

internal/lsp/completions/providers/regov1.go
anderseknert: Ah yes, it was the order of the arguments that I had missed 👍

internal/lsp/completions/providers/packagename.go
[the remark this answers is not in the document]
anderseknert: Yeah, that is indeed a challenge. I guess what we could do is to start from the first 3 (or so) characters that match, so if you have something like:

```
/Users/charlie/projects/regal/rules/bugs
```

And you typed in `package reg` it would base the suggestion from the first matching position in the string, which would be `regal/rules/bugs` in this case. But yeah, I'm sure there would be cases where it could get tricky.

internal/lsp/completions/providers/regov1.go
charlieegan3: Yeah, it does seem like an odd order... but I think it's still valid! 😎

internal/lsp/completions/providers/packagename.go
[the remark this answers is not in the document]
charlieegan3: Using the workspace root from the server's config feels like the right option here.

Do you think that we'd always want to suggest the longest package name we can, and that one only? I.e. only suggest `regal.rules.bugs`


If so, that might be worth doing as part of this PR.

internal/lsp/completions/providers/package.go
charlieegan3: Very open to revisiting when we see what shape the other providers start to look like.

internal/lsp/completions/providers/regov1.go
anderseknert: Yes, clever :)

internal/lsp/completions/providers/packagename.go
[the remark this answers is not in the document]
charlieegan3: Screenshot 2024-05-16 at 11 18 39


See c163642ec742701a0866d417140003278b5f957e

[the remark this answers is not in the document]
anderseknert: very nice!

PR #710, opened 2024-05-09
Add a few tasks for VS Code

anderseknert:
Just a minor convenience really but quite nice for Regal development. Also add `go mod tidy` as part of the rq PR script.


charlieegan3 (approved):
We are getting all in on VS Code!

PR #713, opened 2024-05-13
Single file `impossible-not`

anderseknert:
Improve this new rule to have the `report` rule handle single file violations, while the aggregate report ignores violations where the `not` ref and the partial rule is in the same file.


charlieegan3 (approved):
This looks great!

PR #714, opened 2024-05-13
Rule: `messy-rule`

### LINKED ISSUE #672: Rule: `messy-rule`
While working on support for [document symbols](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_documentSymbol), one of the open questions was how to best represent Rego's incremental rule definitions for a format that largely expects location _ranges_ for identification. @charlieegan3 suggested a rule to enforce that incremental rules and function definitions are grouped together in a package. So something like this would be discouraged:

```rego
package p

allow if something

unrelated_rule if {
    # ...
}

allow if something_else
```

In favor of:

```rego
package p

allow if something

allow if something_else

unrelated_rule if {
    # ...
}
```

Looking at the Regal code base, we're violating this frequently ourselves 😅 And since it "doesn't matter" it's easy to forget about without tooling to help enforce this. Needless to say, this would be a `style` category rule, or possibly even `custom` if we think this should be optional. But as the document symbols feature showed us, even style can have a real-world impact.

anderseknert:
Help identify incremental rule definitions that aren't grouped together.

Also fix two violations against this rule that we had in our own codebase.

Fixes #672

### DIFF
--- a/docs/rules/style/messy-rule.md
+++ b/docs/rules/style/messy-rule.md
@@ -1,0 +1,57 @@
+# messy-rule
+
+**Summary**: Messy incremental rule
+
+**Category**: Style
+
+**Avoid**
+
+```rego
+package policy
+
+allow if something
+
+unrelated_rule if {
+    # ...
+}
+
+allow if something_else
+```
+
+**Prefer**
+
+```rego
+package policy
+
+allow if something
+
+allow if something_else
+
+unrelated_rule if {
+    # ...
+}
+```
+
+## Rationale
+
+Rules that are definecd incrementally should have their definitions grouped together, as this makes the code easier to
+follow. While this is mostly a style preference, having incremental rules grouped also allows editors like VS Code to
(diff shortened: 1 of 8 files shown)


docs/rules/style/messy-rule.md
```diff
@@ -0,0 +1,57 @@
+# messy-rule
```
charlieegan3, line 1: I wonder if 'disjointed' is would be more specific here? Fragmented might be another option.

charlieegan3 (approved):
Looks good to me, I think 'messy' might be a little generic and initially suggests issues relating to whitespace or similar. I think this rule is a really great one to have though.

docs/rules/style/messy-rule.md
anderseknert: Either way you'd need to read the rule docs to learn what it is about, so I think "messy" works well to catch the user's attention :)

PR #715, opened 2024-05-13
Add docs as part of `regal new rule` command

### LINKED ISSUE #700: Add documentation template for new rules created with `regal new`
For built-in rules created with the `regal new` command, we currently add only the Rego files, and an entry in the default configuration. It would be nice if we also added the markdown file to document the rule, as this is currently done manually.

anderseknert:
Fixes #700


charlieegan3 (approved):
LGTM

PR #716, opened 2024-05-13
Rule: `trailing-default-rule`

### LINKED ISSUE #699: rule: Require ordering of default rule before other rule heads
This is an idea for a new rule.

The example below would violate the rule, as the default keyword is after the other rule head.

```rego
package example

import rego.v1

allow if input.admin

default allow := false
```

This fix:

```rego
package example

import rego.v1

default allow := false

allow if input.admin
```

The rationale for this would be that as more rule heads are added, the default definition is pushed farther and farther down the file and this more likely to be missed.

This rule should only apply to rule heads in one source file, if the default is elsewhere, then that's ok.

anderseknert:
Simple rule to check that `default` rules are placed before non-default rules.

Fixes #699

### DIFF
--- a/bundle/regal/rules/style/trailing_default_rule.rego
+++ b/bundle/regal/rules/style/trailing_default_rule.rego
@@ -1,0 +1,24 @@
+# METADATA
+# description: Default rule should be declared first
+package regal.rules.style["trailing-default-rule"]
+
+import rego.v1
+
+import data.regal.ast
+import data.regal.result
+
+report contains violation if {
+	some i, rule in input.rules
+
+	rule["default"] == true
+
+	name := ast.ref_to_string(rule.head.ref)
+	name in all_names(array.slice(input.rules, 0, i))
+
+	violation := result.fail(rego.metadata.chain(), result.location(rule))
+}
+
+all_names(rules) := {name |
+	some rule in rules
+	name := ast.ref_to_string(rule.head.ref)
+}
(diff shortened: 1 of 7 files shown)


bundle/regal/rules/style/trailing_default_rule.rego
```diff
@@ -0,0 +1,24 @@
...
+import data.regal.ast
+import data.regal.result
+
+report contains violation if {
+	some i, rule in input.rules
+
+	rule["default"] == true
+
+	name := ast.ref_to_string(rule.head.ref)
+	name in all_names(array.slice(input.rules, 0, i))
```
charlieegan3, line 16: 🤔 how should I read this? if the default rule is in the 0-ith range of rules? Perhaps it's too early, but it's not clear to me how this is working 😅

charlieegan3 (approved):
LGTM, but I'm not sure how the actual rule is working.

bundle/regal/rules/style/trailing_default_rule.rego
anderseknert:
```
array.slice(input.rules, 0, i)
```
provides a slice of all the rules up until `i`, i.e the `default` rule.. and then we check if the same name of the default has been used for any of those.

charlieegan3: Ahh, yes. that does make sense. I'd assumed that the implementation was going to be like: make sure that the rule is the first when ordered by line or similar, but this works too! thanks.

PR #717, opened 2024-05-14
Send warning to client on CRLF line endings

### LINKED ISSUE #693: Better handling of CRLF
As discussed in [Styra community Slack](https://styracommunity.slack.com/archives/C05HF2KRY3G/p1714476295049779) there is an issue with the `opa-fmt` rule, Windows' line endings and IDEs like VS Code that automatically changes line endings.
image

Regal shows a warning when the file contains `CRLF` line endings. Running `opa fmt` through the IDE does not fix the warning because when `opa fmt` outputs `LF` line endings the IDE automatically changes the `LF` line endings to `CRLF` line endings. 

Users can usually fix this quite easily by configuring their IDE to use `LF` endings, but the the problem of `CRLF` line endings is not obvious for users to begin with. It would be nice if Regal could somehow let the user know about the issue with `CRLF` line endings. 

I have two ideas on how this could be improved: 
1. Update the current warning message to include a note about `CRLF`. Maybe the message could be dynamic and only include a note about `CRLF` if such endings are found in the file? 
2. or a new Regal rule to warn about `CRLF` endings

anderseknert:
In case either `opa-fmt` or `use-rego-v1` is enabled, we'll send back a warning to the client whenever they save a document containing CRLF line endings. This to avoid confusion when the user tries to fix formatting using `opa fmt`, but the editor adding back CRLF line endings after.

Also disable the nestif linter, as it's annoying.

Fixes #693

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -27,6 +27,7 @@ import (
 	rparse "github.com/styrainc/regal/internal/parse"
 	"github.com/styrainc/regal/pkg/config"
 	"github.com/styrainc/regal/pkg/fixer/fixes"
+	"github.com/styrainc/regal/pkg/linter"
 )
 
 const (
@@ -114,6 +115,8 @@ func (l *LanguageServer) Handle(
 		return l.handleTextDocumentDidOpen(ctx, conn, req)
 	case "textDocument/didClose":
 		return struct{}{}, nil
+	case "textDocument/didSave":
+		return l.handleTextDocumentDidSave(ctx, conn, req)
 	case "textDocument/documentSymbol":
 		return l.handleTextDocumentDocumentSymbol(ctx, conn, req)
 	case "textDocument/didChange":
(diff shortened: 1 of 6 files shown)


internal/lsp/server.go
```diff
@@ -800,6 +803,56 @@ func (l *LanguageServer) handleTextDocumentDidChange(
...
+
+		for _, rule := range enabled {
+			if rule == "opa-fmt" || rule == "use-rego-v1" {
+				formattingEnabled = true
+
+				break
+			}
+		}
+
+		resp := types.ShowMessageParams{
```
charlieegan3, line 838: I guess we only need to make this resp if formatting is on?

anderseknert: heh, yeah, well spotted. I'll have it moved 👍

internal/lsp/server.go
```diff
@@ -800,6 +803,56 @@ func (l *LanguageServer) handleTextDocumentDidChange(
...
+			}
+		}
+
+		if formattingEnabled {
+			resp := types.ShowMessageParams{
+				Type:    2, // warning
+				Message: "CRLF line ending detected. Please change editor setting to use LF for line endings.",
+			}
+
+			err := l.conn.Notify(ctx, "window/showMessage", resp)
```
charlieegan3, line 844: I remember there was an issue in `handleWorkspaceExecuteCommand` where we had blocking issues sending notify messages mid request response. I guess this isn't an issue here if it works, but something to double check.

We don't have any other Notify calls that are sent at request time. However, I think this is fine here as the response isn't needed/blocking at the client for this operation - it would seem.

charlieegan3 (approved):

internal/lsp/server.go
anderseknert: Yeah, I just interpreted "notification" to mean "notify" in the spec sense: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showMessage

But it probably just means "notification" as in the UI.. do you think I should use a `Call` here instead?

anderseknert: But yeah, this does seem to work reliably.

PR #721, opened 2024-05-18
Return `null` instead of empty structs

anderseknert:
Where the specification says we should. Did not change this for the notification type events, as they don't render responses. Perhaps we should change the signature of those functions to not return a result?

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -1,3 +1,4 @@
+//nolint:nilnil
 package lsp
 
 import (
@@ -517,7 +518,8 @@ func (l *LanguageServer) handleTextDocumentHover(
 	if !ok {
 		l.logError(fmt.Errorf("could not get builtins for uri %q", params.TextDocument.URI))
 
-		return struct{}{}, nil
+		// return "null" as per the spec
+		return nil, nil
 	}
 
 	for _, bp := range builtinsOnLine[params.Position.Line+1] {
@@ -537,7 +539,8 @@ func (l *LanguageServer) handleTextDocumentHover(
 		}
 	}
 
-	return struct{}{}, nil
+	// return "null" as per the spec
+	return nil, nil
 }
 
 func (l *LanguageServer) handleTextDocumentCodeAction(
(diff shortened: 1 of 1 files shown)


internal/lsp/server.go
```diff
@@ -781,7 +782,7 @@ func (l *LanguageServer) handleTextDocumentDefinition(
 	definition, err := orc.FindDefinition(query)
 	if err != nil {
 		// fail silently — the user could have clicked anywhere
-		return struct{}{}, nil //nolint:nilerr
+		return nil, nil //nolint:nilerr
```
charlieegan3, line 785: Do we need the nolint here too? as well as at the top.

internal/lsp/server.go
```diff
@@ -517,7 +518,7 @@ func (l *LanguageServer) handleTextDocumentHover(
 	if !ok {
 		l.logError(fmt.Errorf("could not get builtins for uri %q", params.TextDocument.URI))
 
-		return struct{}{}, nil
+		return nil, nil
```
charlieegan3, line 521: I'd also add a link to https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_hover just to be really clear that this is an important return value.

internal/lsp/server.go
anderseknert: The one at the top covers the whole file.

internal/lsp/server.go
anderseknert: I added comments where we return null as per the specification

internal/lsp/server.go
charlieegan3: But we don't need `//nolint:nilerr` here right?

charlieegan3 (approved):

internal/lsp/server.go
anderseknert: We do, as `nilnil` is the one we ignore on the file.

```
internal/lsp/server.go:787:3: error is not nil (line 784) but it returns nil (nilerr)
		return nil, nil
```

That one seems to check for return of nil while `err != nil`

PR #722, opened 2024-05-20
lsp: Return null when no hover items

charlieegan3:
The spec states this value should be null when there are no hover items to show.


charlieegan3:
Closing in favour of https://github.com/StyraInc/regal/pull/721

PR #723, opened 2024-05-20
Some fixes for detecting external refs

### LINKED ISSUE #719: Inconsistencies with external-reference warning
When linting my code I stumbled upon a case where I expected an external-reference warning, but there was none. I wrote a few test functions that I believe should trigger the warning but it doesn't

See play here: https://play.openpolicyagent.org/p/o5xBt1XUHP

`func_3` and `func_5` trigger warnings as expected, but the other funcs do not

anderseknert:
The following cases are now covered:

```rego
f(_) := r

f(_) := {"r": r}
```
And variations thereof.

Fixes #719

Although it doesn't fix the `func_2` example there, this is about as much we can do without some major effort.. so I'm not sure it's worth keeping that issue open as I doubt we'll get there anytime soon.


charlieegan3 (approved):
LGTM!

PR #728, opened 2024-05-21
lsp: show correct OPA error hint links

charlieegan3:
When regal encounters a parse error, we can sometimes workout the correct opa errors page to send the user to. This PR makes use of some existing code that matches errors to help pages.

Screenshot 2024-05-21 at 13 52 27


anderseknert (approved):
🥳

PR #735, opened 2024-05-22
Update capabilities.json

anderseknert:


charlieegan3 (approved):

anderseknert:
> Should we think about an automation for this as it's easily forgotten?

Yeah that’d be good. The reason why we don’t have one is that it doesn’t matter *that* much if we forget this. But something in CI to check that it’s up to date should be fairly simple

PR #741, opened 2024-05-23
Better align with specification based on Zed client testing

anderseknert:
Make optional attributes optional, and omit when empty.


charlieegan3 (approved):
👍

PR #754, opened 2024-05-27
Don't suggest completions for internal rules or functions

### LINKED ISSUE #751: Completions: don't suggest rules starting with underscore
While it's not an "official" pattern, it's common enough that we probably should add something about it to the [Rego Style Guide](https://github.com/StyraInc/rego-style-guide), as has been [suggested](https://github.com/StyraInc/rego-style-guide/issues/5) in the past. There's also an old suggestion to have a linter rule [enforce this](https://github.com/StyraInc/regal/issues/94).

While completions should suggest `_rules` from the same package, it should not do so from other packages, as at least if this convention is followed, they're not intended to be exported.

anderseknert:
Fixes #751

### DIFF
--- a/internal/lsp/completions/providers/rulerefs.go
+++ b/internal/lsp/completions/providers/rulerefs.go
@@ -53,6 +53,12 @@ func (*RuleFromImportedPackageRefs) Run(
 				continue
 			}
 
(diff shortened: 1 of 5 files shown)


charlieegan3 (approved):
👍 

All good here!

PR #756, opened 2024-05-27
Use completion item kind table for completions icons

### LINKED ISSUE #755: Completions: incorrect icons used in completions popup
A nit for sure, but looks like we no longer get the correct icons for the completion suggestions:

Screenshot 2024-05-27 at 21 35 56

(that's not the icon for a function)

Looking at the code, I think this comes from reusing _symbols_ from the document/workspace symbols implementation, which while seemingly rational isn't the right mapping for completions, as they have a lookup table of [their own](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItemKind).

anderseknert:
Fixes #755


charlieegan3 (approved):
This is great and much needed. Is there more we can do here to share with the outline icons? (even as a follow on PR)

anderseknert:
@charlieegan3 what do you mean with "share with the outline icons?" 🙂

PR #759, opened 2024-05-28
Add completions for default and import keywords

### LINKED ISSUE #758: Completions: show suggestions for `import` and `default` keywords
These are always at the start of a line, and should be easy to provide suggestions for based on position alone.

anderseknert:
Fixes #758

### DIFF
--- a/internal/lsp/completions/providers/default.go
+++ b/internal/lsp/completions/providers/default.go
@@ -1,0 +1,47 @@
+//nolint:dupl
+package providers
+
+import (
+	"strings"
+
+	"github.com/styrainc/regal/internal/lsp/cache"
+	"github.com/styrainc/regal/internal/lsp/types"
+	"github.com/styrainc/regal/internal/lsp/types/completion"
+)
+
+// Default will return completions for the default keyword when at the start of a line.
+type Default struct{}
+
+func (*Default) Run(c *cache.Cache, params types.CompletionParams, _ *Options) ([]types.CompletionItem, error) {
+	fileURI := params.TextDocument.URI
+
+	_, currentLine := completionLineHelper(c, fileURI, params.Position.Line)
+
+	if params.Context.TriggerKind == completion.Invoked && params.Position.Character == 0 {
+		return defaultCompletionItem(params), nil
+	}
+
+	// the user must type d before we provide completions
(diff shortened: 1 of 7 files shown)


internal/lsp/completions/providers/default.go
```diff
@@ -0,0 +1,46 @@
...
+
+// Default will return completions for the default keyword when at the start of a line.
+type Default struct{}
+
+func (*Default) Run(c *cache.Cache, params types.CompletionParams, _ *Options) ([]types.CompletionItem, error) {
+	fileURI := params.TextDocument.URI
+
+	_, currentLine := completionLineHelper(c, fileURI, params.Position.Line)
+
+	if params.Context.TriggerKind == completion.Invoked && params.Position.Character == 0 {
```
srenatus, line 19: This is assuming that indentation is always correct, isn't it?

So if I write
```rego
package foo
p {
_
```
where `_` is my cursor position, and I invoke completion, I'll get `default` proposed? But I suppose it's not simple to fix this... 🤔 

Ah, reading further, it'll only propose "default" when the situation would be this:

```rego
package foo
p {
d_
```

srenatus (approved):

internal/lsp/completions/providers/default.go
anderseknert: Yep, these assumptions are correct. Although VS Code auto-indents, and Regal will flag non-formatted files, you can force incorrect suggestions if you want to :) It's mostly just string based for now, but we could improve this later if necessary.

anderseknert: Yeah you'll either have to type out "d" at the start of the line or be at position 0 at the line and manually bring up the suggestions.

srenatus: OK I guess it's fine. You can trick it, but who would want to.

anderseknert: Mmm, ideally we'd have a better way to say "I'm at this position in this document, what's here?"

And that's doable, but it quickly becomes complex as for many of the answers (like "a an object key") you'll want to ask more questions, like "OK, which object?" and then it'll answer "the first function argument" and you'll want to ask "which function?" and so on... 😅

Issue #750, opened 2024-05-27
Completions: "fatal error: concurrent map iteration and map write" when trying completion


anderseknert:
Sadly, I'm not 100% sure what I did to cause this, as I was frantically testing different things trying to break completions, and while I succeeded in that, I'm not sure exactly _what_ caused the breakage. But it looks like we need a lock somewhere we don't.

```shell
fatal error: concurrent map iteration and map write

goroutine 34 [running]:
github.com/styrainc/regal/internal/lsp/completions/providers.(*RuleFromImportedPackageRefs).Run(0x140021e9a70?, 0x140003d5420, {{{0x1400e975ec0, 0x31}}, {0x4, 0xa}, {0x1, {0x0, 0x0}}}, 0x1400160c3b0?)
	/Users/anderseknert/git/styra/regal/internal/lsp/completions/providers/rulerefs.go:45 +0x238
github.com/styrainc/regal/internal/lsp/completions.(*Manager).Run(0x140003f62a0, {{{0x1400e975ec0, 0x31}}, {0x4, 0xa}, {0x1, {0x0, 0x0}}}, 0x1400160c3b0)
	/Users/anderseknert/git/styra/regal/internal/lsp/completions/manager.go:44 +0xd0
github.com/styrainc/regal/internal/lsp.(*LanguageServer).handleTextDocumentCompletion(0x1400039ab00, {0x105696840?, 0x1400160c1a0?}, 0x105696840?, 0x140198c7b00)
	/Users/anderseknert/git/styra/regal/internal/lsp/server.go:690 +0x118
github.com/styrainc/regal/internal/lsp.(*LanguageServer).Handle(0x1400039ab00, {0x10579f988, 0x140003b1c20}, 0x140003aa990, 0x140198c7b00)
	/Users/anderseknert/git/styra/regal/internal/lsp/server.go:143 +0x4f4
github.com/sourcegraph/jsonrpc2.(*HandlerWithErrorConfigurer).Handle(0x1400039ec20, {0x10579f988, 0x140003b1c20}, 0x140003aa990, 0x140198c7b00)
	/Users/anderseknert/go/pkg/mod/github.com/sourcegraph/jsonrpc2@v0.2.0/handler_with_error.go:21 +0x4c
github.com/sourcegraph/jsonrpc2.(*Conn).readMessages(0x140003aa990, {0x10579f988, 0x140003b1c20})
[...]

PR #760, opened 2024-05-28
lsp: Return copies of cache maps

charlieegan3:
Fixes: https://github.com/StyraInc/regal/issues/750

### DIFF
--- a/internal/lsp/cache/cache.go
+++ b/internal/lsp/cache/cache.go
@@ -2,6 +2,7 @@ package cache
 
 import (
 	"fmt"
+	"maps"
 	"os"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


internal/lsp/cache/cache.go
```diff
@@ -84,7 +84,12 @@ func (c *Cache) GetAllFiles() map[string]string {
 	c.fileContentsMu.Lock()
 	defer c.fileContentsMu.Unlock()
 
-	return c.fileContents
+	cp := make(map[string]string, len(c.fileContents))
```
srenatus, line 87: I believe we could use [`maps.Clone`](https://pkg.go.dev/golang.org/x/exp/maps#Clone) for this

anderseknert (approved):
LGTM, but @srenatus suggestion sounds good

internal/lsp/cache/cache.go
charlieegan3: Ohh go is cool now.

charlieegan3: I have added this in 51758077f96655a00f3ec60bc68e0d5f98fceac3

anderseknert (approved):

Issue #761, opened 2024-05-28
`fix` fails on non-rego files


oren-zohar:
### Summary
The new `fix` command fails to ignore non-rego files, and fails to complete

### In depth
Trying to use the new `fix` command in [elastic/cloudbeat/security-policies](https://github.com/elastic/cloudbeat/tree/main/security-policies) and I'm getting:

```console
❯ regal fix security-policies
\failed to fix: failed to generate linter input: failed to parse module security-policies/.gitignore: failed to parse module: 1 error occurred: security-policies/.gitignore:1: rego_parse_error: unexpected div token
        /output.json
        ^
```

Even though `regal lint` works and shows the following format issues:

```console
❯ regal lint security-policies
Rule:           messy-rule                                                                                                              
Description:    Messy incremental rule                                                                                                  
Category:       style                                                                                                                   
Location:       security-policies/bundle/compliance/cis_aws/rules/cis_4_1/rule.rego:30:1                                                
Text:           rule_evaluation = trail.at_least_one_trail_satisfied([required_pattern])                                                
Documentation:  https://docs.styra.com/regal/rules/style/messy-rule       
```

I tried to add the following part to `.regal/config.yaml`:
```yaml
# list of paths that should be ignored
ignore:
  files:
    - .gitignore
```

But that didn't help.

### Env information
Platform: `darwin/arm64`
OPA version: `0.61.0`
Regal version: `0.22.0`
Go Version: `1.22.0`

anderseknert:
Thanks for reporting that @oren-zohar! That looks like an oversight, and should be fairly simple to fix for the next release. We'll look into it 👍

charlieegan3:
Hey, I have https://github.com/StyraInc/regal/pull/762 open for this now. Feel free to test this out if you like.

I noticed that you have a built-in function that's called `contains` in the codebase, this is going to cause issues when applying the rego.v1 fix since `contains` will then be a keyword. It'd be ideal to rename or refactor this to avoid such issues.

charlieegan3:
One other option would be to use a quoted ref, but that'd not be recommended and is removed by the opa fmt command anyway. Thanks Anders for opening https://github.com/open-policy-agent/opa/issues/6766

PR #762, opened 2024-05-28
cmd/fix: Filter files loaded for fixing

charlieegan3:
This uses the same logic to filter files when loading them for fixing as used in linting.

Fixes: https://github.com/StyraInc/regal/issues/761


anderseknert (approved):
Thanks Charlie!

Issue #753, opened 2024-05-27
Completions: provide suggestions for data paths


anderseknert:
This works today when using imports and then referencing the import, but when typing out a reference to `data` directly, no suggestions are provided.

![Screenshot 2024-05-27 at 14 06 22](https://github.com/StyraInc/regal/assets/510711/92a4422f-da8c-4da8-b950-1de65adde5c3)

PR #769, opened 2024-05-29
lsp: Provide data ref completions in rules

charlieegan3:
Fixes: https://github.com/StyraInc/regal/issues/753

https://github.com/StyraInc/regal/assets/1774239/fe4684ff-8b8a-4891-ad40-19d12df06937

### DIFF
--- a/internal/lsp/completions/providers/rulerefs.go
+++ b/internal/lsp/completions/providers/rulerefs.go
@@ -8,11 +8,14 @@ import (
 	"github.com/styrainc/regal/internal/lsp/types/completion"
 )
 
-// RuleFromImportedPackageRefs is a completion provider that returns completions for
-// rules found in already imported packages.
-type RuleFromImportedPackageRefs struct{}
+// RuleRefs is a completion provider that returns completions for
+// rules found in:
+// - the current file
+// - imported packages
+// - any other files in the workspace under data.
+type RuleRefs struct{}
 
-func (*RuleFromImportedPackageRefs) Run(
+func (*RuleRefs) Run(
 	c *cache.Cache,
 	params types.CompletionParams,
 	_ *Options,
@@ -33,6 +36,9 @@ func (*RuleFromImportedPackageRefs) Run(
 		return nil, nil
 	}
 
+	words := patternWhiteSpace.Split(currentLine, -1)
+	lastWord := words[len(words)-1]
+
 	// some version of a parsed mod is needed here to filter refs to suggest
 	// based on import statements
 	mod, ok := c.GetModule(fileURI)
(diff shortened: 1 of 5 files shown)


internal/lsp/completions/providers/rulerefs.go
```diff
@@ -33,108 +36,159 @@ func (*RuleFromImportedPackageRefs) Run(
...
+		for _, ref := range refs {
+			// we are not interested in packages here, only rules
 			if ref.Kind == types.Package {
 				continue
 			}
 
-			// don't suggest refs of "private" rules, even
-			// if only just by naming convention
+			// don't suggest refs of "private" rules, even if this
+			// is only just a naming convention
```
anderseknert, line 59: Sneaky move, approving my PR only to correct me later hoping I wouldn't notice 😄

internal/lsp/completions/providers/rulerefs.go
```diff
@@ -33,108 +36,159 @@ func (*RuleFromImportedPackageRefs) Run(
...
-	for file, refs := range c.GetAllFileRefs() {
-		if file == fileURI {
-			continue
-		}
-
-		for key, ref := range refs {
-			// we are not interested in packages here, only the rules
+	for uri, refs := range c.GetAllFileRefs() {
+		for _, ref := range refs {
+			// we are not interested in packages here, only rules
```
anderseknert, line 53: I guess there are a few cases where you might want to refer to a package even in a rule body, like when doing dynamic policy composition:

```
some policy
some decision in data.policies[policy].deny
```
and so on... but I guess that's uncommon enough that it's still better to only suggest rules as a default.

internal/lsp/completions/providers/rulerefs.go
```diff
@@ -33,108 +36,159 @@ func (*RuleFromImportedPackageRefs) Run(
...
+
+			switch {
+			case ref.Kind == types.ConstantRule:
+				symbol = completion.Constant
+				detail = "Constant Rule"
+			case ref.Kind == types.Function:
+				symbol = completion.Function
+				detail = "Function"
+			case ref.Description == "Partial":
+				detail = "Partial path suggestion, continue typing for more suggestions."
```
anderseknert, line 139: cool!

anderseknert (approved):
Excellent! 👏

internal/lsp/completions/providers/rulerefs.go
charlieegan3: Ha, sorry...

internal/lsp/completions/providers/rulerefs.go
charlieegan3: Hmm, that is a thought. I guess this provide is called rulers and is already quite complex. I'd like to consider that for a new provider.

Issue here: https://github.com/StyraInc/regal/issues/771

Issue #768, opened 2024-05-29
Completions: suggest `input` on `i`


anderseknert:
While there is a more [complex issue](https://github.com/StyraInc/regal/issues/725) open for providing completions of `input` based on a provided schema, we should at least suggest `input` _at all_ when the user types `i` until then, as it is a valid item in pretty much any position.

PR #773, opened 2024-05-29
lsp: Rule head & input doc completion improvements

charlieegan3:
These new completion providers will complete keywords and common rule names in a limited number of but common scenarios.

https://github.com/StyraInc/regal/assets/1774239/670db180-b494-4ad8-a225-82aed14feb8d

also

Fixes https://github.com/StyraInc/regal/issues/768

https://github.com/StyraInc/regal/assets/1774239/c92ba19b-19c8-493f-814b-e0f8ab235b77


anderseknert (approved):
How do you even type so fast? 😄 I hope your fingers get some well needed rest, because tomorrow you're going back to the Regal salt mines. 

Awesome work!

PR #789, opened 2024-06-04
lsp: Show docs for diagnostics in Zed

charlieegan3:
This seems to be the best way to get clickable links to diagnostic docs when a linter error is displayed.

https://github.com/StyraInc/regal/assets/1774239/88cd5f89-e0b1-40a6-9a2c-1bc426467caf


anderseknert (approved):
Absolutely awesome!

Issue #774, opened 2024-05-29
Completions: suggest input based on previous references


anderseknert:
While a schema-based input provider would be awesome, and could help users with completions even in an entirely blank project, it also requires that the user provides a schema to begin with. This has many benefits, and we should see what we can do to encourage that via Regal. But that's a larger project, and in the meantime, I think a pretty good solution would be to base suggestions for `input` paths on references to input made previously in the project. I.e. if a user has written `input.request.method` in one rule, we store that as a candidate for future suggestions, and so on. We can probably even determine the data type of a path in most cases based on the path — if it has named children it's an object, index an array, and so on.

PR #794, opened 2024-06-05
lsp: implement used refs completions

charlieegan3:
This implements a completions provider that makes suggestions based on refs users have already typed into their files.

Fixes https://github.com/StyraInc/regal/issues/774

### DIFF
--- a/internal/lsp/completions/refs/used.go
+++ b/internal/lsp/completions/refs/used.go
@@ -1,0 +1,121 @@
+package refs
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/open-policy-agent/opa/ast"
+	"github.com/open-policy-agent/opa/bundle"
+	"github.com/open-policy-agent/opa/rego"
+
+	rbundle "github.com/styrainc/regal/bundle"
+	rio "github.com/styrainc/regal/internal/io"
+	"github.com/styrainc/regal/pkg/builtins"
+	"github.com/styrainc/regal/pkg/config"
+)
+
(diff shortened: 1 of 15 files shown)


internal/lsp/completions/refs/used.go
```diff
@@ -0,0 +1,121 @@
...
+					"capabilities": rio.ToMap(config.CapabilitiesForThisVersion()),
+				},
+			},
+		},
+	}
+
+	regoArgs := []func(*rego.Rego){
+		rego.ParsedBundle("regal", &regalRules),
+		rego.Module("module.rego", `
+package lsp.completions
```
anderseknert, line 44: I'd much prefer it if this was a .rego file embedded in the binary. That way we can work with it as any other Rego... lint, test, so on. Was there anything that prevented that?

anderseknert (approved):
Looks great! And I'm excited to try it out 😃 Just one comment / request, but it's not blocking.

srenatus (approved):
Nicely done. 👍

internal/lsp/completions/refs/used.go
charlieegan3: Yeah I should do that. It was only 3 lines or something at the start but it got more involved as I discovered some edge cases... 

Fixed in 88e76666c426da73f737aa4fa00750761d3614bb

srenatus (approved):

internal/lsp/completions/refs/used.go
anderseknert: Nice, thanks!

PR #798, opened 2024-06-05
Refactor: remove `name` function in favor of `ref_to_string`

anderseknert:
As `name` was rather ambiguous, and only gave the first part of the name, and `ref_to_string` is usable on any ref — let's just have that.


charlieegan3 (approved):
LGTM

PR #812, opened 2024-06-08
Return `null` in textDocument/diagnostic handler

### LINKED ISSUE #810: Error in latest Neovim handling textDocument/diagnostic response
See: https://github.com/StyraInc/regal/issues/578#issuecomment-2155815785

Looks like the latest version enabled some LSP features previously not supported, and a handler (in this case `textDocument/diagnostic`) returning an empty struct instead of `null` once again coming back to bite us. In our defence though, this is **not** a case where the specification defines `null` as a valid response! That and the fact that our response has been handled well by _all_ other clients we've tried suggests that it's probably Neovim getting things wrong here. Perhaps they'll fix that later as I'm sure they'll find issues with other servers too. But no matter who's wrong, it seems like all the clients I've tried accept `null` as a response too, so we can probably just make that change.

Thanks @folliehiyuki for reporting this.

anderseknert:
While the specification isn't clear here, Neovim apparently had issues dealing with the empty object response. And since all the other editors seem to handle a null response too, let's just go with that instead.

Fixes #810

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -1316,7 +1316,7 @@ func (*LanguageServer) handleTextDocumentDiagnostic(
 	// this is a no-op. Because we accept the textDocument/didChange event, which contains the new content,
 	// we don't need to do anything here as once the new content has been parsed, the diagnostics will be sent
 	// on the channel regardless of this request.
-	return struct{}{}, nil
+	return nil, nil
 }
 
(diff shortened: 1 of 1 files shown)


folliehiyuki:
I've tried this branch and can confirm that Neovim could handle the response gracefully. The reported error message doesn't pop up anymore.

anderseknert:
@folliehiyuki thanks!

charlieegan3 (approved):

PR #813, opened 2024-06-08
Use config.docs.resolve_url consistently

anderseknert:
Instead of hard-coding doc URLs.


anderseknert:
The check that fails do that on main too... I think as a result of rushing a PR in for the release. Anyway, it's fixed in another open PR, so will be gone once I rebase from main later.

charlieegan3 (approved):

PR #814, opened 2024-06-10
lsp: Use text edit for builtins providers

### LINKED ISSUE #808: Completions: built-in functions completions sometimes repeat what's already been typed in
Yeah, that is some description... but not sure how to best explain it. If you type "too much" before accepting a suggestion, it seems like the completion you get repeats some of what you've already typed in:

```rego
abc if object.uni|   # suggests object.union and object.union_n
```

Accepting the suggestion of `object.union_n` now results in `object` getting repeated:

```rego
abc if object.object.union_n
```

charlieegan3:
When not supplying a text edit, the editor will just insert the text instead. This is a problem when picking up a suggestion from a word started earlier.

Fixes #808 

before 

https://github.com/StyraInc/regal/assets/1774239/14e48d70-09ed-429c-84ac-28b18e12bf92

after

https://github.com/StyraInc/regal/assets/1774239/bb7b72e4-77d1-45e2-9e99-3bcc84bc5d1e

### DIFF
--- a/internal/lsp/completions/providers/builtins.go
+++ b/internal/lsp/completions/providers/builtins.go
@@ -47,6 +47,19 @@ func (*BuiltIns) Run(c *cache.Cache, params types.CompletionParams, _ *Options) 
 					Kind:  "markdown",
 					Value: hover.CreateHoverContent(builtIn),
 				},
+				TextEdit: &types.TextEdit{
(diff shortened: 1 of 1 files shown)


anderseknert (approved):
Nice! Thanks 👍

Issue #805, opened 2024-06-06
Regal fails when provided an empty configuration file


anderseknert:
An empty configuration file, or one with only comments in it, should just be treated as no configuration file. But instead, users will see something like: 

```
failed to decode user config from /Users/anderseknert/tmp/bug/.regal: EOF
exit status 1
```

We should just treat that as if no configuration file was present, or at most print a warning that an empty configuration file was found (but I'm not sure if even that is needed),

PR #815, opened 2024-06-10
lint/fix: roll over empty config

charlieegan3:
If a config file contains nothing or only comments, then we can roll over it with a more helpful message rather than erroring.

Fixes https://github.com/StyraInc/regal/issues/805


anderseknert (approved):
Awesome!

PR #816, opened 2024-06-10
lsp/refs: Just in time init for refs util

charlieegan3:
This was being slow to start and adding around 100ms to all regal command runs, even though this is only needed for the lsp.

This change makes the initialization of the pq var happen at the first call instead.

### DIFF
--- a/internal/lsp/completions/refs/used.go
+++ b/internal/lsp/completions/refs/used.go
@@ -26,7 +26,11 @@ var refNamesRego string
 //nolint:gochecknoglobals
 var pq *rego.PreparedEvalQuery
 
-func init() {
+// initialize prepares the rego query for finding ref names used in a module.
+// This is run and the resulting prepared query stored for performance reasons.
+// This function is only used by language server code paths and so init() is not
+// used.
+func initialize(ctx context.Context) {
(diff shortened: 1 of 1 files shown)


anderseknert (approved):
LGTM! I suppose pq isn't thread-safe, but worst case would be the initialize function is called more than once, which doesn't seem like a big deal.

srenatus (commented):
I wouldn't have been surprised to see a `*sync.Once` being used here 🤔 But maybe we don't need this.

anderseknert:
Since we both thought the same, let's use that :)

charlieegan3:
Thanks both, have updated.

internal/lsp/completions/refs/used.go
```diff
@@ -65,6 +73,10 @@ func init() {
 // See the rego above for more details on what's included and excluded.
 // This function is run when the parse completes for a module.
 func UsedInModule(ctx context.Context, module *ast.Module) ([]string, error) {
+	if pq == nil {
+		pqInitOnce.Do(initialize)
+	}
```
srenatus, line 78: Should be save to do
```go
pqInitOnce.Do(initialize)
```
shouldn't it? 🤔

PR #820, opened 2024-06-10
Cache rq binary

charlieegan3:


.github/workflows/build.yaml
```diff
@@ -37,7 +37,20 @@ jobs:
...
           static: ${{ matrix.os.static }}
       - run: npm install -g markdownlint-cli
+      - name: Restore rq cache
+        id: cache-rq
+        uses: actions/cache@v3
+        with:
+          path: ~/go/bin/rq
+          key: ${{ runner.os }}-rq
       - run: go install git.sr.ht/~charles/rq/cmd/rq@latest
+      - run: which rq
```
anderseknert, line 47: Was this for debugging?

anderseknert (commented):
Wow, that was fast! LGTM, but I suspect the `which` command is redundant?

.github/workflows/build.yaml
charlieegan3: Yeah, it's gone now. I have also updated to work for more than one macOS arch

.github/workflows/build.yaml
```diff
@@ -37,7 +37,19 @@ jobs:
           version: edge
           static: ${{ matrix.os.static }}
       - run: npm install -g markdownlint-cli
+      - name: Restore rq cache
+        id: cache-rq
+        uses: actions/cache@v3
+        with:
+          path: ~/go/bin/rq
+          key: ${{ runner.os }}-${{ runner.arch }}-go-rq
```
srenatus, line 45: Worth adding a version? I guess we can worry about that when there's an rq version we want and it's not getting here because of the cache 😅

charlieegan3: Yeah... I did wonder about that. Added now.

srenatus (approved):
Thanks!

charlieegan3:
Screenshot 2024-06-10 at 15 08 13


I am seeing this working now on this PR.

.github/workflows/build.yaml
```diff
@@ -37,7 +39,20 @@ jobs:
...
+      - name: Restore rq cache
+        id: cache-rq
+        uses: actions/cache@v4
+        with:
+          path: ~/go/bin/rq
+          key: ${{ runner.os }}-${{ runner.arch }}-go-rq-${{ env.RQ_VERSION }}
+      - run: go install git.sr.ht/~charles/rq/cmd/rq@${{ env.RQ_VERSION }}
+        if: steps.cache-rq.outputs.cache-hit != 'true'
+      - name: Cache rq binary
+        if: steps.cache-binary.outputs.cache-hit != 'true'
```
srenatus, line 51: ⚠️ typo
```suggestion
        if: steps.cache-rq.outputs.cache-hit != 'true'
```

PR #824, opened 2024-06-11
Check Regal version at start-up

### LINKED ISSUE #802: Log warning to Stderror when Regal is outdated
Regal should log to stderr if it checks the latest releases and determines that its version is not the latest. This should not happen in development builds.

charlieegan3:
This adds functionality to the regal LS and lint commands what will check the current version against the latest release in GH.

In order to keep the number of requests down, the version is cached to disk in the .regal dir. The task runs in another goroutine to keep things fast too.

Config has been updated with a features key, and
REGAL_FEATURES_REMOTE_CHECK_VERSION can also be set to disable to functionality.

Sharing for review, if we think it looks good, I'll add some docs and we can get this in.

Fixes #802

### DIFF
--- a/cmd/lint.go
+++ b/cmd/lint.go
@@ -18,6 +18,7 @@ import (
 	"github.com/open-policy-agent/opa/metrics"
 	"github.com/open-policy-agent/opa/topdown"
 
+	rbundle "github.com/styrainc/regal/bundle"
 	rio "github.com/styrainc/regal/internal/io"
 	regalmetrics "github.com/styrainc/regal/internal/metrics"
 	"github.com/styrainc/regal/internal/update"
@@ -245,7 +246,13 @@ func lint(args []string, params *lintCommandParams) (report.Report, error) {
 		m.Timer(regalmetrics.RegalConfigSearch).Stop()
 	}
 
-	regal := linter.NewLinter().
+	// regal rules are loaded here and passed to the linter separately
+	// as the configuration is also used to determine feature toggles
+	// and the defaults from from the data.yaml here.
+	regalRules := rio.MustLoadRegalBundleFS(rbundle.Bundle)
+
+	regal := linter.NewEmptyLinter().
+		WithAddedBundle(regalRules).
 		WithDisableAll(params.disableAll).
 		WithDisabledCategories(params.disableCategory.v...).
 		WithDisabledRules(params.disable.v...).
@@ -315,18 +322,21 @@ func lint(args []string, params *lintCommandParams) (report.Report, error) {
 	}
 
 	go func() {
-		dir := ""
-		if regalDir != nil {
-			dir = regalDir.Name()
+		mergedConfig, err := config.LoadConfigWithDefaultsFromBundle(&regalRules, &userConfig)
+		if err != nil {
+			if params.debug {
+				log.Printf("failed to merge user config with default config when checking version: %v", err)
+			}
+			return
 		}
 
-		if userConfig.Features.Remote.CheckVersion &&
+		if mergedConfig.Features.Remote.CheckVersion &&
 			os.Getenv(update.CheckVersionDisableEnvVar) != "false" {
 			update.CheckAndWarn(update.Options{
 				CurrentVersion: version.Version,
 				CurrentTime:    time.Now().UTC(),
 				Debug:          params.debug,
-				StateDir:       dir,
+				StateDir:       config.GlobalDir(),
 			}, os.Stderr)
 		}
 	}()
--- a/cmd/lint.go
+++ b/cmd/lint.go
@@ -20,10 +20,12 @@ import (
 
 	rio "github.com/styrainc/regal/internal/io"
 	regalmetrics "github.com/styrainc/regal/internal/metrics"
+	"github.com/styrainc/regal/internal/update"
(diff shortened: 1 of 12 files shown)


cmd/lint.go
```diff
@@ -312,6 +316,23 @@ func lint(args []string, params *lintCommandParams) (report.Report, error) {
 		m.Timer(regalmetrics.RegalConfigParse).Stop()
 	}
 
+	go func() {
+		dir := ""
+		if regalDir != nil {
+			dir = regalDir.Name()
+		}
+
+		if userConfig.Features.RemoteFeatures.CheckVersion &&
```
charlieegan3, line 325: Config loading is handled in a few places, lint, fix and the lsp, in each place, there are minor differences in when and how the config is loaded. Depending on config for this version check functionality makes it more tightly coupled, and harder to just run anywhere. So not sure, about the config stuff in this PR, but we can remove if we don't like it and just offer the env var.

internal/update/update.rego
srenatus, line 1: 👏

internal/update/update.go
```diff
@@ -0,0 +1,186 @@
...
+
+	"github.com/open-policy-agent/opa/rego"
+
+	_ "embed"
+)
+
+//go:embed update.rego
+var updateModule string
+
+const CheckAndWarnIgnoreVariable = "REGAL_FEATURES_REMOTE_CHECK_VERSION"
```
srenatus, line 24: The const is called IgnoreVariable, but the env var name is a feature flag... So it's unclear if this is something to opt-out or to opt-in to, from the env name alone. 

My recent favorite telemetry transparency example is observable framework: https://observablehq.com/framework/telemetry. They have a `_DISABLE` var:
> You can [opt-out of telemetry](https://observablehq.com/framework/telemetry#disabling-telemetry) by setting the `OBSERVABLE_TELEMETRY_DISABLE` environment variable to `true`.

pkg/config/config.go
```diff
@@ -167,6 +177,14 @@ func FromMap(confMap map[string]any) (Config, error) {
 	return conf, nil
 }
 
+func (config *Config) SetDefaults() {
```
anderseknert, line 180: Can't we set this in the the default configuration instead? 
https://github.com/StyraInc/regal/blob/main/bundle/regal/config/provided/data.yaml

cmd/lint.go
```diff
@@ -312,6 +316,23 @@ func lint(args []string, params *lintCommandParams) (report.Report, error) {
 		m.Timer(regalmetrics.RegalConfigParse).Stop()
 	}
 
+	go func() {
+		dir := ""
+		if regalDir != nil {
+			dir = regalDir.Name()
```
anderseknert, line 322: I'm a little weary about having `regal lint` change the worktree for users. They can .gitignore it of course, but it's an extra annoyance. I wonder if we can use something like `~/.config/regal` as the default instead? We'd have to create the directory though, and fall back gracefully if we can't (perhaps by _then_ trying the `.regal` in the project directory?).

anderseknert (commented):
Change looks good to me! With a few things to consider / discuss. 

Also, we'd need some docs on this, right?

pkg/config/config.go
charlieegan3: Not trivial, but this is the right way you're right...

72dcabf6d09415a7adee1bd7759cb4f099578222 is the commit that adjusts this PR to do that, lmk if you think that looks ok.

internal/update/update.go
```diff
@@ -21,7 +21,7 @@ import (
 //go:embed update.rego
 var updateModule string
 
-const CheckAndWarnIgnoreVariable = "REGAL_FEATURES_REMOTE_CHECK_VERSION"
+const CheckVersionDisableEnvVar = "REGAL_FEATURES_REMOTE_CHECK_VERSION_DISABLE"
```
anderseknert, line 24: I'd go with REGAL_DISABLE_VERSION_CHECK, but naming is hard :D

internal/update/update.go
charlieegan3: Thanks, I have adjusted the name for this one now.

internal/update/update.go
charlieegan3: Ok, let's go with that, it's shorter and reads more for humans than for computers.

My thinking with the other one was:

- it's for regal
- then it's for regal features
- then it's for remote features
- then it's for check-version
- then it's to disable it.

but it's perhaps verbose.

@srenatus had thoughts here too so will let him comment, but have used REGAL_DISABLE_VERSION_CHECK for now.

charlieegan3:
This is all updated again, I just need to do some docs for it.

charlieegan3:
dedd3bbdc8952912468e210b047a742b7fe5191a adds the docs.

cmd/lint.go
charlieegan3: added in https://github.com/StyraInc/regal/pull/824/commits/72dcabf6d09415a7adee1bd7759cb4f099578222

anderseknert:
There's some build failures. I just realized it might be good to mention this briefly in the README too, and provide a link to the docs page on that!

charlieegan3:
Have fixed the md lint error and updated readme too.

PR #831, opened 2024-06-12
Don't call completion providers inside of comments

### LINKED ISSUE #826: Code completion in comments (Zed)
For some reason, Zed seems to give code completion inside of comments. We should investigate why that differs from VS Code, where I don't seem to be able to reproduce this.

Screenshot 2024-06-12 at 14 14 56

anderseknert:
It seems like this is handled client side by some editors, but Zed triggered completion requests inside of comments. While we could have each provider handle this, I think an early exit in the manager already is a better solution, as there's unlikely to be any provider that wants to do suggestions in a comment. If that changes in the future, we'll handle it then.

Fixes #826

### DIFF
--- a/internal/lsp/completions/manager.go
+++ b/internal/lsp/completions/manager.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/styrainc/regal/internal/lsp/cache"
 	"github.com/styrainc/regal/internal/lsp/completions/providers"
+	"github.com/styrainc/regal/internal/lsp/rego"
 	"github.com/styrainc/regal/internal/lsp/types"
 )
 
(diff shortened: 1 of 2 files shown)


internal/lsp/completions/manager.go
```diff
@@ -70,3 +77,19 @@ func (m *Manager) Run(params types.CompletionParams, opts *providers.Options) ([
...
 }
+
+func (m *Manager) isInsideOfComment(params types.CompletionParams) bool {
+	if module, ok := m.c.GetModule(params.TextDocument.URI); ok {
+		for _, comment := range module.Comments {
+			cp := rego.PositionFromLocation(comment.Location)
+
+			if cp.Line == params.Position.Line {
+				if cp.Character <= params.Position.Character {
+					return true
```
charlieegan3, line 88: This is nice!

charlieegan3 (approved):

PR #834, opened 2024-06-13
Don't suggest `if` or `contains` following import

### LINKED ISSUE #830: Completions: `if` and `contains` should not be suggested following import
Same goes for "package".

Screenshot 2024-06-12 at 23 33 20

While we're at it — we don't seem to provide completions following "import " when invoked manually.

anderseknert:
Also:
- improve handling of manually invoked suggestions
- don't suggest "import" after "import "

Fixes #830

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -11,6 +11,7 @@ linters:
     - varnamelen
     - nonamedreturns
     - testpackage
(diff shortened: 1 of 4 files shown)


charlieegan3 (approved):
thanks for this, looks good!

PR #835, opened 2024-06-13
Rule: `if-object-literal`

### LINKED ISSUE #822: Rule: `if-object-literal`
Came across this old issue in the OPA backlog: https://github.com/open-policy-agent/opa/issues/5294
And since that's unlikely to be fixed in OPA, we should at least make sure Regal flags it.

```rego
package policy

import rego.v1

allow if {input.message: "world"}
```

Note that we currently have the [if-empty-object](https://docs.styra.com/regal/rules/bugs/if-empty-object) rule to cover the case OPA previously would fail for being an empty rule body.. but doesn't do anymore. While we _could_ retrofit that rule to cover _any_ object following an `if`, we'd have to do so much work around the docs, redirects and whatnot, that it's probably a better option to add a new rule that checks for `if` followed by a _non-empty_ object.

Question is if there is _ever_ a valid `if { ..object.. }` case? I can't think of one now, but happy to be proven wrong.

anderseknert:
This replaces `if-empty-object`, and covers any object literal.

Fixes #822


charlieegan3 (commented):
Makes sense!

charlieegan3 (approved):

Issue #811, opened 2024-06-08
Completions: rule head (if, contains, etc) improvements


anderseknert:
As this will be one of the more common completion suggestions, I think it's worth spending some time making it perfect.

Currently both the `if` and `contains` completions require the user to type `i` or `c` respectively to show up. Since there isn't much else that can follow a rule name, we should just suggest them already after the first whitespace. This would also help people new to Rego know what's possible at that location, without having to know about `if` or `contains`. On that note — we should add some docs on both of those keywords to show up in the suggestions! So that you can learn as you type.

> there isn't much else that can follow a rule name

The one (?) exception is of course `:=`, which we should also add a suggestion for, with an explanation that is the value returned by the rule if evaluation succeeds.

PR #836, opened 2024-06-13
Better support rule head keyword completions

charlieegan3:
This handles the manual trigger case better.

Fixes https://github.com/StyraInc/regal/issues/811

https://github.com/StyraInc/regal/assets/1774239/e6d7bb88-ca6f-4781-aa8b-71d371e0f7a1


anderseknert (approved):
Very nice!

PR #840, opened 2024-06-17
Completion suggestions for variables in local scope

### LINKED ISSUE #792: Completions: add suggestions for vars in local scope
This one might be trickier to do, but certainly helpful!

```rego
allow if {
    rules_with_annotations := [
        # ... some processing here
    ]

    # should suggest "rules_with_annotations"
    r|

    # should suggest "rules_with_annotations"
    count(r|
}
```

anderseknert:
This is the first provider that is using Rego policy to determine suggestions!

That makes the PR more extensive than normally, but much of this will be reusable for future providers, or even existing ones we may choose to convert.

Fixes #792

### DIFF
--- a/internal/lsp/completions/providers/policy_test.go
+++ b/internal/lsp/completions/providers/policy_test.go
@@ -1,0 +1,50 @@
+package providers
+
+import (
+	"testing"
+
+	"github.com/open-policy-agent/opa/ast"
+
+	"github.com/styrainc/regal/internal/lsp/cache"
+	"github.com/styrainc/regal/internal/lsp/types"
+)
+
+func TestPolicyProvider(t *testing.T) {
+	t.Parallel()
+
+	locals := NewPolicy()
+	policy := `package p
+
+import rego.v1
+
+allow if {
+	user := data.users[0]
+	# try completion on next line
+	roles := u
+}
+`
+	module := ast.MustParseModule(policy)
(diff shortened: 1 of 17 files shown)


internal/lsp/completions/providers/policy_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/open-policy-agent/opa/ast"
+
+	"github.com/styrainc/regal/internal/lsp/cache"
+	"github.com/styrainc/regal/internal/lsp/types"
+)
+
+func TestPolicyProvider(t *testing.T) {
+	t.Parallel()
+
+	locals := NewPolicy()
```
charlieegan3, line 15: I wonder if there's a way here to make it so that only one of the Rego evaluators is run, so that when we have more, testing them individually is easier.

charlieegan3 (approved):
This looks really great! Looking forward to seeing which of the other providers we can migrate to it, too.

internal/lsp/completions/providers/policy_test.go
anderseknert: The Rego providers are "all" (well, in the future anyway) thoroughly tested in Rego. What we do here should be closer to an integration test... but you're still right — it'd be good to be able to have at least a very basic such test for each provider. Let's look into that as we add more.

PR #837, opened 2024-06-13
lsp: perform no operation when processing events for ignored files

charlieegan3:
Fixes https://github.com/StyraInc/regal/issues/766

Uploading Screen Recording 2024-06-13 at 17.57.19.mov…

PR #845, opened 2024-06-17
[lsp] Add support for ignored files in ls formatting

charlieegan3:
Adding functionality to manage the contents of ignored files when formatting documents.

Addresses some issues with ignored files missed in https://github.com/StyraInc/regal/pull/837

### DIFF
--- a/README.md
+++ b/README.md
@@ -449,6 +449,11 @@ rules:
 
 ### Ignoring Files Globally
 
+**Note**: Ignoring files will disable most language server features
+for those files. Only formatting will remain available.
+Ignored files won't be used for completions, linting, or definitions
+in other files.
+
 If you want to ignore certain files for all rules, you can use the global ignore attribute in your configuration file:
 
 ```yaml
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -449,6 +449,8 @@ rules:
 
 ### Ignoring Files Globally
 
+**Note**: Ignoring files globally will also disable language server features for those files too.
```
anderseknert, line 452: Nit for sure, but I don't think you need the ending "too" after having used "also" before :P

anderseknert (approved):
LGTM! We can extend this to cover more "local" operations later if ever requested.

README.md
charlieegan3: I made it a little more explicit: "**Note**: Ignoring files will disable most language server features for those files. Only formatting will remain available. Ignored files won't be used for completions, linting, or definitions in other files."

anderseknert: Excellent!

PR #854, opened 2024-06-19
Fix `top-level-iteration` false positive with ref head vars

### LINKED ISSUE #852: False positive in `top-level-iteration` when using ref head rules
```rego
package play

import rego.v1

other := {1: "foo", 2: "bar", 3: "baz"}

something[x] := other[x] if {
    some x in [1, 2, 3]
}
```
While it's iteration, it's not being done at the top level, and should not be reported as such.

anderseknert:
Fixes #852


charlieegan3 (approved):
LGTM!

Issue #843, opened 2024-06-17
Completion suggestions should not contain duplicates


anderseknert:
https://github.com/StyraInc/regal/assets/510711/fdc57615-d524-485b-a65c-724f04fd754c

Also from this, I think we'd need to either update the logic of "previously used ref" as this ref is known not to have been used anywhere else — or we'd need to update the description to reflect what it actually suggests this for. In this case however, that suggestion should be hidden in favor of the "rule ref" suggestion.

Issue #832, opened 2024-06-12
Completions: avoid suggesting recursive function call


anderseknert:
It would be good if the name of the "current" function was removed from the suggestions, as this is a recursive call. Unlikely to be doable using text-based context matching, but we have the parsed module, so..

<img width="239" alt="Screenshot 2024-06-13 at 00 07 20" src="https://github.com/StyraInc/regal/assets/510711/cf850394-932b-4405-8282-47e05db79635">

PR #855, opened 2024-06-19
Clear the store when reloading config

charlieegan3:
This also ensures that config is loaded before doing the initial load of the workspace. This means that loaded ignored files do not need to then be immediately dropped.

Also

Fixes https://github.com/StyraInc/regal/issues/832

Fixes https://github.com/StyraInc/regal/issues/843


anderseknert (approved):
LGTM! Great stuff!

PR #859, opened 2024-06-20
Precompute refs in file at update

charlieegan3:
Running `workspace_rule_refs` on each key change was proving to be too slow, so I've used a mechanism more similar to the old provider where the refs are computed ahead of time when the file contents changes.


anderseknert (approved):
Thanks!

PR #869, opened 2024-06-24
Completions: package name now suggested from any path component

### LINKED ISSUE #846: Improve package name suggestion based on directory path
We currently suggest the whole path no matter what the user typed in. But while we know where the project root is, we don't know where the _bundle root_ might be, which is really where the path should start from. 

So in the case of Regal, where the bundle root is `bundle`, but the first patch component is `regal`:

Screenshot 2024-06-17 at 15 35 59

We should suggest completions based on the first word the user is typing in. So in this case, it would be better to suggest `regal.lsp.completion.providers.default` than `bundle.regal.lsp.completion.providers.default`.

### LINKED ISSUE #856: Completions: package name suggestion should never be an empty string
This would best be fixed when moving the provider to Rego, but just noting it down here in case I forget.

When you create a new rego file in the workspace root, the package name provider will currently suggest "" as the package name. It should just not suggest anything in that case.

anderseknert:
And fix bug that would sometimes suggest "" as the package name.

Oh, and the provider is now written in Rego :)

Fixes #846
Fixes #856


bundle/regal/lsp/completion/providers/packagename/packagename.rego
```diff
@@ -0,0 +1,51 @@
...
+	position := location.to_position(input.regal.context.location)
+	line := input.regal.file.lines[position.line]
+
+	invoke_suggestion(line, position)
+
+	ps := input.regal.context.path_separator
+
+	abs_dir := base(input.regal.file.name)
+	rel_dir := trim_prefix(abs_dir, input.regal.context.workspace_root)
+	fix_dir := replace(replace(trim_prefix(rel_dir, ps), ".", "_"), ps, ".")
```
srenatus, line 18: oh my 🙈

srenatus (approved):
Looks good to me. I haven't tried it locally ⚠️ 

I don't suppose this would deal with directories like
```
foo/x-y/baz
```
properly, would it? I.e. Propose `package foo["x-y"].baz`?

bundle/regal/lsp/completion/providers/packagename/packagename.rego
anderseknert: heh, yeah, not sure how that could be made more clear... perhaps an extra intermediate assignment or something

anderseknert:
> I don't suppose this would deal with directories like

That's correct. We should have some general way of rewriting refs to account for that. I was hoping someone would just fix that "keywords in refs need quoting" in OPA first, so we'd at least have one less thing to handle here 😅

PR #870, opened 2024-06-25
Add snippets provider

anderseknert:
Snippets are special inserts that allow users to enter values and press tab to move to the next location.

https://github.com/StyraInc/regal/assets/510711/04c412a8-1933-4a0d-b041-69059927bd81

https://code.visualstudio.com/docs/editor/userdefinedsnippets

Add snippets for:
- `some v in coll`,
- `some k, v in coll`
- `every v in coll`
- `every k, v in coll`

Got interested in trying this after learning that Zed supports snippets in the latest release. I've tried and confirmed this works in both VS Code and in Zed.

### DIFF
--- a/bundle/regal/lsp/completion/providers/snippet/snippet_test.rego
+++ b/bundle/regal/lsp/completion/providers/snippet/snippet_test.rego
@@ -1,0 +1,110 @@
+package regal.lsp.completion.providers.snippet_test
+
+import rego.v1
+
+import data.regal.lsp.completion.providers.snippet as provider
+import data.regal.lsp.completion.providers.utils_test as util
+
+# regal ignore:rule-length
+test_snippet_completion_on_typing if {
+	policy := `package policy
+
+import rego.v1
+
+allow if {
+	e
+}`
+	items := provider.items with input as util.input_with_location(policy, {"row": 6, "col": 2})
+	items == {
+		{
+			"detail": "every key-value iteration",
+			"insertTextFormat": 2,
+			"kind": 15,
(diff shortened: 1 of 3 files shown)


srenatus (approved):

srenatus:
Nice how little code is needed to add this :+1:

bundle/regal/lsp/completion/providers/snippet/snippet_test.rego
```diff
@@ -0,0 +1,110 @@
...
+				},
+			},
+		},
+		{
+			"detail": "some key-value iteration",
+			"insertTextFormat": 2,
+			"kind": 15,
+			"label": "some key-value iteration (snippet)",
+			"textEdit": {
+				"newText": "some ${1:key}, ${2:value} in ${3:collection}\n$0",
```
charlieegan3, line 89: Didn't realise this was so simple, we might want to update things like package completions to use the same format.

charlieegan3 (approved):

bundle/regal/lsp/completion/providers/snippet/snippet_test.rego
anderseknert: Totally! This should be trivial to extend with more cases. I just wanted to start from somewhere.

anderseknert: We could also do things like more advanced function calls (`http.send` et. al.). where we pre-fill some example values. Like how the VS Code docs have an example where they do this for some AJAX call.

PR #877, opened 2024-07-01
perf: refactor to avoid excessive `walk`ing

anderseknert:
Store result of vars found in `walk` into an intermediate object, keyed by rule index and the variable's "context".

The result is quite an improvement:

```
❯ hyperfine --warmup 2 'regal_main lint bundle' 'regal_new lint bundle'
Benchmark 1: regal_main lint bundle
  Time (mean ± σ):      4.387 s ±  0.174 s    [User: 31.888 s, System: 0.906 s]
  Range (min … max):    4.260 s …  4.820 s    10 runs

Benchmark 2: regal_new lint bundle
  Time (mean ± σ):      3.502 s ±  0.120 s    [User: 24.250 s, System: 0.688 s]
  Range (min … max):    3.308 s …  3.700 s    10 runs

Summary
  regal_new lint bundle ran
    1.25 ± 0.07 times faster than regal_main lint bundle
```

### DIFF
--- a/bundle/regal/ast/ast.rego
+++ b/bundle/regal/ast/ast.rego
@@ -122,11 +122,7 @@ all_rules_refs contains value if {
 # scope: document
 all_refs contains value if some value in all_rules_refs
 
-all_refs contains value if {
-	walk(input.imports, [_, value])
-
-	is_ref(value)
-}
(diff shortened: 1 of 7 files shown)


srenatus (approved):
Nicely done ✨

PR #878, opened 2024-07-01
Rule: `pointless-reassignment`

### LINKED ISSUE #867: Rule: `pointless-reassignment`
Since values and variables are immutable in Rego, re-assigning a variable from one name to another doesn't come with any benefits.

**Avoid**
```rego
allow if {
    users := all_users
    # do something with users
}
```

Re-assigning a long ref to a variable should however be considered OK:

```rego
allow if {
    users := data.context.permissions.users
    # do something with users
}
```

anderseknert:
Also:
- fix the one violation of this rule found in our code :P
- add metadata annotations to a bunch of rules

Fixes #867

### DIFF
--- a/bundle/regal/rules/imports/circular_import.rego
+++ b/bundle/regal/rules/imports/circular_import.rego
@@ -43,11 +43,10 @@ aggregate_report contains violation if {
 
 	sorted_group := sort(g)
 
-	location := [e |
+	location := [loc |
 		some m1 in sorted_group
 		some m2 in sorted_group
 		some loc in package_locations[m1][m2]
-		e := loc
 	][0]
 
 	violation := result.fail(
--- a/docs/rules/style/pointless-reassignment.md
+++ b/docs/rules/style/pointless-reassignment.md
@@ -1,0 +1,62 @@
+# pointless-reassignment
+
+**Summary**: Pointless reassignment of variable
+
+**Category**: Style
+
+**Avoid**
+```rego
+package policy
+
+allow if {
+    users := all_users
+    any_admin(users)
+}
+```
+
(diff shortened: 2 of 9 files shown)


docs/rules/style/pointless-reassignment.md
```diff
@@ -0,0 +1,62 @@
...
+```rego
+package policy
+
+allow if {
+    users := input.context.permissions.users
+    any_admin(users)
+}
+```
+
+This rule does not consider that a violation.
```
charlieegan3, line 44:
```suggestion
This rule does not consider such examples violations.
```

charlieegan3: I wonder what if I a had a rule like `foo_bar_whizzz_bang_bar_foo_baz_v1_foo`, and then reassigned it for readability, would this be good to ignore too?

bundle/regal/rules/imports/circular_import.rego
```diff
@@ -43,11 +43,10 @@ aggregate_report contains violation if {
 
 	sorted_group := sort(g)
 
-	location := [e |
+	location := [loc |
 		some m1 in sorted_group
 		some m2 in sorted_group
 		some loc in package_locations[m1][m2]
-		e := loc
```
charlieegan3, line 50: my handiwork!

charlieegan3 (approved):

docs/rules/style/pointless-reassignment.md
anderseknert: I'd say the issue then would be you naming a rule or var like that in the first place :D

PR #883, opened 2024-07-02
Rule: `argument-always-wildcard`

### LINKED ISSUE #872: Rule: `argument-always-wildcard`
Encountered in the wild, but most likely quite uncommon. If all definitions of a function have a wildcard argument at the same position, that argument should just be removed.

```rego
fun(x, _) if {
    # ...
}

fun(x, _) if {
    # ...
}
```

I think this is most likely to happen in refactoring, where you might remove the one definition where it's actually used.

anderseknert:
I did not expect that I'd have to fix any issues in Regal as a result of this, but that was a rather nice surprise!

Fixes #872

### DIFF
--- a/docs/rules/bugs/argument-always-wildcard.md
+++ b/docs/rules/bugs/argument-always-wildcard.md
@@ -1,0 +1,94 @@
+# argument-always-wildcard
+
+**Summary**: Argument is always a wildcard
+
+**Category**: Bugs
+
+**Avoid**
+```rego
+package policy
+
+import rego.v1
+
+# there's only one definition of the last_name function in
+# this package, and the second argument is never used
+last_name(name, _) := lname if {
+    parts := split(name, " ")
+    lname := parts[count(parts) - 1]
+}
+```
+
+**Prefer**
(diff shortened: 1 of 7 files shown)


docs/rules/bugs/argument-always-wildcard.md
```diff
@@ -0,0 +1,92 @@
...
+
+**Category**: Bugs
+
+**Avoid**
+```rego
+package policy
+
+import rego.v1
+
+last_name(name, _) := lname if {
```
charlieegan3, line 13: would a clearer example be a function with two heads?

charlieegan3: Or a comment to explain it's the end of the file after the single head.

charlieegan3 (approved):

docs/rules/bugs/argument-always-wildcard.md
anderseknert: It's provided later in the doc, but I'll add a comment above this to clarify

PR #884, opened 2024-07-02
Use OSSF scorecard

anderseknert:

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -1,0 +1,44 @@
+name: Scorecard supply-chain security
+on:
+  branch_protection_rule:
+  schedule:
+    - cron: '16 9 * * 2'
+  push:
+    branches: [ "main" ]
+
+permissions: read-all
+
+jobs:
+  analysis:
+    name: Scorecard analysis
+    runs-on: ubuntu-latest
+    permissions:
+      # Needed to upload the results to code-scanning dashboard.
+      security-events: write
+      # Needed to publish results and get a badge (see publish_results below).
+      id-token: write
+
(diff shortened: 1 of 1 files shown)


.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,44 @@
+name: Scorecard supply-chain security
+on:
+  branch_protection_rule:
+  schedule:
+    - cron: '16 9 * * 2'
```
srenatus, line 5: Now that's a funny-looking cron. crontab.guru says
> “At 09:16 on Tuesday.” 

OK fine.

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,44 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      # Needed to upload the results to code-scanning dashboard.
+      security-events: write
+      # Needed to publish results and get a badge (see publish_results below).
+      id-token: write
+
+    steps:
+      - name: "Checkout code"
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
```
srenatus, line 23: I like this. We should do this more often.

srenatus (approved):
LGTM

.github/workflows/scorecard.yml
anderseknert: Oddly specific 😆

.github/workflows/scorecard.yml
anderseknert: Indeed... something tells me that's one of the things this tool is going to tell us 😅

PR #886, opened 2024-07-02
Fix token permissions

anderseknert:
Use `read-all` at the top level. Not sure if we actually need anything more somewhere, but I guess this is one way to find out.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -11,6 +11,8 @@ on:
   pull_request:
   workflow_dispatch:
 
+permissions: read-all
+
 jobs:
   build-matrix:
     env:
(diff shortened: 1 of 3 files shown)


srenatus:
I'd expect that release creation through goreleaser _might_ need more. But let's find out? 🤔

srenatus (approved):

anderseknert:
@srenatus yeah, will push a release this week, so shouldn’t have forgotten about it by then :)

charlieegan3:
FWIW, I added 34202b5a9b82a3b611fa588bb93675a10544912e and 31bee8d467c3d2d416952e5a116dd8060b079478 to set the permissions to:

```
permissions:
   # this is needed to create a release
   contents: write
   # this is needed to create artifacts in ghcr
   packages: write
```

based on the details here: https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

I'd not have worked this out without trial and error I think!

PR #893, opened 2024-07-02
Rule: `var-shadows-builtin`

### LINKED ISSUE #823: Rule: `var-shadows-builtin`
The [rule-shadows-builtin](https://docs.styra.com/regal/rules/bugs/rule-shadows-builtin) rule covers rules named to shadow built-in functions or namespaces. It would be a good addition to have a similar rule that tracked shadowing in local vars.  

Related: https://github.com/open-policy-agent/opa/issues/3729

anderseknert:
Also in this PR is a slight change of behavior, where variables found in comprehension heads are no longer treated as assigned. This as they _aren't_, and the assignment happens in the comprehension body (or outside the comprehension). This change will uncover a few more `use-some-for-output-vars` issues in policy, which was also fixed as part of this PR.

Fixes #823

### DIFF
--- a/docs/rules/bugs/var-shadows-builtin.md
+++ b/docs/rules/bugs/var-shadows-builtin.md
@@ -1,0 +1,61 @@
+# var-shadows-builtin
+
+**Summary**: Variable shadows built-in
+
+**Category**: Bugs
+
+**Avoid**
+```rego
+package policy
+
+import rego.v1
+
+# variable `http` shadows `http.send` built-in function
+allow if {
+    http := startswith(input.url, "http://")
+    # do something with http
+}
+```
+
+**Prefer**
+```rego
+package policy
+
+import rego.v1
+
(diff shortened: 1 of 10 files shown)


docs/rules/bugs/var-shadows-builtin.md
```diff
@@ -0,0 +1,61 @@
...
+
+**Avoid**
+```rego
+package policy
+
+import rego.v1
+
+# variable `http` shadows `http.send` built-in function
+allow if {
+    http := true
```
charlieegan3, line 15: Perhaps `http := startswith(url, "http://")` or similar, just to make it more real feeling?

charlieegan3 (approved):

docs/rules/bugs/var-shadows-builtin.md
anderseknert: Alright!

PR #894, opened 2024-07-02
Fix negative number returned by rulerefs provider issue

### LINKED ISSUE #887: Moving back to edit before accepted suggestion results in error
It seems like even an accepted suggestion is "remembered", and moving back on the line to make an edit will calculate the offset based off of that. This results in the offset being negative, which it should never be (it's stored as `uint`):

https://github.com/StyraInc/regal/assets/510711/be90e045-36f6-4dcc-8814-337ad138dbbb

anderseknert:
Since character position is a `uint`, this would fail when returned to the Go handler.

Also:
- Remove some prints we did to debug performance
- Some cleanups to align rulerefs with other providers

Fixes #887

### DIFF
--- a/bundle/regal/lsp/completion/main.rego
+++ b/bundle/regal/lsp/completion/main.rego
@@ -10,4 +10,9 @@ import rego.v1
 # METADATA
 # description: main entry point for completion suggestions
 # entrypoint: true
-items contains data.regal.lsp.completion.providers[_].items[_]
+items contains item if {
+	some provider
+	completion := data.regal.lsp.completion.providers[provider].items[_]
+
+	item := object.union(completion, {"_regal": {"provider": provider}})
+}
--- a/bundle/regal/lsp/completion/providers/rulerefs/rulerefs.rego
+++ b/bundle/regal/lsp/completion/providers/rulerefs/rulerefs.rego
@@ -10,15 +10,13 @@ ref_is_internal(ref) if contains(ref, "._")
 
 default determine_ref_prefix(_) := ""
 
-determine_ref_prefix(word) := word if {
-	word != ":="
-}
+determine_ref_prefix(word) := word if word != ":="
 
 position := location.to_position(input.regal.context.location)
 
 line := input.regal.file.lines[position.line]
 
-last_word := regal.last(regex.split(`\s+`, trim_space(line)))
+word := location.ref_at(line, input.regal.context.location.col)
 
 workspace_rule_refs contains ref if {
 	some refs in data.workspace.defined_refs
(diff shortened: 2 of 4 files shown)


bundle/regal/lsp/completion/providers/rulerefs/rulerefs.rego
```diff
@@ -10,15 +10,13 @@ ref_is_internal(ref) if contains(ref, "._")
...
-	word != ":="
-}
+determine_ref_prefix(word) := word if word != ":="
 
 position := location.to_position(input.regal.context.location)
 
 line := input.regal.file.lines[position.line]
 
-last_word := regal.last(regex.split(`\s+`, trim_space(line)))
+word := location.ref_at(line, input.regal.context.location.col)
```
anderseknert, line 19: This is the actual fix. As an added bonus — and quite a nice one too! — we'll now also get completions for refs even when the ref is not the last thing being typed on the line.

bundle/regal/lsp/completion/providers/rulerefs/rulerefs.rego
```diff
@@ -120,23 +118,16 @@ grouped_refs[size] contains ref if {
...
 
 	item := {
 		"label": ref,
 		"kind": kind.variable,
 		"detail": "rule ref",
 		"textEdit": {
-			"range": {
-				"start": {
-					"line": position.line,
-					"character": position.character - count(last_word),
```
anderseknert, line 134: Here's where the result could previously be a negative number.

bundle/regal/lsp/completion/providers/rulerefs/rulerefs_test.rego
```diff
@@ -98,11 +100,11 @@ another_local_rule := imp`])
 		},
 		"context": {"location": {
 			"row": 10,
-			"col": 21,
+			"col": 26,
```
anderseknert, line 103: Not sure how this worked before 😅 I've used `|` to mark `col` 21 and 26 respectively below, and col 21 is in the middle of `:=` 🤔 

```
another_local_rule :|= imp|
```

Either way, after the change to use `location.ref_at` that function found no word at the given location, and would thus determine the "word" to be `""` — which typically means the completion was manually invoked and that all possible completions should be returned.

bundle/regal/lsp/completion/main.rego
```diff
@@ -10,4 +10,9 @@ import rego.v1
 # METADATA
 # description: main entry point for completion suggestions
 # entrypoint: true
-items contains data.regal.lsp.completion.providers[_].items[_]
+items contains item if {
+	some provider
+	completion := data.regal.lsp.completion.providers[provider].items[_]
+
+	item := object.union(completion, {"_regal": {"provider": provider}})
```
anderseknert, line 17: Only the `rulerefs` provider used to return this before, but knowing which provider is responsible for a suggestion is really helpful for debugging, so I moved this to the main router to get it for all providers.

charlieegan3 (approved):
👏 great to get this one sorted.

PR #906, opened 2024-07-09
lsp/hover: Show example links

charlieegan3:
Where examples to https://docs.styra.com/opa/rego-by-example are available, then show the links to these pages.

https://github.com/StyraInc/regal/assets/1774239/85fc6034-e7db-4b99-8d5a-e85a4f0e48de

### DIFF
--- a/bundle/regal/ast/keywords.rego
+++ b/bundle/regal/ast/keywords.rego
@@ -73,8 +73,6 @@ keywords[value.row] contains keyword if {
 
 	walk(expr.terms, [path, value])
 
-	# regal.last(path) == "location"
-
 	value.col
 	value.row
 
--- a/bundle/regal/ast/keywords.rego
+++ b/bundle/regal/ast/keywords.rego
@@ -64,8 +64,7 @@ keywords[loc.row] contains keyword if {
 }
 
 keywords[value.row] contains keyword if {
-	some rule in input.rules
-	some expr in rule.body
+	some expr in exprs[_]
 
 	walk(expr.terms, [path, value])
 
--- a/bundle/regal/ast/keywords.rego
+++ b/bundle/regal/ast/keywords.rego
@@ -1,0 +1,124 @@
+package regal.ast
+
+import rego.v1
+
+keywords[row] contains keyword if {
+	some rule in input.rules
+
+	rule_text := base64.decode(rule.location.text)
+
+	some idx, line in split(rule_text, "\n")
+
+	col := indexof(line, " if ")
+	col > 0
+
+	row := rule.location.row + idx
+
+	every comment in object.get(input, "comments", []) {
+		comment.Location.row != row
+	}
+
+	keyword := {
+		"name": "if",
+		"location": {
+			"row": rule.location.row + idx,
+			"col": col + 2,
+		},
+	}
+}
+
+keywords[pkg.location.row] contains keyword if {
+	pkg := input["package"]
+
+	keyword := {
+		"name": "package",
+		"location": {
+			"row": pkg.location.row,
+			"col": pkg.location.col,
+		},
+	}
+}
+
+keywords[imp.location.row] contains keyword if {
+	some imp in input.imports
+
+	keyword := {
+		"name": "import",
+		"location": {
+			"row": imp.location.row,
+			"col": imp.location.col,
+		},
+	}
+}
+
+keywords[loc.row] contains keyword if {
+	some rule in input.rules
+
+	loc := rule.head.location
+
+	col := indexof(base64.decode(loc.text), " contains ")
+
+	col > 0
+
+	keyword := {
+		"name": "contains",
+		"location": {
+			"row": loc.row,
+			"col": col + 2,
+		},
+	}
+}
+
+keywords[value.row] contains keyword if {
+	walk(input, [path, value])
+
+	regal.last(path) == "location"
+
+	name := {
+		"aW4=": "in",
+		"c29tZQ==": "some",
+		"ZXZlcnk=": "every",
+	}[value.text]
+
+	parent_path := array.slice(path, 0, count(path) - 1)
+	context := object.get(input, parent_path, {})
+
+	some keyword in _determine_keywords(context, value, name)
+}
+
+_determine_keywords(_, value, name) := {keyword} if {
+	name in {"in", "some"}
+
+	keyword := {
+		"name": name,
+		"location": {
+			"row": value.row,
+			"col": value.col,
+		},
+	}
+}
+
+_determine_keywords(context, value, name) := keywords if {
+	name == "every"
+
+	v := object.get(context, "value", {})
+
+	text := base64.decode(v.location.text)
+
+	keywords := {
+		{
+			"name": name,
+			"location": {
+				"row": value.row,
+				"col": value.col,
+			},
+		},
+		{
+			"name": "in",
+			"location": {
+				"row": value.row,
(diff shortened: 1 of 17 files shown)


internal/lsp/examples/examples.go
```diff
@@ -0,0 +1,49 @@
...
+type indexData struct {
+	BuiltIns map[string]string `json:"builtins"`
+	Keywords map[string]string `json:"keywords"`
+}
+
+const baseURL = "https://docs.styra.com/opa/rego-by-example"
+
+var index *indexData
+
+func init() {
```
charlieegan3, line 43: This takes `~63.709µs` on my machine.

internal/lsp/examples/index.json
```diff
@@ -0,0 +1,17 @@
+{
```
charlieegan3, line 1: This file exists in JSON so that it can be periodically updated based on the sitemap. I have code to generate this JSON which will come later.

```
curl -L https://docs.styra.com/sitemap.xml | xq . - > output.json

cat output.json | opa eval 'data.play.symbols' -d play.rego --format=raw --stdin-input | jq
```


play.rego

```rego
package play

import rego.v1

symbols := {"keywords": _keywords, "builtins": _builtins}

_keywords[name] := path if {
	some p in _pages
	p[0] == "keywords"

	name := p[1]

	path := concat("/", p)
}

_builtins[name] := path if {
	some p in _pages
	p[0] == "builtins"

	l := count(p)

	l == 2

	name := p[1]

	not count({p |
		some p in _pages
		p[0] == "builtins"
		p[1] == name
	}) > 1

	path := concat("/", p)
}

_builtins[name] := path if {
	some p in _pages
	p[0] == "builtins"

	l := count(p)

	l > 2

	name := concat(
		".",
		[
			replace(p[1], "_", "."),
			concat(".", array.slice(p, 2, l)),
		],
	)

	path := concat("/", p)
}

_prefix := "https://docs.styra.com/opa/rego-by-example/"

_pages contains page if {
	some i
	url := input.urlset.url[i].loc

	startswith(url, _prefix)

	page := split(trim_prefix(url, _prefix), "/")
}

```

internal/lsp/rego/rego.go
```diff
@@ -78,6 +92,46 @@ func AllBuiltinCalls(module *ast.Module) []BuiltInCall {
...
 
+//nolint:gochecknoglobals
+var regalRules = func() bundle.Bundle {
+	regalRules := rio.MustLoadRegalBundleFS(rbundle.Bundle)
+
+	return regalRules
+}()
+
+// AllKeywords returns all keywords in the module.
+func AllKeywords(ctx context.Context, module *ast.Module) (map[string][]KeywordUse, error) {
```
charlieegan3, line 103: This is currently pretty slow for ast.rego so I'm working on it's performance.

charlieegan3: I have this down to around 400ms on the first run, and then <50ms on the later runs. This is still not ideal, but a lot better than where it started at ~3s

charlieegan3: c92431ecc20054f6d646fa84b3f697f021b8ff3f
contains the improvements

bundle/regal/ast/keywords.rego
```diff
@@ -0,0 +1,132 @@
+package regal.ast
+
+import rego.v1
+
+keywords[row] contains keyword if {
+	some rule in input.rules
+
+	rule_text := base64.decode(rule.location.text)
+
+	some idx, line in split(rule_text, "\n")
```
anderseknert, line 10: You've got the decoded text representation available in `input.regal.file.lines`... I think we could save some processing here (and in more places in this file) using that.

bundle/regal/ast/keywords.rego
```diff
@@ -0,0 +1,132 @@
...
+			"col": col + 2,
+		},
+	}
+}
+
+keywords[value.row] contains keyword if {
+	some rule in input.rules
+	some expr in rule.body
+
+	walk(expr.terms, [path, value])
```
anderseknert, line 74: Probably good to add an `ast.exprs` similar to `ast.vars` at this point... as a few rules do this kind of walk now, and more likely to be added. It'd be nice to be able to reuse the result of that

bundle/regal/ast/keywords.rego
```diff
@@ -0,0 +1,132 @@
...
+			"row": value.row,
+			"col": value.col,
+		},
+	}
+}
+
+_determine_keywords(context, value, name) := keywords if {
+	name == "every"
+
+	v := object.get(context, "value", {})
```
anderseknert, line 102: Could just be `v := context.value` here AFAICS

bundle/regal/ast/keywords_test.rego
```diff
@@ -0,0 +1,155 @@
+package regal.ast_test
```
anderseknert, line 1: Nice tests!

internal/lsp/examples/index.json
anderseknert: 👍 There's some Rego in `build` already that's meant to be run periodically, but nothing to enforce that. Might be a good time to look into when you add this later.

bundle/regal/ast/keywords.rego
```diff
@@ -0,0 +1,132 @@
...
+		"name": name,
+		"location": {
+			"row": value.row,
+			"col": value.col,
+		},
+	}
+}
+
+_determine_keywords(context, value, name) := keywords if {
+	name == "every"
```
anderseknert, line 100:
```rego
_determine_keywords(context, value, "every") := keywords if {
```

🙂

anderseknert (commented):
Good work here! It's late and I'm OOO, lol, but I left some comments to consider

bundle/regal/ast/keywords.rego
charlieegan3: Thanks! fixed in 727ea45395f4045efc6f08da01083405bf65c930

bundle/regal/ast/keywords.rego
charlieegan3: 👍 I have fixed in 727ea45395f4045efc6f08da01083405bf65c930

bundle/regal/ast/keywords.rego
charlieegan3: This took me a while to test, but I have updated to do this in 73370ea after finding that it was in fact a little faster. The commit message has the testing details if interested.

internal/lsp/examples/index.json
charlieegan3: I was planning to add this automation in later, but shared the script here for reference when that time comes. since the content is quite static.

bundle/regal/ast/keywords.rego
charlieegan3: Hmm, I think I've managed to realise this in d8bd38bde5515c3088ce35a2be23d67c496e8969, but let me know if that's not what you had in mind. I had some trickiness with the rule indexing but it all works now. We just need to be careful to use the same meaning for rule index everywhere and it works!

charlieegan3:
I think I've addressed everything here, I'll aim to merge tomorrow.

charlieegan3:
3a599f147ba5f5fad88f41f696b4acaf05bf76d4 contains some fixes for head keywords.

charlieegan3:
I'm merging this in order to allow others to test all the new stuff ahead of a release next week.

Issue #654, opened 2024-04-16
Locations reported should include end position


anderseknert:
Regal currently inherits location reporting from OPA, which only provides the start location of the AST node. Regal should extend this in it's output reports to try and always include an end location as well, as that provides a better experience in UI's like editors. If this is difficult to calculate, we'll just settle for end of line. For most violations though, I think this should be fairly trivial. It's still a lot of work though, as we'll need to do this for all the rules in Regal, and update the report formats accordingly.

anderseknert:
Here's a good example of a rule that would greatly benefit from this:

![Screenshot 2024-05-15 at 13 35 52](https://github.com/StyraInc/regal/assets/510711/6110557a-bd0f-4b20-b9da-688ba421d5fc)

anderseknert:
This has been fixed with the use of [roast](https://github.com/StyraInc/roast). Might be there's a few rules still that doesn't correctly scope its locations, but we can fix those on a case by case basis.

PR #917, opened 2024-07-15
Begin to Include end location where applicable

anderseknert:
This PR attempts to establish end location as an optional attribute of location data returned in violations, and updates ~10 rules in the bugs category to use this convention.

If no end location is provided, the default is as it used to be — end location will be set to the end of the line. For some rules, this is the right thing to do anyway, and therefore, not all rules will need to be updated..

That obviously doesn't "fix" #654, but it's a start, and as far as I've seen so far, works quite well. Importantly this change is non-breaking method for API users! Ideally, we'd put the current "row" and "col" into a "start" object too, but that *would* be a breaking change, and one that would impact all API users, so let's consider that for our first stable release instead.

NOTE that this does not change anything in the output of `regal lint` using the default "pretty"-reporter, but is only included in the more detailed JSON and SARIF formats. First and foremost this change helps LSP clients display diagnostics more accurately, and that is really where this is most useful.

When this has been merged, I'll create some new issues to get this backported to each of the rule categories.

### DIFF
--- a/bundle/regal/result.rego
+++ b/bundle/regal/result.rego
@@ -56,7 +56,15 @@ _category_title_from_path(path) := [category, title] if ["regal", "rules", categ
 
 _category_title_from_path(path) := [category, title] if ["custom", "regal", "rules", category, title] = path
 
-# Provided rules, i.e. regal.rules.category.title
+# METADATA
+# description: |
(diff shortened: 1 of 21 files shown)


charlieegan3 (approved):
Looks good to me, thanks for working on this while you're out. I'll get it in, do some testing and start on the release.

PR #922, opened 2024-07-17
Rule: `unused-output-variable`

### LINKED ISSUE #60: Unused output variable
```rego
foo {
	some x
	input[x]
}
```
Even in strict mode, OPA considers `x` to be "used", as it's referenced _somewhere_. Howver, unless `x` is referenced later in the rule (or in the rule head) it's not really _used_, and the above could be replaced by:

```rego
foo {
	input[_]
}
```
Or 
```rego
foo {
	some _ in input
}
```

anderseknert:
One of the rules I've wanted since the project was started, but not until now would this have been wildly expensive to add. With the recent fixes that allow us to reuse the result of `walk`ing the AST, that's no longer the case. Still, this required quite some time and effort to get right!

NOTE: this rule currently only considers output variables as found in references, i.e. `x` in  `input[x]` and NOT `x` in `some x, _ in input`. While that would be easy to add, I'm not sure if those are really "output variables" or should be considered as such... 🤔 or if we should just have another rules for "unused iteration variable" or something like that. Either way, this is a big step forward!

Fixes #60

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,7 @@ The following rules are currently available:
 | bugs        | [rule-shadows-builtin](https://docs.styra.com/regal/rules/bugs/rule-shadows-builtin)                  | Rule name shadows built-in                                |
 | bugs        | [top-level-iteration](https://docs.styra.com/regal/rules/bugs/top-level-iteration)                    | Iteration in top-level assignment                         |
 | bugs        | [unassigned-return-value](https://docs.styra.com/regal/rules/bugs/unassigned-return-value)            | Non-boolean return value unassigned                       |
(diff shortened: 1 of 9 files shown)


charlieegan3 (approved):
Looks great to me, nice coverage in the tests.

anderseknert:
@srenatus I’d love to hear your thoughts on my note above! Am I right in my hunch that “x” in “some x in y” isn’t an “output variable” in the unification sense?

anderseknert:
blank-meme-template-075-Alright-then-keep-your-secrets-55314551

PR #929, opened 2024-07-21
add JUnit output format

### LINKED ISSUE #928: support JUnit XML output format
It would be nice to allow outputting linting results as JUnit XML because that can be parsed directly by GitLab and shown in merge requests.

sebhoss:
fixes #928 

The output groups violations by files in a testsuite and expresses each violation inside a file as a testcase. The files are sorted so that we have stable output during tests and each failure message contains the documentation link for a failed rule.

I'm not entirely sure whether notices should be part of the output as well. I've skipped them because most other reporters do not output them either.

### DIFF
--- a/README.md
+++ b/README.md
@@ -599,12 +599,13 @@ are:
   from the linter report
 - `sarif` - [SARIF](https://sarifweb.azurewebsites.net/) JSON output, for consumption by tools processing code analysis
   reports
+- `junit` - JUnit XML output, e.g. for CI servers like GitLab that show these results in a merge request.
 
 ## OPA Check and Strict Mode
 
 Linting with Regal assumes syntactically correct Rego. If there are errors parsing any files during linting, the
 process is aborted and any parser errors are logged similarly to OPA. OPA itself provides a "linter" of sorts,
-via the `opa check` comand and its `--strict` flag. This checks the provided Rego files not only for syntax errors,
+via the `opa check` command and its `--strict` flag. This checks the provided Rego files not only for syntax errors,
 but also for OPA [strict mode](https://www.openpolicyagent.org/docs/latest/policy-language/#strict-mode) violations.
 
 > **Note** It is recommended to run `opa check --strict` as part of your policy build process, and address any violations
--- a/cmd/lint.go
+++ b/cmd/lint.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -12,6 +13,7 @@ import (
 	"time"
 
 	"github.com/fatih/color"
+	"github.com/jstemmer/go-junit-report/v2/junit"
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v3"
 
@@ -372,6 +374,8 @@ func getReporter(format string, outputWriter io.Writer) (reporter.Reporter, erro
 		return reporter.NewFestiveReporter(outputWriter), nil
 	case formatSarif:
 		return reporter.NewSarifReporter(outputWriter), nil
+	case formatJunit:
+		return reporter.NewJUnitReporter(outputWriter), nil
 	default:
 		return nil, fmt.Errorf("unknown format %s", format)
 	}
@@ -406,6 +410,29 @@ func formatError(format string, err error) error {
 		}
 
 		return fmt.Errorf("%s", string(bs))
+	} else if format == formatJunit {
+		testSuites := junit.Testsuites{
+			Name: "regal",
+		}
+		testsuite := junit.Testsuite{
+			Name: "lint",
+		}
+		testsuite.AddTestcase(junit.Testcase{
+			Name: "Command execution failed",
+			Error: &junit.Result{
+				Message: err.Error(),
+			},
+		})
+		testSuites.AddSuite(testsuite)
+
+		buf := &bytes.Buffer{}
+
+		err := testSuites.WriteXML(buf)
+		if err != nil {
+			return fmt.Errorf("failed to format errors for output: %w", err)
+		}
+
+		return fmt.Errorf("%s", buf.String())
 	}
 
 	return err
--- a/pkg/reporter/reporter.go
+++ b/pkg/reporter/reporter.go
@@ -7,9 +7,11 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"sort"
 	"strings"
 
 	"github.com/fatih/color"
+	"github.com/jstemmer/go-junit-report/v2/junit"
 	"github.com/olekukonko/tablewriter"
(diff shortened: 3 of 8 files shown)


pkg/reporter/reporter.go
```diff
@@ -423,3 +436,41 @@ func getUniqueViolationURLs(violations []report.Violation) map[string]string {
...
+		testsuite := junit.Testsuite{
+			Name: file,
+		}
+
+		for _, violation := range violationsPerFile[file] {
+			testsuite.AddTestcase(junit.Testcase{
+				Name:      violation.Title,
+				Classname: violation.Location.String(),
+				Failure: &junit.Result{
+					Message: fmt.Sprintf("%s. To learn more, see: %s", violation.Description, getDocumentationURL(violation)),
```
anderseknert, line 467: Most notable omission here would be the category... and understandably so given that it's a test result format after all. If there isn't some clever meta attribute we can use for that, perhaps we could format the `Name` like `$category/$title: $message`, or perhaps have the `Message` include it?

Knowing the category helps with prioritization of fixing the issues reported, so it'd be good to have it in there _somewhere_, but again, if there's some other place to put it, I'm open to suggestions.

anderseknert (commented):
Amazing work on such a short notice! Just a suggestion to consider / discuss, but once that's resolved we'll have this merged. Thanks!

cmd/lint.go
```diff
@@ -396,8 +398,8 @@ func getWriterForOutputFile(filename string) (io.Writer, error) {
 }
 
 func formatError(format string, err error) error {
-	// currently, JSON and SARIF will get the same generic JSON error format
-	if format == formatJSON || format == formatSarif {
+	// currently, JSON, SARIF, and JUnit will get the same generic JSON error format
+	if format == formatJSON || format == formatSarif || format == formatJunit {
```
anderseknert, line 402: I would probably have expected the errors to be reported as "JUnit errors"... but I guess we'd need to update the reporters to allow for that. We'll deal with that later.

pkg/reporter/reporter.go
sebhoss: oh yeah good idea! What would `$message` be though? Maybe just the description of a violation and have the `Message` contain both description and link to the docs?

I think we can add any kind of additional metadata to the `Data` field of a `junit.Result`. It's currently not used in the code but we could write all the violation details in there. I'll try it out on our instance and post screenshots to further the discussion

cmd/lint.go
sebhoss: OK thanks for pinging me about this - I wasn't really sure why this code exists. In my understanding, this function is only called if the lint command itself failed. As a user I would not expect to get any kind of formatted report for that but rather a harsh error message telling me to fix my setup before regal can work properly.

That said, I could add proper JUnit XML output here as well. Besides `Failure` it does contain an `Error` field that could be used here..

pkg/reporter/reporter.go
sebhoss: This shows the test report widget inside a merge request:

2024-07-22T10:15:42,541965099+02:00

that's the popup you get when clicking on `View details`:

2024-07-22T10:15:59,187202303+02:00

and this is what the `Full report` looks like

2024-07-22T10:16:15,244480379+02:00

when you click on `View details` in the full report, you get the same popup as in the merge request

anderseknert: Looks great! Where does the category come from there? You've updated your code to show that? I don't see it in the PR, but that looks great to me, so please do include it!

cmd/lint.go
anderseknert: Yeah, but the lint command can fail for reasons like parser errors too. Ideally, other tools like `opa test` would have failed the build before linting in that case, but still. In most of the contexts where e.g. SARIF is used, like in CI/CD pipelines, users would likely want to find the report in the same place regardless of what made it fail, but yeah, there's a reason we've not exactly made this our highest priority too.

anderseknert:
Sorry, I didn't think of this before, but can yoy please update [this section](https://github.com/styrainc/regal?tab=readme-ov-file#output-formats) in the README too?

And I suppose [this guide](https://github.com/StyraInc/regal/blob/main/docs/cicd.md) could use an update too so that GitLab users can benefit from this.

pkg/reporter/reporter.go
sebhoss: er yeah I pushed it to another branch so not to pollute this PR. I also just realized that GitLab no longer shows the failure message which included the link to the documentation. So the latest version now includes the documentation link and its output is aligned with the pretty printer. It looks like this:

2024-07-22T12:50:15,296173773+02:00

It's kinda sad that GitLab formats this output as a code block so links are not clickable but it think there is no way around that

cmd/lint.go
sebhoss: OK OK sounds good - will add JUnit output here as well!

anderseknert: If non-trivial to pull off, it's alright to defer this to later. We should look into better error reporting altogether sometime soon anyway.

pkg/reporter/reporter.go
anderseknert: Yeah, but not much to do about it I suppose. GitHub doesn't make it a link either, even if it's not in a codeblock. In that context we can at least display real markdown links in the summary though.

anderseknert: Looks great though!

README.md
```diff
@@ -599,12 +599,13 @@ are:
...
 - `sarif` - [SARIF](https://sarifweb.azurewebsites.net/) JSON output, for consumption by tools processing code analysis
   reports
+- `junit` - JUnit XML output, e.g. for CI servers like GitLab that show these results in a merge request.
 
 ## OPA Check and Strict Mode
 
 Linting with Regal assumes syntactically correct Rego. If there are errors parsing any files during linting, the
 process is aborted and any parser errors are logged similarly to OPA. OPA itself provides a "linter" of sorts,
-via the `opa check` comand and its `--strict` flag. This checks the provided Rego files not only for syntax errors,
+via the `opa check` command and its `--strict` flag. This checks the provided Rego files not only for syntax errors,
```
anderseknert, line 608: Thanks :)

anderseknert (approved):
Thanks a lot!

@charlieegan3 take a look if you want, or feel free to merge directly.

anderseknert:
Apologies for the flaky tests, @sebhoss. We'll [fix that](https://github.com/StyraInc/regal/issues/931) this week, but do know it's not from your change.

cmd/lint.go
sebhoss: I think the latest version is a reasonable start - it basically just creates a single testsuite with a single testcase that contains the given error string.

sebhoss:
OK I rebased on latest main, fixed all linting errors, and squashed all commits. Good to go from my side ^^

Issue #936, opened 2024-07-25
Truncate `Text` field value in default pretty reporter when needed


anderseknert:
When violations occur on some _very_ long lines, the pretty reporter prints it in its entirety under the `Text` attribute. This can cause the terminal output to be wrapped, which makes everything look bad. The following screenshot demonstrates this:

<img width="676" alt="Screenshot 2024-07-25 at 08 12 24" src="https://github.com/user-attachments/assets/51228fac-2142-4a68-8a84-0a93b2ac42b6">

This is the rule that caused this. The wrapping however taints all violations, and not just the one that contained a long line.

We should truncate the output of the `Text` field to max 120 characters, as that's the default max line length recommendation.

```
Rule:         	rule-name-repeats-package                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               	
Description:  	Rule name repeats package                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               	
Category:     	style                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   	
Location:     	cloud-foundry/polices/pas/cert/certificate_test.rego:90:1                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               	
[...]

PR #937, opened 2024-07-25
reporter: Trim long lines of location text

charlieegan3:
Fixes https://github.com/StyraInc/regal/issues/936

### DIFF
--- a/pkg/reporter/reporter.go
+++ b/pkg/reporter/reporter.go
@@ -204,8 +204,8 @@ func buildPrettyViolationsTable(violations []report.Violation) string {
 		table.Append([]string{yellow("Location:"), cyan(violation.Location.String())})
 
 		if violation.Location.Text != nil {
-			if len(*violation.Location.Text) > 77 {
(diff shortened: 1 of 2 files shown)


pkg/reporter/reporter.go
```diff
@@ -204,7 +204,11 @@ func buildPrettyViolationsTable(violations []report.Violation) string {
 		table.Append([]string{yellow("Location:"), cyan(violation.Location.String())})
 
 		if violation.Location.Text != nil {
-			table.Append([]string{yellow("Text:"), strings.TrimSpace(*violation.Location.Text)})
+			if len(*violation.Location.Text) > 77 {
```
anderseknert, line 207: I suggested 120 characters in the issue. Was there some specific reason you went with 80? Totally fine with me, just curious 🙂

anderseknert (approved):
Thanks a lot, Charlie! 😃

pkg/reporter/reporter.go
charlieegan3: Ahh, I can change it. I got distracted by trying to use `tput cols` and then opted for a static value. I'll make it 120.

PR #947, opened 2024-07-29
Include function args in `ast.vars` (and `prefer-snake-case` rule)

### LINKED ISSUE #942: Function arguments not considered by `prefer-snake-case`
For a policy like this:

```rego
package p

import rego.v1

fun(fooBar) := fooBar + 2
```

I expect `fooBar` to be flagged by the [prefer-snake-case](https://docs.styra.com/regal/rules/style/prefer-snake-case) rule. But it seems like it isn't.

anderseknert:
Fixes #942

### DIFF
--- a/bundle/regal/rules/style/prefer_snake_case_test.rego
+++ b/bundle/regal/rules/style/prefer_snake_case_test.rego
@@ -8,7 +8,13 @@ import data.regal.rules.style["prefer-snake-case"] as rule
 
 test_fail_camel_cased_rule_name if {
 	r := rule.report with input as ast.policy(`camelCase := 5`)
-	r == expected_with_locations([{"col": 1, "file": "policy.rego", "row": 3, "text": `camelCase := 5`}])
+	r == expected_with_locations([{
+		"col": 1,
+		"file": "policy.rego",
+		"row": 3,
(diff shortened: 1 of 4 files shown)


bundle/regal/util/util.rego
```diff
@@ -25,8 +25,8 @@ find_duplicates(arr) := {indices |
 
 # METADATA
 # description: returns true if array has duplicates of item
-has_duplicates(array, item) if count([x |
-	some x in array
+has_duplicates(arr, item) if count([x |
```
anderseknert, line 28: Nice side effect of args counting as vars is how this now got flagged as shadowing the built-in `array` namespace.

bundle/regal/rules/style/prefer_snake_case_test.rego
```diff
@@ -43,9 +56,26 @@ test_success_snake_cased_multiple_some_declaration if {
 	r == set()
 }
 
+test_fail_camel_cased_function_argument if {
```
charlieegan3, line 59: Do we want a `_function_argument` success case for completeness?

anderseknert: Sure, I'll add one before merging 👍

PR #948, opened 2024-07-29
Rule: `prefer-strings-count`

### LINKED ISSUE #940: Rule: `prefer-strings-count`
Simple rule! 

From OPA v0.67.0 and onwards (this should be checked for in capabilities), suggest using `strings.count` over `count(indexof_n(..))`. Easier to read and more performant.

anderseknert:
Fixes #940

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,6 +217,7 @@ The following rules are currently available:
 | idiomatic   | [use-if](https://docs.styra.com/regal/rules/idiomatic/use-if)                                         | Use the `if` keyword                                      |
 | idiomatic   | [use-in-operator](https://docs.styra.com/regal/rules/idiomatic/use-in-operator)                       | Use in to check for membership                            |
 | idiomatic   | [use-some-for-output-vars](https://docs.styra.com/regal/rules/idiomatic/use-some-for-output-vars)     | Use `some` to declare output variables                    |
+| idiomatic   | [use-strings-count](https://docs.styra.com/regal/rules/idiomatic/use-strings-count)                   | Use `strings.count` where possible                        |
 | imports     | [avoid-importing-input](https://docs.styra.com/regal/rules/imports/avoid-importing-input)             | Avoid importing input                                     |
 | imports     | [circular-import](https://docs.styra.com/regal/rules/imports/circular-import)                         | Circular import                                           |
 | imports     | [ignored-import](https://docs.styra.com/regal/rules/imports/ignored-import)                           | Reference ignores import                                  |
--- a/bundle/regal/capabilities.rego
+++ b/bundle/regal/capabilities.rego
@@ -16,6 +16,8 @@ provided := data.internal.capabilities
 
 has_object_keys if "object.keys" in object.keys(config.capabilities.builtins)
 
+has_strings_count if "strings.count" in object.keys(config.capabilities.builtins)
+
 # if if if!
 has_if if "if" in config.capabilities.future_keywords
 
--- a/bundle/regal/config/provided/data.yaml
+++ b/bundle/regal/config/provided/data.yaml
@@ -6,8 +6,8 @@ rules:
     annotation-without-metadata:
       level: error
     argument-always-wildcard:
+      except-function-name-pattern: '^mock_'
       level: error
-      except-function-name-pattern: "^mock_"
     constant-condition:
       level: error
     deprecated-builtin:
@@ -82,6 +82,8 @@ rules:
     use-in-operator:
       level: error
     use-some-for-output-vars:
+      level: error
+    use-strings-count:
       level: error
   imports:
     avoid-importing-input:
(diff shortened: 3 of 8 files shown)


docs/rules/idiomatic/prefer-strings-count.md
```diff
@@ -0,0 +1,50 @@
+# prefer-strings-count
+
+**Summary**: Prefer to use `strings.count`
```
charlieegan3, line 3: question: Is there a difference between our 'use' and 'prefer' rules?

```
docs/rules/custom/prefer-value-in-head.md
3:**Summary**: Prefer value in rule head

docs/rules/imports/prefer-package-imports.md
3:**Summary**: Prefer importing packages over rules

docs/rules/idiomatic/prefer-set-or-object-rule.md
3:**Summary**: Prefer set or object rule over comprehension

docs/rules/idiomatic/boolean-assignment.md
3:**Summary**: Prefer `if` over boolean assignment

docs/rules/idiomatic/equals-pattern-matching.md
3:**Summary**: Prefer pattern matching in function arguments

docs/rules/style/prefer-some-in-iteration.md
3:**Summary**: Prefer `some .. in` for iteration

docs/rules/style/default-over-not.md
3:**Summary**: Prefer default assignment over negated condition

docs/rules/style/prefer-snake-case.md
3:**Summary**: Prefer snake_case for names

docs/rules/style/default-over-else.md
3:**Summary**: Prefer default assignment over fallback `else`

docs/rules/style/use-assignment-operator.md
3:**Summary**: Prefer `:=` over `=` for assignment
```

```
3:**Summary**: Use raw strings for regex patterns

docs/rules/idiomatic/use-if.md
3:**Summary**: Use the `if` keyword

docs/rules/idiomatic/use-some-for-output-vars.md
3:**Summary**: Use `some` to declare output variables

docs/rules/idiomatic/use-in-operator.md
3:**Summary**: Use `in` to check for membership

docs/rules/idiomatic/use-contains.md
3:**Summary**: Use the `contains` keyword

docs/rules/imports/use-rego-v1.md
3:**Summary**: Use `import rego.v1`

docs/rules/bugs/not-equals-in-loop.md
3:**Summary**: Use of != in loop
```

Still working out how I feel, but I think 'Use' is more opinionated and clear than 'Prefer'.

I don't really want to update existing rules, but I think this new rule might be a 'Use' rule since there is no preference about it, strings.Count is better right?

docs/rules/idiomatic/prefer-strings-count.md
```diff
@@ -0,0 +1,50 @@
...
+package policy
+
+import rego.v1
+
+num_as := strings.count("foobarbaz", "a")
+```
+
+## Rationale
+
+The `strings.count` function added in OPA v0.67.0 is both more readable and efficient compared to using
```
charlieegan3, line 27:
```suggestion
The `strings.count` function added in [OPA v0.67.0](https://github.com/open-policy-agent/opa/releases/tag/v0.67.0) is both more readable and efficient compared to using
```

bundle/regal/rules/idiomatic/prefer_strings_count.rego
```diff
@@ -0,0 +1,35 @@
...
+import data.regal.result
+
+# METADATA
+# description: Missing capability for built-in function `strings.count`
+# custom:
+#   severity: warning
+notices contains result.notice(rego.metadata.chain()) if not capabilities.has_object_keys
+
+# METADATA
+# description: flag calls to `count`` where the first argument is a call to `indexof_n`
```
charlieegan3, line 18:
```suggestion
# description: flag calls to `count` where the first argument is a call to `indexof_n`
```

bundle/regal/rules/idiomatic/prefer_strings_count.rego
```diff
@@ -0,0 +1,35 @@
...
+report contains violation if {
```
charlieegan3, line 19: Is it too much effort to catch cases like this?

```
indexes := indexof_n(foo)
count := count(indexes) 
```

charlieegan3: I suppose then we'd need to check if indexes is used elsewhere too...

charlieegan3 (approved):
Looks good to me, just some comments on 'prefer' vs 'use'

bundle/regal/rules/idiomatic/prefer_strings_count.rego
anderseknert: Yeah, that's too much effort IMHO :D I'd be surprised if this was a violation in more than 1% of the policy repos out there already... but if we find it being "popular" later, we can work on improving it.

docs/rules/idiomatic/prefer-strings-count.md
charlieegan3: 👍 sounds good to me!

Issue #879, opened 2024-07-01
Editing an ignored file shows completion errors


anderseknert:
Despite the file being ignored, an error is displayed for each time a key is prressed:
```
[Error - 10:18:53 AM] Request textDocument/completion failed.
  Message: failed to find completions: error running completion provider: could not get file contents for: file:///Users/anderseknert/git/styra/regal/e2e/testdata/violations/most_violations.rego
  Code: 0 
```

PR #953, opened 2024-07-30
lsp: Skip textDocument/completion for ignored files

charlieegan3:
When files are ignored, they will have no file contents in the cache, and so completions are impossible. This leads to an error message which is now avoided by just returning an empty list.

Fixes https://github.com/StyraInc/regal/issues/879


anderseknert (approved):
Spasiba!

PR #956, opened 2024-08-01
lsp: Auto update examples index

charlieegan3:
As pages are added to docs.styra.com/opa/rego-by-example, this json file will be updated meaning more keywords and builtins will be linked in the Regal hover messages.


internal/lsp/examples/index.json
```diff
@@ -1,17 +1,17 @@
...
-        "time.clock": "builtins/time/clock",
-        "time.format": "builtins/time/format",
-        "time.now_ns": "builtins/time/now_ns",
-        "time.parse_ns": "builtins/time/parse_ns"
-    },
-    "keywords": {
-        "contains": "keywords/contains",
-        "some": "keywords/some"
-    }
+  "builtins": {
```
charlieegan3, line 2: Changes here will be updated when the workflow is manually run on main, or on schedule.

build/workflows/update-example-index/process.rego
```diff
@@ -0,0 +1,62 @@
...
+	some p in _pages
+	p[0] == "builtins"
+
+	l := count(p)
+
+	l == 2
+
+	name := p[1]
+
+	not count({p |
```
anderseknert, line 26: `not count` looks fishy... even if it's the `>` being negated. `< 2` instead?

anderseknert (commented):
Just a few minor remarks, but looks good!

.github/workflows/update-example-index.yaml
```diff
@@ -0,0 +1,51 @@
...
+    runs-on: ubuntu-22.04
+    permissions:
+      contents: write # this workflow updates repo contents
+    steps:
+      - name: Check out code
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+
+      - name: Install xq (part of yq)
+        run: |
+          pip install yq
```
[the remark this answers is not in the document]
anderseknert:
```
➜ cat p.xml | rq -i xml --indent "  "
{
  "root": {
    "p": {
      "a": "1",
      "b": "2",
      "c": "3"
    }
  }
}
```

build/workflows/update-example-index/process.rego
charlieegan3: Hmm, that doesn't look good... I have fixed to your version now.

.github/workflows/update-example-index.yaml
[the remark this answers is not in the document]
charlieegan3: Interesting! Have fixed this now to use rq and the cached binary too.

charlieegan3:
Think all addressed here now so going to merge for a final test on main.

PR #957, opened 2024-08-01
fix: Only run fix linter with fixable rules

charlieegan3:
This doesn't appear to improve the performance sadly in my testing on the regal bundle and https://github.com/kubescape/regolibrary. However, I feel that this makes sense as a minor tidy up here.


charlieegan3:
Oops, realise we do this already...

PR #958, opened 2024-08-01
lsp/format,cmd/fix: Apply rego.v1 formatting by default

charlieegan3:
Summary of changes:

- when running the fix cmd, the rego.v1 formatter will be run for files regardless of `rego.v1` being imported.
- when formatting on save in a language server client, the rego.v1 formatter will run, and all fixable violations will also be corrected at save time.

This is implemented by adding the concept of mandatory fixes, where these are run on all files even if there are no violations, before the linter based fixes are run.

LSP on save:

https://github.com/user-attachments/assets/7be3d304-62ab-4b3a-8b71-2364b758033a

Fix cmd:
Screenshot 2024-08-01 at 18 04 54

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -38,6 +38,8 @@ import (
 	"github.com/styrainc/regal/internal/update"
 	"github.com/styrainc/regal/internal/util"
 	"github.com/styrainc/regal/pkg/config"
+	"github.com/styrainc/regal/pkg/fixer"
+	"github.com/styrainc/regal/pkg/fixer/fileprovider"
 	"github.com/styrainc/regal/pkg/fixer/fixes"
 	"github.com/styrainc/regal/pkg/linter"
 	"github.com/styrainc/regal/pkg/version"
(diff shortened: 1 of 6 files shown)


internal/lsp/server.go
```diff
@@ -1313,24 +1315,46 @@ func (l *LanguageServer) handleTextDocumentFormatting(
...
+	input, err := memfp.ToInput()
 	if err != nil {
-		l.logError(fmt.Errorf("failed to format file: %w", err))
+		return nil, fmt.Errorf("failed to create fixer input: %w", err)
+	}
 
-		// return "null" as per the spec
-		return nil, nil
+	li := linter.NewLinter().
+		WithUserConfig(*l.loadedConfig).
```
anderseknert, line 1329: This is a nil-dereference panic if no config is provided. I know because I did the same mistake on my branch 😆

anderseknert (approved):
There are a few things to change here, like the mentioned possible nil dereference. And I I think we should ensure that fixes are either mandatory or not, but not both. But I can do that in a new PR. Thanks a lot for helping out here!

PR #960, opened 2024-08-02
Allow using `regal fix` as a formatter

### LINKED ISSUE #839: Allow using `regal fix` as a formatter
It would be nice to be able to set the equivalent LSP command for `regal fix` as the formatter for Rego, which would make it possible to have "format on save" run `regal fix` instead of `opa fmt`.  Since `regal fix` is 100% compatible with `opa fmt`, this would simply be fmt with some nice additions, and more to come if we start looking into ordering of imports, and so on.

Looking into if it's possible to provide this via the LSP initialization options from the client would be a good first step.

anderseknert:
Also:
- Removed the code that would send a response back for formatter options that `opa fmt` can't handle. This was a good idea, but since there isn't any obvious way for users to fix that, it hurts more than it helps
- Removed the fmt fix from the default fixer options as formatting is done anyway by the rego-v1 fixer

Fixes #839

### DIFF
--- a/docs/fixing.md
+++ b/docs/fixing.md
@@ -22,3 +22,8 @@ Currently, the following rules are automatically fixable:
 Need to fix individual violations? Checkout the editors Regal supports
 [here](/regal/editor-support).
 :::
+
+## Community
+
+If you'd like to discuss Regal development or just talk about Regal in general, please join us in the `#regal`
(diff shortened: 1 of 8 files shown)


charlieegan3 (approved):
This all looks great to me, thanks for finishing this up!

PR #965, opened 2024-08-06
perf: Walk less

anderseknert:
- Use `ast.found.refs` instead of `walk` in `use-some-for-output-vars`
- Cheaper `walk` over `else` in `default-over-else` and `use-assignment-operator`
- Remove some unnecessary code in `external-reference`

The plan was also to consolidate `ast.found.symbols` with `ast.found.refs`, but that had disastrous consequences leading to something I think is a bug in OPA. I have demonstrated that to @charlieegan3, but will need to consult the OPA team once they're back from vacation.


charlieegan3 (approved):
Looks good to me, no requests!

PR #968, opened 2024-08-07
lsp: Add code lens support for evaluating rules

anderseknert:
- This is currently limited to VS Code clients, but may be extended to other clients too in the furure, if they can support it on their side.

- Implementation wil recursively look for an `input.json` file and pick the one closest to the file evaluated, down to the level of the workspace root (if any file is found).

### DIFF
--- a/internal/lsp/eval.go
+++ b/internal/lsp/eval.go
@@ -1,0 +1,113 @@
+package lsp
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+
+	"github.com/open-policy-agent/opa/ast"
+	"github.com/open-policy-agent/opa/bundle"
+	"github.com/open-policy-agent/opa/rego"
+	"github.com/open-policy-agent/opa/topdown"
+
+	"github.com/styrainc/regal/internal/lsp/uri"
+	"github.com/styrainc/regal/pkg/builtins"
+)
+
+func (l *LanguageServer) Eval(ctx context.Context, query string, input string) (rego.ResultSet, error) {
+	modules := l.cache.GetAllModules()
+	moduleFiles := make([]bundle.ModuleFile, 0, len(modules))
+
+	for fileURI, module := range modules {
+		moduleFiles = append(moduleFiles, bundle.ModuleFile{
+			URL:    fileURI,
+			Parsed: module,
+			Path:   uri.ToPath(l.clientIdentifier, fileURI),
+		})
+	}
+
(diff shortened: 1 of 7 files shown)


internal/lsp/server.go
```diff
@@ -447,6 +450,59 @@ func (l *LanguageServer) StartCommandWorker(ctx context.Context) {
...
+				}
+
+				responseParams := map[string]any{
+					"result": result,
+					"line":   line,
+				}
+
+				responseResult := map[string]any{}
+
+				err = l.conn.Call(ctx, "regal/showEvalResult", responseParams, &responseResult)
```
charlieegan3, line 502: We must use a custom event here as there is client code that will actually show this - rather than being in a generic message/location? Just checking I have that right.

internal/lsp/server.go
```diff
@@ -1610,6 +1739,13 @@ func (l *LanguageServer) handleInitialize(
 		},
 	}
 
+	// Since evaluation requires some client side handling, this can't be supported
+	// purely by the LSP. Clients that are capable of handling the code lens callback
+	// should be added here though.
+	if l.clientIdentifier == clients.IdentifierVSCode {
+		initializeResult.Capabilities.CodeLensProvider = &types.CodeLensOptions{}
```
charlieegan3, line 1746: Perhaps to come later, but something we can write to an result.json instead perhaps, but still show the command. (as discussed)

charlieegan3 (approved):

internal/lsp/server.go
anderseknert: Exactly! Client side code [over here](https://github.com/open-policy-agent/vscode-opa/compare/main...anderseknert:vscode-opa:lsp-eval).

internal/lsp/server.go
anderseknert: Yep, that'll come next!

internal/lsp/eval.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/open-policy-agent/opa/ast"
+	"github.com/open-policy-agent/opa/bundle"
+	"github.com/open-policy-agent/opa/rego"
+	"github.com/open-policy-agent/opa/topdown"
+
+	"github.com/styrainc/regal/internal/lsp/uri"
+	"github.com/styrainc/regal/pkg/builtins"
+)
+
+func (l *LanguageServer) Eval(ctx context.Context, query string, input string) (rego.ResultSet, error) {
```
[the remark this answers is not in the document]
anderseknert: No, just was the easy choice I guess, but I can look into changing that in the next PR 👍

PR #969, opened 2024-08-07
docs: Fix spelling mistakes

charlieegan3:
There were a number of spelling mistakes in the documentation. This PR fixes them as the spelling will soon be enforced in docs.styra.com.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -77,7 +77,7 @@ build/do.rq test
 
 ### E2E tests
 
-End-to-End (E2E) tests assert the behaviour of the `regal` binary called with certain configs, and test files.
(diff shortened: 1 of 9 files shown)


anderseknert (approved):
Thanks a lot! Would it not make sense to run the same spell checker in this repo? That way we'd be able to catch mistakes where they happen. But of course, not blocking.

charlieegan3:
Indeed, I would like to have a spell checker here too. I _think_ the way to go is to not check the regal spelling in docs.styra.com eventually and check here instead. I have a think about the best way to do it.

anderseknert:
Inspires me to think about spell checking as a Regal linter rule too :) Might be some good Go library for that we could turn into a custom function, and then have at least an optional rule that could spell check comments, metadata or even code if desired.

charlieegan3:
💯 checking spelling in comments would be great.

PR #972, opened 2024-08-07
LSP: Provide output.json option for non-VS Code clients

anderseknert:
And use a reader instead of a string for input as suggested by @charlieegan3

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,9 @@ dist/
 
 /regal
 /regal.exe
+
+# These two files are used by the Regal evaluation Code Lens, where input.json
+# defines the input to use for evaluation, and output.json is where the output
+# ends up unless the client supports presenting it in a different way.
+input.json
+output.json
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -480,10 +480,8 @@ func (l *LanguageServer) StartCommandWorker(ctx context.Context) {
 					break
 				}
 
-				input := FindInput(
-					uri.ToPath(l.clientIdentifier, file),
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -4,3 +4,6 @@ dist/
 
 /regal
 /regal.exe
+
+input.json
```
charlieegan3, line 8: Might be good to add a comment where these come from single it's not obvious?

internal/lsp/server.go
```diff
@@ -492,16 +490,43 @@ func (l *LanguageServer) StartCommandWorker(ctx context.Context) {
...
+
+					f, err = os.OpenFile(output, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, 0o755)
+
+					if err == nil {
+						var jsonVal []byte
+
+						value := result.Value
+						if result.IsUndefined {
+							// Display undefined as an empty object
+							// we could also go with "<undefined>" or similar
```
charlieegan3, line 518: I quite like the idea of calling it `undefined` rather than the empty object myself...

charlieegan3 (approved):
LGTM

internal/lsp/server.go
anderseknert: I'm not too sure, as it makes it hard to differentiate from the actual string value of "undefined"... 
Using "<undefined>" as  `print` does is a little better. But for now I went with empty object as that's what `opa eval` prints by default, e.g.
```
opa eval input
{}
```

.gitignore
anderseknert: Yeah, good call, I'll add that!

PR #978, opened 2024-08-09
lsp: include `print` output in eval response

anderseknert:

### DIFF
--- a/internal/lsp/eval.go
+++ b/internal/lsp/eval.go
@@ -14,13 +14,18 @@ import (
 	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/bundle"
 	"github.com/open-policy-agent/opa/rego"
-	"github.com/open-policy-agent/opa/topdown"
+	"github.com/open-policy-agent/opa/topdown/print"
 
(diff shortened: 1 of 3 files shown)


internal/lsp/eval.go
```diff
@@ -88,33 +94,49 @@ func FindInput(file string, workspacePath string) io.Reader {
...
+		return EvalPathResult{IsUndefined: true, PrintOutput: hook.Output}, nil
 	}
 
 	res, ok := result[0].Bindings["result"]
 	if !ok {
 		return EvalPathResult{}, errors.New("expected result in bindings, didn't get it")
 	}
 
-	return EvalPathResult{Value: res}, nil
+	return EvalPathResult{Value: res, PrintOutput: hook.Output}, nil
```
charlieegan3, line 120: This is what's sent to the client right?

So it's up to the client to work out things like:

```
package play

import rego.v1

allow if {print(true)}
```

charlieegan3 (approved):

internal/lsp/eval.go
anderseknert: Indeed!

PR #983, opened 2024-08-12
Document the Code Lens Evaluation feature

anderseknert:


charlieegan3 (approved):
This is great, I wonder if an animation would make this even more exciting as a page? rather than just an image? Just an idea.

anderseknert:
Agreed on the animation @charlieegan3 👍 Just rushing to get things in for the release. Let's add that after, and perhaps for some of the other features too,

PR #987, opened 2024-08-14
lsp: Implement bundle use in workspace Eval

charlieegan3:
This makes it possible to workspace eval rules based on data.

https://github.com/user-attachments/assets/0b79221b-21bc-4344-8873-a849222c9533

### DIFF
--- a/internal/lsp/bundles/cache.go
+++ b/internal/lsp/bundles/cache.go
@@ -1,0 +1,209 @@
+package bundles
+
+import (
+	"bytes"
+	//nolint:gosec
+	"crypto/md5"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/open-policy-agent/opa/bundle"
+)
+
+// Cache is a struct that maintains a number of bundles in memory and
+// provides a way to refresh them when the source files change.
+type Cache struct {
+	workspacePath string
+	bundles       map[string]*cacheBundle
+}
+
+func NewCache(workspacePath string) *Cache {
+	if !strings.HasSuffix(workspacePath, string(filepath.Separator)) {
+		workspacePath += string(filepath.Separator)
+	}
+
+	return &Cache{
+		workspacePath: workspacePath,
+		bundles:       make(map[string]*cacheBundle),
+	}
+}
+
+// Refresh walks the workspace path and loads or refreshes any bundles that
+// have changed since the last refresh.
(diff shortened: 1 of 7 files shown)


internal/lsp/bundles/cache.go
```diff
@@ -0,0 +1,238 @@
...
+	return refreshedBundles, nil
+}
+
+// List returns a list of all the bundle roots that are currently present in
+// the cache.
+func (c *Cache) List() []string {
+	roots := make([]string, 0, len(c.bundles))
+	for root := range c.bundles {
+		roots = append(roots, root)
+	}
```
anderseknert, line 129: Could use `util.Keys` :)

internal/lsp/bundles/cache.go
```diff
@@ -0,0 +1,238 @@
...
+	// if there has been any change in any of the source files, then
+	// reload the bundle
+	c.bundle, err = LoadDataBundle(path)
+	if err != nil {
+		return false, fmt.Errorf("failed to load bundle %q: %w", path, err)
+	}
+
+	// update the bundle's sourceDigests to the new on-disk state after a
+	// successful refresh
+	c.sourceDigests = onDiskSourceDigests
```
anderseknert, line 219: Nice!

internal/lsp/server.go
```diff
@@ -1887,6 +1896,17 @@ func (l *LanguageServer) loadWorkspaceContents(ctx context.Context, newOnly bool
...
 	}
 
+	if l.bundleCache != nil {
+		rb, err := l.bundleCache.Refresh()
+		if err != nil {
+			return nil, fmt.Errorf("failed to refresh the bundle cache: %w", err)
+		}
+
+		if len(rb) > 0 {
+			fmt.Fprintln(os.Stderr, "Reloaded bundles:", rb)
```
anderseknert, line 1906: Do we need to print this each time?

anderseknert (approved):
Wow, awesome work!

internal/lsp/bundles/cache.go
charlieegan3: Very nice, done in 9d2aa3bfad3159476acf2b8ba9ddf0e47cddde9a

internal/lsp/server.go
charlieegan3: I've dropped it in 9d2aa3bfad3159476acf2b8ba9ddf0e47cddde9a. It'd be nice to have some log level control but that can come another time.

PR #1000, opened 2024-08-20
Charles/eopa engine

charlesdaniels:
This PR adds initial support for an EOPA engine. It depends on the upstream PR [#92](https://github.com/StyraInc/enterprise-opa/pull/92), and should not be merged before that one is.

It includes 3 main components:

1. The `config` package has been changed so that if the `eopa` engine is used, the new helpers in the upstream PR are used to look up the capabilities that are attached to the `config.Config` type. The engine and version are saved in the config for later use.
2. The `internal/lsp` package has been modified to inspect the engine and version from the configuration, and to look up capabilities for use in the `internal/lsp/rego` package.
3. A new `capabilities` package has been added to provide a general-purpose way for Regal to access OPA capabilities. It unifies access to local, remote, and embedded capabilities using URLs. Both `config` and `internal/lsp` have been refactored to use this new system.

In summary, `capabilities.Lookup()` supports loading OPA capabilities in the following ways:
* `file://` URLs are loaded as JSON from disk, just like specifying `from.file`.
* `http{s}:// URLs are loaded as JSON from a web request, which was not previously supported but has been discussed.
* `regal:///capabilities/default` loads the capabilities using `ast.CapabilitiesForThisVersion()`
* `regal:///capabilities/{engine}` loads the latest (according to SemVer) capabilities version for the engine
* `regal:///capabilities/{engine}/{version}` loads the specified capabilities version for the engine

Semantic versioning support is implemented using `github.com/coreos/go-semver/semver`, which is the same library OPA uses for its semver builtins. 

The existing `from.engine`, `from.file`, and `from.version` will continue to work, but internally they are translated to URLs so that access to capabilities is consistent everywhere in Regal. `from.url` has been added as well. 

These changes do appear to work, and enable Enterprise OPA specific builtins to appear in the VSCode plugin with only a config change.

Remaining work:
- [x] implement additional LSP tests to ensure capabilities loading works as expected
- [x] implement additional LSP tests to ensure capabilities hot-reload works as expected 
- [x] document new `capabilities.Lookup` functionality
- [x] switch to fetching EOPA capabilities JSON during build
- [x] refactor LSP to avoid global lock for the `rego` package


pkg/capabilities/capabilities.go
```diff
@@ -29,9 +151,97 @@ func Lookup(engine, version string) (*ast.Capabilities, error) {
...
+	return ast.LoadCapabilitiesJSON(fd)
+}
+
+func lookupWebURL(parsedURL *url.URL) (*ast.Capabilities, error) {
+	req, err := http.NewRequest(http.MethodGet, parsedURL.String(), nil)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := http.DefaultClient.Do(req)
```
srenatus, line 169: [nit] Since we do nothing complex here, I think we can use the convenience method: `resp, err := http.Get(parsedURL.String())`

internal/lsp/rego/builtins.go
```diff
@@ -2,15 +2,23 @@ package rego
...
 
-var BuiltIns = builtinMap() //nolint:gochecknoglobals
+var BuiltInsLock = &sync.RWMutex{}                          // nolint:gochecknoglobals
+var BuiltIns = builtinMap(ast.CapabilitiesForThisVersion()) //nolint:gochecknoglobals
 
-func builtinMap() map[string]*ast.Builtin {
+func UpdateBuiltins(caps *ast.Capabilities) {
+	BuiltInsLock.Lock()
+	defer BuiltInsLock.Unlock()
+	BuiltIns = builtinMap(caps)
```
srenatus, line 16: [nit]: If we add a concurrency-safe update method here, should we perhaps make `Builtins` private (↔️ `builtins`)?

srenatus (approved):
Looks good so far 👍

pkg/capabilities/capabilities.go
charlesdaniels: I was not aware of that method, but that would be a nice simplification.

internal/lsp/rego/builtins.go
charlesdaniels: b618eb5761a86ddc7420f796a504a3d4b9057fce

pkg/capabilities/capabilities.go
charlesdaniels: 0527f06

build/do.rq
```diff
@@ -27,6 +27,17 @@ package script
...
+regal_root := rq.abs(rq.dir(rq.dir(rq.scriptpath())))
+
+# Root of the directory hierarchy where engine-specific capabilities JSON for
+# embedding into the Regal binary should be placed.
+embedded_caps_dir := rq.joinpath([regal_root, "pkg", "capabilities", "embedded"])
+
+# Directory where EOPA-specific JSON capabilities for embedding should be
+# placed.
+eopa_caps_dir := rq.joinpath([embedded_caps_dir, "eopa"])
+
```
srenatus, line 40: Hah. Eating your own dog food 🙃

build/do.rq
```diff
@@ -163,6 +182,28 @@ write_readme {
...
+		"srcdir": "capabilities",
+		"destdir": eopa_caps_dir
+	}
+
+	# NOTE(charles): My git config forces git to use ssh instead of http
+	# URLs, which is not desired in this case. It follows others my have
+	# similar setups locally. Setting GIT_CONFIG=/dev/null in this case
+	# prevents this.
+
+	fetch_cmd := rq.template("sh -c 'GIT_CONFIG=/dev/null git archive --remote={{.remote | quote}} {{.ref}} {{.srcdir | quote}} | tar -x -C {{.destdir | quote}}'", templ)
```
srenatus, line 202: No idea what's preferable, but if you'd consider it less of a hassle, [you could grab the zip](https://docs.github.com/en/repositories/working-with-files/using-files/downloading-source-code-archives):
https://github.com/StyraInc/enterprise-opa/archive/refs/heads/main.zip

internal/lsp/rego/builtins.go
```diff
@@ -7,13 +7,20 @@ import (
 	"github.com/open-policy-agent/opa/ast"
 )
 
-var BuiltInsLock = &sync.RWMutex{}                          // nolint:gochecknoglobals
-var BuiltIns = builtinMap(ast.CapabilitiesForThisVersion()) //nolint:gochecknoglobals
+var builtInsLock = &sync.RWMutex{}                          // nolint:gochecknoglobals
+var builtIns = builtinMap(ast.CapabilitiesForThisVersion()) //nolint:gochecknoglobals
```
srenatus, line 11: I think we can drop the nolint comment now.

build/do.rq
charlesdaniels: I was trying to find a good way to only have to download the subset of the repo that is needed, rather than the entire thing. If all else fails, I'll just do that, and/or clone the whole repo into `/tmp` or something.

I don't know _why_ the `git archive` thing doesn't work. StackOverflow says it should. The man page leads me to believe it should. But `run()` is complaining about a mismatched `'`, and on top of that when I run it manually, `git` complains about trying to clone a `git://` URL with the `ssh://` protocol, which is just bizarre.

build/do.rq
```diff
@@ -187,20 +189,102 @@ fetch_engine_caps {
...
+					"raw.fs": "/",
+					"raw.rs": "[\n\r]"
+				},
+			}
+		}
+	)
+
+	{ rq.error(sprintf("\nstdout: %s\nstderr: %s\n", [eopa_tags_result.stdout, eopa_tags_result.stderr])) | eopa_tags_result.exitcode != 0 }
+
+	eopa_tags := {r[2] | r := eopa_tags_result.stdout[_]}
```
anderseknert, line 241: Can we use `some ... in` iteration instead? Here and on line 257. This file isn't lintable at this point in time, as we'd need something similar for `rq` as this PR provides for `eopa`. But that doesn't mean should ignore the best practices and idioms Regal recommends.

pkg/capabilities/embedded/eopa/v0.100.5.json
```diff
@@ -0,0 +1 @@
+404: Not Found
```
anderseknert, line 1: You'll want to address this, I'm sure :) But it's actually a good reminder that we need to test somewhere that the files downloaded are legit.

anderseknert (commented):
Looking good! Just some minor remarks. There were other changes in the past though that I'm curious to know more about. Were they not needed?

internal/lsp/rego/builtins.go
charlesdaniels: done, forgot what commit it was in though 😅

pkg/capabilities/embedded/eopa/v0.100.5.json
charlesdaniels: 2 steps ahead of you 😉 646a37db8db27a6f6661f280c37962cd8a806a99

build/do.rq
charlesdaniels: I'm not actually sure how to use `some ... in`, can that handle additional clauses? I ended up having to filter the tag list down with further clauses.

charlesdaniels: hmm, yeah after 646a37db8db27a6f6661f280c37962cd8a806a99 I'm not sure `some ... in` can work, based on reading the docs

build/do.rq
charlesdaniels: I did get to try out a few patterns that I had in mind when I implemented various features/builtins, and I think they have held up well under real usage.

I ended up not needing it in the final version, but I did get to use `rq.template()`'s quoting feature. It's nice to be confident that strings being subbed into shell commands will have correct quoting, without having to fuss with backslashes and stuff. 

e.g. 

```plain
$ rq --raw 'rq.template("./myscript.sh {{.path | quote}}", {"path": "/some/path with/spaces in it"})'
./myscript.sh "/some/path with/spaces in it"
```

I think I still need to see a use case for this that does not involve shell scripting before I advocate for upstreaming rq's extras into OPA's template builtin though.

build/do.rq
```diff
@@ -163,6 +184,134 @@ write_readme {
...
+		"v0.49.0-5", # tag missing capabilities file
+		"v0.49.0-6", # tag missing capabilities file
+		"v0.49.0-7", # tag missing capabilities file
+		"v0.49.0-8", # tag missing capabilities file
+		"v1.15.0",   # tag missing capabilities file (misnamed v0.15.0)
+		"v1.4.1",    # tag missing capabilities file
+		"v1.5.0",    # tag missing capabilities file
+	}
+
+	eopa_tags := {t | r := eopa_tags_result.stdout[_]; t := r[2] ; not known_bad_tags[t]}
```
srenatus, line 259: I don't think we need the entire history of EOPA caps -- what for? As an arbitrary cut-off, I think I'd drop everything pre-1.0.0.

build/do.rq
```diff
@@ -163,6 +184,134 @@ write_readme {
...
+	}
+
+	print(sprintf("fetching %d capabilities files missing locally", [count(missing_locally)]))
+
+	# Download the capabilities from the constructed URLs.
+	new_caps := {
+		{"local": m.local, "content": c}
+		|
+		m := missing_locally[_]
+		print(sprintf("\tfetcing %s", [m.remote]))
```
srenatus, line 294:
```suggestion
		print(sprintf("\tfetching %s", [m.remote]))
```
or even
```suggestion
		print("\tfetching", m.remote)
```

build/do.rq
```diff
@@ -163,6 +184,134 @@ write_readme {
...
+		not regex.match("404: Not Found", c)
+	}
+
+	# Commit the retrieved content to disk.
+	{
+		true
+		|
+		cap := new_caps[_]
+		rq.write(cap.content, {"format": "raw", "file_path": cap.local})
+	}
```
srenatus, line 311: hehe funny. I'm not saying that we need to change anything here, but it seems like a thing OPA could eventually learn to optimize (early-exit style): with only constant elements before the `|` in a set or object comprehension, it could stop evaluating the body when it's succeeded _once_ 🤔

build/do.rq
charlesdaniels: 9d500efb5ceac61864441fd608918e5362f2ae1b

build/do.rq
charlesdaniels: 2efda57

build/do.rq
charlesdaniels: Hmm, that's a good point, I should probably move the `rq.write()` to the LHS of the comprehension. Just because it's safe _now_, I don't want this to bite us N years in the future when OPA has such an optimization. It's probably also clearer.

charlesdaniels: 39a51f0

build/do.rq
srenatus: Thanks!

build/do.rq
anderseknert: It works anywhere:

```
eopa_tags := {t | some r in eopa_tags_result.stdout; t := r[2] ; not known_bad_tags[t]}
```

charlesdaniels: Ahh I see, I didn't realize you could use it in that way. I'll switch it over.

README.md
```diff
@@ -364,6 +364,28 @@ reached. If no configuration file is found, Regal will use the default configura
 A custom configuration may be also be provided using the `--config-file`/`-c` option for `regal lint`, which when
 provided will be used to override the default configuration.
 
+### Loading Capabilities from URLs
```
anderseknert, line 367: This section should be under `## Capabilities` line 533 in your PR.

README.md
```diff
@@ -364,6 +364,28 @@ reached. If no configuration file is found, Regal will use the default configura
...
+
+Starting with Regal version TODO, Regal can load capabilities from URLs with
+the `file`, `http`, or `https` schemes using the `capabilities.from.url` config
```
anderseknert, line 370: I'd drop `file` here, unless there's someting that can do which isn't already covered by the other file loading option.

build/do.rq
```diff
@@ -163,6 +184,131 @@ write_readme {
...
+				"format": "raw",
+				"options": {
+					"raw.fs": "/",
+					"raw.rs": "[\n\r]"
+				},
+			}
+		}
+	)
+
+	{ rq.error(sprintf("\nstdout: %s\nstderr: %s\n", [eopa_tags_result.stdout, eopa_tags_result.stderr])) | eopa_tags_result.exitcode != 0 }
```
anderseknert, line 239: To avoid a really long line, and a fairly esoteric comprehension, I'd use a helper function like:

```rego
error_on_non_zero(result) if {
	result.exitcode != 0
	rq.error(sprintf("\nstdout: %s\nstderr: %s\n", [result.stdout, result.stderr])) 
}
```

build/do.rq
```diff
@@ -163,6 +184,131 @@ write_readme {
...
+		"v1.4.1",    # tag missing capabilities file
+		"v1.5.0",    # tag missing capabilities file
+	}
+
+	# Note that we use the `not startswith` to explicitly drop any
+	# pre-1.0.0 tags. There is little reason anyone would want them, and if
+	# they really have a need, they can manually download them and use the
+	# file locally.
+
+	eopa_tags := {t | some r in eopa_tags_result.stdout; t := r[2] ; not known_bad_tags[t] ; not startswith(t, "v0.")}
```
anderseknert, line 257: I guess the formatter doesn't work on .rq files, but since we use it everywhere else in the project, I'd prefer the same style here, i.e. with newlines instead of semicolons:
```rego
eopa_tags := {t |
    some r in eopa_tags_result.stdout
    t := r[2]
    not known_bad_tags[t]
    not startswith(t, "v0.")
}
```

build/do.rq
```diff
@@ -163,6 +184,131 @@ write_readme {
...
+		|
+
+		# construct the local path we expect the caps to exist at
+		t := eopa_tags[_]
+		p := rq.joinpath([eopa_caps_dir, sprintf("%s.json", [t])])
+
+		# eopa_caps_tree has object values, so if the path is missing,
+		# retrieving a non-object type is sufficient to check for
+		# nonexistance
+		object.get(eopa_caps_tree, p, "-") == "-"
```
anderseknert, line 279: can't you just do `eopa_caps_tree[p]`?

build/do.rq
```diff
@@ -163,6 +184,131 @@ write_readme {
...
+		m := missing_locally[_]
+		print("\tfetching ", m.remote)
+		resp := http.send({"url": m.remote, "method": "GET"})
+			
+		{ rq.error(sprintf("non-200 status code '%d' for URL '%s'", [resp.status_code, m.remote])) | resp.status_code != 200 }
+
+		c := resp.raw_body
+
+		# sanity check in case we got the URL wrong
+		not regex.match("404: Not Found", c)
```
anderseknert, line 300: this would have failed on `resp.status_code != 200 ` already, no?

internal/capabilities/capabilities_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+func TestLookupFromURL(t *testing.T) {
+	t.Parallel()
+
+	// Test that we can load a one of the existing OPA capabilities files
+	// via GitHub.
+
+	caps, err := Lookup(
+		context.Background(),
+		"https://raw.githubusercontent.com/open-policy-agent/opa/main/capabilities/v0.55.0.json",
```
anderseknert, line 44: This test should probably go in a separate file with a tag, like "integration", as I don't think unit tests should require an internet connection.

anderseknert (commented):
Great work on this!

I left a couple of comments. You can ignore the style-related ones if you wish, as Regal itself will tend to most of them once we can lint .rq files :)

The rest should be considered however. 

Very nice contribution!

build/do.rq
charlesdaniels: You also need an `else { true }`, or it will stop evaluation of the rule body it is called from.

build/do.rq
charlesdaniels: `not eopa_caps_tree[p]`

I switched it to that and it seems to work i nthis case, but I usually avoid using the `not` keyword as I have found it to be quite buggy in the past. It seems more reliable to simply write policies in ways that don't need it.

build/do.rq
anderseknert: Aha, I thought the purpose of `rq.error` was to blow up

build/do.rq
anderseknert: Nothing buggy about it, but some things may be unexpected :) I think mostly around using it with function calls, like:

```rego
not func(input.x)
```
which will fail if `input.x` is undefined, but won't trigger `not`, as it's effectively rewritten to something like:

```rego
x := input.x    # will fail here already
not func(x)
```

build/do.rq
charlesdaniels: It is, but the purpose of the esoteric comprehension is to make it _conditional_. With just the function as you wrote it, the happy path `exitcode == 0` case is `undefined` when you call the function directly.

README.md
charlesdaniels: b5cb5c5

README.md
charlesdaniels: b5cb5c5

build/do.rq
charlesdaniels: b5cb5c5

build/do.rq
charlesdaniels: b5cb5c5

build/do.rq
charlesdaniels: b5cb5c5

build/do.rq
charlesdaniels: b5cb5c5

internal/capabilities/capabilities_test.go
charlesdaniels: b5cb5c5

charlesdaniels:
Interestingly, I do see a similar test failure when I run in an Alpine docker container (ARM64). This smells like some concurrency problem to me.

anderseknert (approved):
Thanks a lot for this Charles! I'll merge this as it is, and will have a look at the build issues reported when merged.

PR #1001, opened 2024-08-20
Use new Roast library for custom AST serialization

anderseknert:
And the `jsoniter` library for any JSON handling.

This is quite a big win for linting performance:

```
➜ hyperfine --warmup 2 'regal_main lint bundle' 'regal_jsoniter lint bundle'
Benchmark 1: regal_main lint bundle
  Time (mean ± σ):      2.707 s ±  0.035 s    [User: 19.171 s, System: 0.636 s]
  Range (min … max):    2.662 s …  2.785 s    10 runs

Benchmark 2: regal_jsoniter lint bundle
  Time (mean ± σ):      2.185 s ±  0.019 s    [User: 14.932 s, System: 0.466 s]
  Range (min … max):    2.147 s …  2.207 s    10 runs

Summary
  regal_jsoniter lint bundle ran
    1.24 ± 0.02 times faster than regal_main lint bundle
```

But importantly, it doesn't make the AST harder to read or work with — quite the opposite actually!

See the Roast README for all the details:
https://github.com/anderseknert/roast


srenatus (approved):
LGTM. It's a bit of a bummer that we can't measure the impact of jsoniter alone vs perf improvements from the AST change. 🤔

anderseknert:
@srenatus yeah ... we'd have to use jsoniter in OPA for that too, or else all the marshalling of AST nodes will just pass through the encoding/json implementations. For the few places where we're serializing other things, like capabilities, the difference was negligible.

PR #1005, opened 2024-08-23
Add `input.json` completion provider

anderseknert:
This provider suggests completions based on the nested attributes found in the `input.json` file, if such a file exists.


bundle/regal/lsp/completion/providers/inputdotjson/inputdotjson.rego
```diff
@@ -0,0 +1,60 @@
...
+
+	# don't traverse into arrays
+	every value in path {
+		is_string(value)
+	}
+
+	input_type := type_name(value)
+
+	input_path := concat(".", ["input", concat(".", path)])
+	input_path != "input."
```
srenatus, line 59: Let's check `count(path) > 0` instead? 🤔

srenatus (approved):

PR #1011, opened 2024-08-27
Rule: `sprintf-arguments-mismatch`

### LINKED ISSUE #1009: Rule: `sprintf-arguments-mismatch`
If too many or too few arguments are in the array provided to `sprintf`, OPA will return the "error" back in the string created.
```
> opa eval -f raw 'sprintf("%v", [1,2,3])'
1%!(EXTRA int=2, int=3)
```
```
opa eval -f raw 'sprintf("%v", [])'
%!v(MISSING)
```
It would be helpful if Regal did a count of the args in the string and compared it to the number of args in the array provided (if known), and flagged this as an issue.

anderseknert:
Fixes #1009

### DIFF
--- a/bundle/regal/rules/bugs/sprintf_argument_mismatch.rego
+++ b/bundle/regal/rules/bugs/sprintf_argument_mismatch.rego
@@ -1,0 +1,48 @@
+# METADATA
+# description: Mismatch in `sprintf` arguments count
+package regal.rules.bugs["sprintf-arguments-mismatch"]
+
+import rego.v1
+
+import data.regal.ast
+import data.regal.config
+import data.regal.result
+
+# METADATA
+# description: Missing capability for built-in `sprintf`
+# custom:
+#   severity: none
+notices contains result.notice(rego.metadata.chain()) if not "sprintf" in object.keys(config.capabilities.builtins)
+
+# METADATA
+# description: |
+#   Count the number of distinct arguments ("verbs", denoted by %) in the string argument,
+#   compare it to the number of items in the array (if known), and flag when the numbers
+#   don't match
+report contains violation if {
+	some fn
+	ast.function_calls[_][fn].name == "sprintf"
+
+	fn.args[1].type == "array" # can only check static arrays, not vars
+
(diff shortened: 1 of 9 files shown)


bundle/regal/rules/bugs/sprintf_argument_mismatch.rego
```diff
@@ -0,0 +1,48 @@
...
+# METADATA
+# description: |
+#   Count the number of distinct arguments ("verbs", denoted by %) in the string argument,
+#   compare it to the number of items in the array (if known), and flag when the numbers
+#   don't match
+report contains violation if {
+	some fn
+	ast.function_calls[_][fn].name == "sprintf"
+
+	fn.args[1].type == "array" # can only check static arrays, not vars
```
srenatus, line 26: 😱 that's a possibility... `sprintf("%s %s %d", input)` etc. what a nightmare

bundle/regal/rules/bugs/sprintf_argument_mismatch.rego
```diff
@@ -0,0 +1,48 @@
...
+notices contains result.notice(rego.metadata.chain()) if not "sprintf" in object.keys(config.capabilities.builtins)
+
+# METADATA
+# description: |
+#   Count the number of distinct arguments ("verbs", denoted by %) in the string argument,
+#   compare it to the number of items in the array (if known), and flag when the numbers
+#   don't match
+report contains violation if {
+	some fn
+	ast.function_calls[_][fn].name == "sprintf"
```
srenatus, line 24: This looks expensive. Have you ever tried using an reverse-index here? Like "all locations where function `sprintf` is used"

anderseknert: I think we've mostly paid the cost at this point, as `ast.function_calls` essentially filters `ast.found.refs` keeping only function calls, and then providing them in compact format nicer to work with.

But yeah, `ast.found.refs` is a beast, and as I'm sure you remember, I tried to collect more types in that traversal, and things... broke :P

PR #1013, opened 2024-08-27
Remove some code made redundant by roast

anderseknert:
Some special case we previously needed to handle no longer apply since we moved to the roast format. Also fixed a TODO asking to move over to `strings.count` when released, and that is now :)


srenatus (approved):
Yay

charlieegan3 (approved):
👏 LGTM

PR #1019, opened 2024-08-28
Pin OPA version in Actions to v0.67.1

anderseknert:
This should honestly have been the case from the start..


charlesdaniels (approved):
👍

PR #1020, opened 2024-08-28
fix: don't serialize error location using roast but encoding/json

### LINKED ISSUE #1017: Language server fails to start when opening the OPA repo
I haven't looked into it yet, but it seems like `lsp.astError.Location` is getting marshalled into the roast format, but is then unmarshalled again somewhere, which unsurprisingly fails.

```
[Error - 7:54:52 PM] Server initialization failed.
  Message: failed to load workspace contents: failed to walk workspace dir "/Users/anderseknert/git/opa/opa": failed to update parse: failed to unmarshal parse errors: []lsp.astError: lsp.astError.Location: readObjectStart: expect { or n, but found ", error found in #10 byte of ...|ocation":"3:1:dmlvbG|..., bigger context ...|e":"var cannot be used for rule name","location":"3:1:dmlvbGF0aW9ucw=="},{"code":"rego_parse_error",|...
  Code: 0 
[Error - 7:54:52 PM] Regal LSP client client: couldn't create connection to server.
  Message: failed to load workspace contents: failed to walk workspace dir "/Users/anderseknert/git/opa/opa": failed to update parse: failed to unmarshal parse errors: []lsp.astError: lsp.astError.Location: readObjectStart: expect { or n, but found ", error found in #10 byte of ...|ocation":"3:1:dmlvbG|..., bigger context ...|e":"var cannot be used for rule name","location":"3:1:dmlvbGF0aW9ucw=="},{"code":"rego_parse_error",|...
  Code: 0 
ERROR: failed to refresh workspace contents: failed to walk workspace dir "/Users/anderseknert/git/opa/opa": failed to update parse: failed to unmarshal parse errors: []lsp.astError: lsp.astError.Location: readObjectStart: expect { or n, but found ", error found in #10 byte of ...|ocation":"7:1:YXV0aH|..., bigger context ...|e":"var cannot be used for rule name","location":"7:1:YXV0aHo="}]|...
[Error - 7:54:54 PM] Server process exited with signal SIGKILL.
[Error - 7:54:54 PM] Connection to server got closed. Server will not be restarted.
```

anderseknert:
Fixes #1017

Which turned into a bug hunt..

Also fixed:
- ast.resolved_imports crashing on duplicate identifiers
- node_modules not excluded from fs walk, as @srenatus reported

Opening the OPA repo now works as expected. The 500+ linter issues notwithstanding.

### DIFF
--- a/bundle/regal/ast/imports.rego
+++ b/bundle/regal/ast/imports.rego
@@ -19,19 +19,28 @@ imported_identifiers contains _imported_identifier(imp) if {
 	some imp in imports
 
 	imp.path.value[0].value in {"input", "data"}
+	count(imp.path.value) > 1
 }
 
 # METADATA
 # description: |
 #   map of all imported paths in the input module, keyed by their identifier or "namespace"
 resolved_imports[identifier] := path if {
-	some _import in imports
+	some identifier in imported_identifiers
 
-	_import.path.value[0].value == "data"
-	count(_import.path.value) > 1
+	# this should really be just a 1:1 mapping, but until OPA 1.0 we cannot
+	# trust that there are no duplicate imports, or imports shadowing other
+	# imports, which may render a runtime error here if two paths are written
+	# to the same identifier key ... simplify this post 1.0
+	paths := [path |
+		some imp in imports
 
-	identifier := _imported_identifier(_import)
-	path := [part.value | some part in _import.path.value]
+		_imported_identifier(imp) == identifier
+
+		path := [part.value | some part in imp.path.value]
+	]
+
+	path := paths[0]
 }
 
 # METADATA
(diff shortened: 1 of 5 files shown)


charlesdaniels (approved):
This looks reasonable to me, but since it seems like there are 3 different places where we want to filter out certain directories that should normally be ignored, maybe that should be centralized into one library in internal?

anderseknert:
@charlesdaniels yeah, I guess the [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) applies now.. I did look into that in the past but in the context of allowing providing that list via configuration... and that turned out to be complex. So they've remained hard-coded. You're right though — they could probably find a common location. I'll look into that next.

PR #1022, opened 2024-08-28
fix: respect ref head rules in `rule-name-repeats-package`

### LINKED ISSUE #1015: `rule-name-repeats-package` misses some classes of rules
Not sure if there are more cases, but at least the form `rule["key"] contains "value"` seems to be ignored by the rule.

Screenshot 2024-08-28 at 09 54 04

anderseknert:
Also made some adjustments to make the rule follow more modern conventions, and more importantly, to report end locations.

Fixes #1015

### DIFF
--- a/bundle/regal/rules/style/rule_name_repeats_package.rego
+++ b/bundle/regal/rules/style/rule_name_repeats_package.rego
@@ -1,19 +1,30 @@
 # METADATA
 # description: Rule name repeats package
 # related_resources:
-# - description: documentation
-#   ref: https://docs.styra.com/regal/rules/style/rule-name-repeats-package
-# schemas:
-# - input: schema.regal.ast
+#   - description: documentation
+#     ref: https://docs.styra.com/regal/rules/style/rule-name-repeats-package
 package regal.rules.style["rule-name-repeats-package"]
 
 import rego.v1
 
+import data.regal.ast
 import data.regal.result
 
-titleize(str) := upper(str) if count(str) == 1
+# METADATA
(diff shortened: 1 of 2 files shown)


bundle/regal/rules/style/rule_name_repeats_package.rego
```diff
@@ -23,51 +34,41 @@ titleize(str) := result if {
...
+		_num_package_path_components,
 	)
 }
 
-possible_offending_prefixes contains prefix if {
-	some combination in possible_path_component_combinations
+_possible_offending_prefixes contains prefix if {
+	some combination in _possible_path_component_combinations
 
 	prefix := concat("_", combination)
```
charlieegan3, line 57: Should we also catch things like this with this rule?

```rego
package foo.bar

import rego.v1

foo.bar := true
```

charlieegan3 (approved):

bundle/regal/rules/style/rule_name_repeats_package.rego
anderseknert: Good question! I'd say no, as at least applying the same logic as we currently do for _any_ path component would mean that something like:

```rego
package region.us.policies

user.permissions := { }
```

Would count `user` as repeating `us`... 

I think even for a simple naming scheme, this can sometimes be problematic. Like this one for example:

```rego
package my.db

dba if {
    "dba" in input.user.roles
}
```

A config toggle to only match full words would be nice. But it's also not a problem I've enountered in any real policy yet.

PR #1024, opened 2024-08-29
Rule: `directory-package-mismatch`

### LINKED ISSUE #1007: Rule: `directory-package-mismatch`
Following the "bundle structure" for policies should be considered idiomatic, as it makes it easy to find a package (like `foo.bar.baz`) just by navigating the directory tree (`foo/bar/baz`). 

While OPA only considers this structure when reading `data.json`/`data.yaml` files from bundle directories, this convention should be considered a best practice for policies as well. Note that we only consider _directory structure_ and **not** filenames, as there can be any number of files in a directory contributing to a package.

If there's a mismatch between the package name and the directory structure, the package name should be considered the "fact", and the suggestion should be to move the file into a matching directory structure.

This rule should also include a `regal fix` implementation, which automatically creates the required directory/directories as defined by the package path, and moves the file over there.

anderseknert:
Time to ruffle some feathers! Or move to the `custom` category, I haven't decided yet 😆 Either way, while this rule can certainly feel disruptive, larger policy projects *should* have a uniform structure.

Some notes on the implementation:
- Does not complain if a single file is provided, i.e. when there is no path except for the file. So `regal lint p.rego` will not result in a violation.
- Allows `_test` packages to either reside in the same directory as the package they test (this is the default) or to be kept in a separate directory matching the package exactly (optional).

Other than the rule implementation itself, this naturally required that we we restructured all of our own directories to follow this convention. I wasn't too sure about it at first, but the result definitely seems both clean and correct, so I think we're right about this. Also helped correct a few typos in names!

Fixes #1007


docs/rules/idiomatic/directory-package-mismatch.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Editor Support
+
+Editors integrating Regal's [language server](https://docs.styra.com/regal/language-server) will automatically display
+suggestions for idiomatic package paths based on the directory structure in which a policy resides. The image below
+demonstrates a new policy being created inside an `authorization/rbac/roles` directory, and the editor
+([via Regal](https://docs.styra.com/regal/language-server#code-completions)) suggesting the package path
+ `authorization.rbac.roles`.
+
+![package path auto-completion](../../assets/rules/pkg_name_completion.png)
```
charlieegan3, line 96: I think this will need to similar to this: https://github.com/StyraInc/docs/blob/main/docs/regal/language-server.md?plain=1#L26-L28

charlieegan3 (approved):
LGTM! It's an opinion, but opinions like these are going to really help new Rego users in particular here.

docs/rules/idiomatic/directory-package-mismatch.md
anderseknert: Ugh :/ But thanks! Fixed in last push.

PR #1033, opened 2024-09-01
Some schema fixes

anderseknert:
These don't matter to the type checker, so wasn't caught before.

### DIFF
--- a/internal/embeds/schemas/regal-ast.json
+++ b/internal/embeds/schemas/regal-ast.json
@@ -139,8 +139,7 @@
       "additionalProperties": false,
       "type": "object",
       "required": [
-        "head",
-        "body"
+        "head"
       ]
     },
     "else": {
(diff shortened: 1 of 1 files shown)


srenatus (approved):
How did you notice if the type checker doesn't care? 💭

anderseknert:
I moved the schema to the bundle directory, added this snippet to one of the entrypoints:
```rego
print(json.match_schema(input, data.regal.schema))
```
then ran `go run main.go lint --enable-print bundle` and watched the cascade of errors :) 

Should really look into making that part of the tests, but it's a little tricky as the schemas are placed where they are to appease Go's`embed` directive and its limitations on where it can embed files from... and I'd rather not have to deal with duplicates as I'd most certainly would forget to change things in two places. But I'm sure it's all solveable with some more time.

Issue #1029, opened 2024-08-29
For `regal fix`: determine if clean git working directory


anderseknert:
Really a "sub-task" of #1025, but may very well be applicable to later fixes too.

Instead of building some sort of "undo" functionality, it would be helpful if we could rely on Git for this. In order to that that, we should inform users if they're not on a clean working tree before we do any changes with `regal fix`. This should be possible to override with some flag, but by default we should bail out when:

1. We can't detect .git (or whatever way is best to do it) in the directory provided or any parent dir
2. The working tree isn't clean

PR #1035, opened 2024-09-01
Allow fixing directory structure not matching package paths

### LINKED ISSUE #1025: Provide `regal fix` and LSP Code Action for `directory-package-mismatch`
Having restructured our own code as part of building the `directory-package-mismatch` rule, I'm now more convinced that this change is good, or even great, for large projects in particular. In order to make it simpler for users to remediate this though, we'll need to follow up with a `regal fix` fixer (that potentially can fix entire projects) and a Code Action that can fix this for a single file from within the comfort of one's editor.

## regal fix

One thing we need to ensure here is that the fix must **not run** if there are notices emitted by the linter. Notices mean we don't have enough information to determine whether the package is correct or not. So either we respect that, or we make a new and better attempt to find out.

Another thing we'll need to ensure is that this command cleans up empty directories that it has left behind. So for example, if there's a project that looks like this:

```
code/
  foo/
    policy1.rego  # package policy.foo.bar
    policy2.rego. # package policy.foo.baz
README.md
```
Running `regal fix .` in the project root directory, the result would be:
```
policy/
  foo/
    bar/
      policy1.rego  # package policy.foo.bar
    baz/
      policy2.rego. # package policy.foo.baz
README.md
```
And the `code` directory no longer there.

* Ideally, we'd have a `--dry-run` option, that would print the structure after change without actually changing anything.. similar to the output of `tree`. But a better approach to be able to get this into the next release might be to try and detect if we're in a git repo, and that the working tree is clean. That way the user can try it out and easily revert using just git.

## Code Action

Should not be too complicated. Really just a new code action triggering an execute command which does the fix. Since it'll involve moving a file, we'll want to ensure the result is immediately apparent in the editor, and that things like possible issues are now reported at the new location.

anderseknert:
Following the recent addition of the `directory-package-mismatch` rule, this PR brings a corresponding `regal fix` remediator as well as a code action to fix this directly in the editor.

Fixes #1025

Todo:
- [x] Make it possible to set roots like how the OPA VS Code extension does
  - [x] For VS Code, we can send `opa.roots` as part of the init options, but what about updates?
  - [x] For `regal lint` we will need to either add a flag or read from config (or both)
- [ ] Docs for `regal fix` (including settings for DAS compliance)
- [ ] Docs for LSP integration (Code Action / Execute Command)
- [x] Setting config without `exclude-test-suffix` or to `false` leads to no Code Action shown
- [ ] #1029 
- [x] Show before/after paths in Code Action hover, and ideally in `regal fix`
  - [ ] Later... allow `--dry-run` for all fixes of `regal lint`

### DIFF
--- a/pkg/fixer/fixes/fixes.go
+++ b/pkg/fixer/fixes/fixes.go
@@ -3,11 +3,30 @@ package fixes
 import (
 	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/format"
+
+	"github.com/styrainc/regal/internal/lsp/clients"
+	"github.com/styrainc/regal/pkg/config"
 )
 
 // NewDefaultFixes returns a list of default fixes that are applied by the fix command.
 // When a new fix is added, it should be added to this list.
 func NewDefaultFixes() []Fix {
+	return []Fix{
+		&Fmt{
+			NameOverride: "use-rego-v1",
+			OPAFmtOpts: format.Opts{
+				RegoVersion: ast.RegoV0CompatV1,
+			},
+		},
+		&UseAssignmentOperator{},
+		&NoWhitespaceComment{},
+		&DirectoryPackageMismatch{},
+	}
+}
+
+// NewDefaultFormatterFixes returns a list of default fixes that are applied by the formatter.
+// Notably, this does not include fixers that move files around.
+func NewDefaultFormatterFixes() []Fix {
 	return []Fix{
 		&Fmt{
 			NameOverride: "use-rego-v1",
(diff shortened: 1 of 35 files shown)


tmp/unrelated.txt
```diff
@@ -0,0 +1 @@
+foo
```
charlieegan3, line 1: Added by mistake?

pkg/fixer/fixes/fixes.go
```diff
@@ -40,8 +50,22 @@ type FixCandidate struct {
...
 
+// Rename represents a file that has been moved (renamed).
+type Rename struct {
+	FromPath string
+	ToPath   string
+}
+
 // FixResult is returned from the Fix method and contains the new contents or fix recommendations.
-// In future this might support diff based updates, or renames.
+// In future this might support diff based updates.
```
charlieegan3, line 60: Very nice! as intended 🙂

tmp/unrelated.txt
anderseknert: Certainly! I’m surprised there aren’t more given how many tempdirs I’ve burned in this process 😄 Thanks for catching that. Will fix!

anderseknert:
Will defer docs to new PRs once this is in main.

charlieegan3 (approved):
lgtm, have been testing in my own PR and will open shortly as a follow on.

PR #1042, opened 2024-09-03
fix: Add support for dry run and dirty git abort

charlieegan3:


cmd/fix.go
```diff
@@ -263,18 +277,122 @@ func fix(args []string, params *fixCommandParams) error {
...
+
+	if !params.dryRun {
+		for _, file := range fileProvider.DeletedFiles() {
+			err := os.Remove(file)
+			if err != nil {
+				return fmt.Errorf("failed to delete file %s: %w", file, err)
+			}
+		}
+
+		for _, file := range fileProvider.ModifiedFiles() {
```
anderseknert, line 364: Something like a `fileprovider.WriteToDisk()` might be a nice abstraction.

pkg/fixer/fixes/fixes.go
```diff
@@ -52,6 +52,7 @@ type Fix interface {
 type RuntimeOptions struct {
 	// BaseDir is the base directory for the files being fixed. This is often the same as the
 	// workspace root directory, but not necessarily.
+	// TODO: use a slice and allow more than one?
```
anderseknert, line 55: I removed this, as it's no longer relevant.

cmd/fix.go
```diff
@@ -263,18 +277,122 @@ func fix(args []string, params *fixCommandParams) error {
...
+	}
 
 	fixReport, err := f.Fix(ctx, &l, fileProvider)
 	if err != nil {
 		return fmt.Errorf("failed to fix: %w", err)
 	}
 
+	gitRepo, err := git.FindGitRepo(args...)
+	if err != nil {
+		return fmt.Errorf("failed to establish git repo: %w", err)
```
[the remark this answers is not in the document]
charlieegan3: I've made it so this is only run when not in dry mode in d963eb5f8982956b9c09088707f0645baaba8358

cmd/fix.go
```diff
@@ -263,18 +277,122 @@ func fix(args []string, params *fixCommandParams) error {
...
+		return fmt.Errorf(
+			`the following files have been changed since the fixer was run:
+- %s
+please run fix from a clean state to support the use of git checkout for undo`,
+			strings.Join(conflictingFiles, "\n- "),
+		)
+	}
+
+	if params.verbose {
+		fmt.Fprintln(outputWriter, "Dry run mode enabled, the following changes would be made:")
```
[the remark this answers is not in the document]
charlieegan3: Fixed in d963eb5f8982956b9c09088707f0645baaba8358

cmd/fix.go
charlieegan3: Yeah, this is true. I guess that I'd count this as in the 'to refactor' list for now, we only have one file provider, and it's in memory. It wasn't immediately clear how this func would work relative paths etc.

pkg/fixer/fixes/fixes.go
charlieegan3: Removed in d963eb5f8982956b9c09088707f0645baaba8358

cmd/fix.go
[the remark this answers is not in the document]
anderseknert: Thanks!

PR #1046, opened 2024-09-04
codecov: Disable comments on PRs

charlieegan3:

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -7,7 +7,7 @@ concurrency:
 on:
   push:
     branches:
-      - main
+    - main
   pull_request:
   workflow_dispatch:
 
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -32,38 +32,38 @@ jobs:
...
-      - run: build/do.rq pull_request
-      - uses: golangci/golangci-lint-action@aaa42aa0628b4ae2578232a66b541047968fac86 # v6.1.0
-        if: matrix.os.name == 'linux'
-        with:
-          version: v1.60.3
-      - uses: actions/upload-artifact@50769540e7f4bd5e21e526ee35c689e35e0d6874 # v4.4.0
-        with:
-          name: regal-${{ matrix.os.name }}
-          path: regal
+    - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
```
charlieegan3, line 35: Updates indentation to what's used below...

srenatus (approved):
👏 👏 👏

.github/workflows/build.yaml
anderseknert: Can't we change it below instead? 😅 It's not super consitent, but the other files in this dir seem to prefer space for array items

anderseknert (approved):

.github/workflows/build.yaml
charlieegan3: yeah I can put it back.

PR #1049, opened 2024-09-04
fix: Update root detection

charlieegan3:
Now the .regal dir will be used as the root, or any bundles configured under it or manually set in the file, unless it does not exist. If this is the case, then the roots will be used as the args to the fix cmd.

### DIFF
--- a/e2e/cli_test.go
+++ b/e2e/cli_test.go
@@ -779,6 +779,7 @@ func TestFix(t *testing.T) {
 	td := t.TempDir()
 
 	initialState := map[string]string{
+		".regal/config.yaml": "", // needed to find the root in the right place
 		"foo/main.rego": `package wow
 
 import rego.v1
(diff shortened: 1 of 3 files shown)


e2e/cli_test.go
```diff
@@ -821,13 +822,13 @@ test_allow {
...
 
-	err := regal(&stdout, &stderr)("fix", td)
+	err := regal(&stdout, &stderr)("fix", filepath.Join(td, "foo"), filepath.Join(td, "bar"))
 
 	// 0 exit status is expected as all violations should have been fixed
 	expectExitCode(t, err, 0, &stdout, &stderr)
 
 	exp := fmt.Sprintf(`8 fixes applied:
-In project root: %s
+In project root: %[1]s
```
anderseknert, line 831: This should not be needed if only one

anderseknert (approved):
👍

e2e/cli_test.go
anderseknert: But I guess if the linter isn't complaining, neither should I :)

PR #1050, opened 2024-09-04
lsp: Clear old directories when renaming

charlieegan3:
This matches the behaviour of the fix cmd


charlieegan3:
The delete function is tested in the cli e2e test, but agree that it's not ideal...

Issue #1048, opened 2024-09-04
Automatically create package line (and `import rego.v1`) on .rego file created


anderseknert:
When I create a new .rego file in my editor, it would be neat if the rules applicable to that file was automatically "fixed" on creation. So that rather than seeing a blank file where I need to enter the package path and import rego.v1, that's written automatically to the file on creation.

<img width="293" alt="Screenshot 2024-09-04 at 12 30 00" src="https://github.com/user-attachments/assets/db518b3f-61d0-4b4d-ad89-ed88e2495265">

<img width="512" alt="Screenshot 2024-09-04 at 12 30 55" src="https://github.com/user-attachments/assets/317ee390-5789-4e6a-9ef4-70521b489b2a">

If `directory-package-mismatch` is disabled, we'll do nothing, just like now. If `use-rego-v1` is disabled, we'll simply not write that import out.

PR #1051, opened 2024-09-04
lsp: Template new empty files & template on format

charlieegan3:
Empty files will be created with a template content instead of being an error.

New empty files will be immediately updated with the template content.

Some client operations will not trigger the new file event, so in such cases, a save might be required to trigger the template update.

Fixes https://github.com/StyraInc/regal/issues/1048

https://github.com/user-attachments/assets/ebcf7699-bb96-43d6-ae21-f02a883f46e0

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -8,6 +8,8 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"regexp"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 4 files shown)


internal/lsp/server.go
```diff
@@ -1649,6 +1756,27 @@ func (l *LanguageServer) handleTextDocumentFormatting(
 		oldContent, ok = l.cache.GetFileContents(params.TextDocument.URI)
 	}
 
+	// if the file is empty, then the formatters will fail, so we template
+	// intstead
```
anderseknert, line 1760: instead

anderseknert (approved):
Very cool!

PR #1052, opened 2024-09-04
fix: Require use of --force to fix without git

charlieegan3:
This also skips the unchanged check when working within a git repo.


main.rego
```diff
@@ -0,0 +1 @@
+package main
```
anderseknert, line 1: :)

anderseknert (approved):
Remove that main.rego file and we can merge. Great work!

main.rego
charlieegan3: lol whoops

PR #1055, opened 2024-09-05
Fix LSP formatting that broke in v0.26.0

### LINKED ISSUE #1054: Formatting fails with `ERROR: failed to format file: runtime options are required`
Seems like we're not passing the runtime options in some code path. This is v0.26.0.

anderseknert:
Fixes #1054


charlieegan3 (commented):
Thanks too for adding a test

charlieegan3 (approved):

Issue #1010, opened 2024-08-26
Auto-completion of boolean values


anderseknert:
Not that it'll save you that many keystrokes, but it'd be good to see in the list of available options anyway.

<img width="431" alt="Screenshot 2024-08-26 at 14 09 23" src="https://github.com/user-attachments/assets/b6c6e15c-9bf9-4104-a573-890655e1ebe6">

(and on a related note, we should not suggest function calls in `default` rules, as they're not allowed there)

PR #1056, opened 2024-09-05
completion: Don't show builtins on default rules

charlieegan3:
Related to https://github.com/StyraInc/regal/issues/1010

### DIFF
--- a/internal/lsp/completions/providers/builtins.go
+++ b/internal/lsp/completions/providers/builtins.go
@@ -29,6 +29,11 @@ func (*BuiltIns) Run(c *cache.Cache, params types.CompletionParams, _ *Options) 
 		return []types.CompletionItem{}, nil
 	}
 
+	// default rules cannot contain calls
+	if strings.HasPrefix(strings.TrimSpace(currentLine), "default") {
+		return []types.CompletionItem{}, nil
+	}
+
 	words := patternWhiteSpace.Split(strings.TrimSpace(currentLine), -1)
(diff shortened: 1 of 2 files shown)


internal/lsp/completions/providers/builtins.go
```diff
@@ -29,6 +29,11 @@ func (*BuiltIns) Run(c *cache.Cache, params types.CompletionParams, _ *Options)
 		return []types.CompletionItem{}, nil
 	}
 
+	// default rules cannot contain calls
+	if strings.HasPrefix(strings.TrimSpace(currentLine), "default") {
```
srenatus, line 33:
```suggestion
	if strings.HasPrefix(strings.TrimSpace(currentLine), "default ") {
```
I could call my rule `defaultrule` and I'd be surprised why there are no builtin completions...? Or am I misreading  this? 😅

charlieegan3: I think that makes sense yeah!

srenatus (approved):
Thanks

internal/lsp/completions/providers/builtins_test.go
```diff
@@ -184,6 +184,39 @@ allow if c`
 	}
 }
 
+func TestBuiltIns_noDefeaultRule(t *testing.T) {
```
charlieegan3, line 187:
```suggestion
func TestBuiltIns_noDefaultRule(t *testing.T) {
```

PR #1058, opened 2024-09-05
Increase test coverage

anderseknert:
I started looking into increasing the test coverage the other day, and wow, there really turned out to be some bugs under those stones!

While there are a few silly additions here just for coverage, many of the changes made here uncovered issues that when fixed had Regal report many issues in our own code! So this was definitely a worthwhile endevour.

Will continue this work for sure, but to avoid it becoming a monster, and to leave it in a good state, I'll follow up with more PRs later.


anderseknert:
honest

charlieegan3 (approved):

PR #1059, opened 2024-09-05
lsp/completions: add boolean provider

charlieegan3:
Fixes https://github.com/StyraInc/regal/issues/1010

### DIFF
--- a/bundle/regal/lsp/completion/providers/booleans/booleans.rego
+++ b/bundle/regal/lsp/completion/providers/booleans/booleans.rego
@@ -1,0 +1,39 @@
+package regal.lsp.completion.providers.booleans
+
+import rego.v1
+
+import data.regal.lsp.completion.kind
+import data.regal.lsp.completion.location
+
+parsed_current_file := data.workspace.parsed[input.regal.file.uri]
+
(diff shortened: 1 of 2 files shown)


bundle/regal/lsp/completion/providers/booleans/booleans.rego
```diff
@@ -0,0 +1,39 @@
...
+
+	previous_word := words[words_on_line - 2]
+
+	previous_word in {"==", ":="}
+
+	word := location.word_at(line, input.regal.context.location.col)
+
+	some b in ["true", "false"]
+
+	startswith(b, word.text)
```
anderseknert, line 28:
```
some b in ["true", "false"]

startswith(b, word.text)
```

can be replaced by `strings.any_prefix_match`, I think

anderseknert (approved):
Nits, but looks great!

bundle/regal/lsp/completion/providers/booleans/booleans.rego
charlieegan3: I guess I need to know if we're going to be using true or false in the creation of the item though?

Issue #1021, opened 2024-08-28
Add parens to auto-completed function suggestions


anderseknert:
<img width="808" alt="Screenshot 2024-08-29 at 00 20 57" src="https://github.com/user-attachments/assets/cdb2cc1e-cafd-4407-baed-19773926bfd8">

It would be nice if auto-completed functions added parens to the replaced text, so that instead of:

```
print
```

you'd get

```
print()
```

with the cursor placed in the middle of the parens. Not sure if we can control that, and if not, perhaps just `print(` would be better.

PR #1062, opened 2024-09-05
lsp/completions: Support snippets based function signatures for built in suggestions

charlieegan3:
TODO: check this works ok with LSP signature help

At minimum, will fix https://github.com/StyraInc/regal/issues/1021


anderseknert:
While this looks promising, let's come back to this once we have implemented signature help.

charlieegan3:
👍 thanks for cleaning up!

PR #1069, opened 2024-09-06
fix: Have `use-some-for-output-vars` find comprehension body vars

### LINKED ISSUE #528: Not considering `some` used in the comprehension body when checking the comprehension assignment

anderseknert:
Previously reported as a false positive in some cases.

Fixes #528


anderseknert:
@nevumx finally ;) Will be included in the Regal v0.26.2 patch release due next week!

srenatus (approved):

PR #1070, opened 2024-09-06
build: introduce regal_standalone build flag, use for lint's "fix" hint

srenatus:
Projects using Regal via its Go module would no longer output the `regal fix --help` hint.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -16,6 +16,8 @@ builds:
     ignore:
       - goos: windows
         goarch: arm64
+    tags:
+      - regal_standalone
     ldflags:
       - -s -w
       - -X github.com/styrainc/regal/pkg/version.Version={{ .Version }}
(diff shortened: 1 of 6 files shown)


.goreleaser.yaml
```diff
@@ -68,7 +72,7 @@ checksum:
   name_template: "checksums.txt"
 
 snapshot:
-  name_template: "{{ incpatch .Version }}-next"
+  version_template: "{{ incpatch .Version }}-next"
```
srenatus, line 75: side-track, this was deprecated in some Gorelease version. I don't think we're using snapshot releases, but this will squelch the warning.

build/do.rq
srenatus, line 1: My editor did this 🙈 I've taught it to treat .rq files as rego. Yay or nay? I can back out of this change.

anderseknert (approved):

build/do.rq
anderseknert: The change makes the rules look less "task like"... but that's just a matter of time anyway :) So it's good.

PR #1078, opened 2024-09-09
lsp: Address empty module issues on rename

charlieegan3:
This PR makes changes to where the contents of files come from when being moved. Previously, preference was given to the contents on disk. This turned out to be incorrect as the files on disk are often empty until saved.

I found that I was able to replicate one of the empty module problems by:

- creating a file, and not saving it
- creating a dir
- editing the file
- moving the file into the dir while the file on disk is empty

before

https://github.com/user-attachments/assets/7bbb6442-5dc1-4897-81e2-8f2356d2e7c4

after

https://github.com/user-attachments/assets/c5a5ac35-1867-4214-8ad2-3cb9aa75fcd6

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -747,6 +747,8 @@ func (l *LanguageServer) StartTemplateWorker(ctx context.Context) {
 			newContents, err := l.templateContentsForFile(evt.URI)
 			if err != nil {
 				l.logError(fmt.Errorf("failed to template new file: %w", err))
+
+				continue
 			}
 
 			// generate the edit params for the templating operation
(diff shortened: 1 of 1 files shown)


internal/lsp/server.go
```diff
@@ -747,6 +747,8 @@ func (l *LanguageServer) StartTemplateWorker(ctx context.Context) {
 			newContents, err := l.templateContentsForFile(evt.URI)
 			if err != nil {
 				l.logError(fmt.Errorf("failed to template new file: %w", err))
+
+				continue
```
charlieegan3, line 751: This is not the fix, but when there is an error in templating we should skip the rest.

internal/lsp/server.go
```diff
@@ -790,6 +792,14 @@ func (l *LanguageServer) templateContentsForFile(fileURI string) (string, error)
 		return "", errors.New("file already has contents, templating not allowed")
 	}
 
+	diskContent, err := os.ReadFile(uri.ToPath(l.clientIdentifier, fileURI))
```
charlieegan3, line 795: this is also not the fix but is an additional guard for templating over files that have contents on disk already.

internal/lsp/server.go
```diff
@@ -1955,17 +1965,24 @@ func (l *LanguageServer) handleWorkspaceDidRenameFiles(
...
 		}
 
-		_, content, err := cache.UpdateCacheForURIFromDisk(
-			l.cache,
-			uri.FromPath(l.clientIdentifier, renameOp.NewURI),
-			uri.ToPath(l.clientIdentifier, renameOp.NewURI),
-		)
-		if err != nil {
-			return nil, fmt.Errorf("failed to update cache for uri %q: %w", renameOp.NewURI, err)
+		content, ok := l.cache.GetFileContents(renameOp.OldURI)
```
charlieegan3, line 1968: This is the fix for the empty module issue, were we load from the cache first, and only load from disk if it's missing or empty in the cache

internal/lsp/server.go
anderseknert: Any added safeguards are good 👍

internal/lsp/server.go
anderseknert: Nice!

anderseknert (approved):
LGTM! Test would be good, as with anything touching the LSP, but I know they can be hard to write given the many steps required in sequence. Let's look into how we improve on that later 👍

PR #1080, opened 2024-09-09
lsp: Always send empty diagnostics when deleted

charlieegan3:
Previously, we would only send empty diagnostics when the file had had diagnostics before, this did not include parse errors so when a file still had parse errors, it'd remain in client long after the server had cleared it.

This PR also makes some changes to process the cases of clearing files outside the `l.diagnosticRequestFile` loop as it was tricky to handle the types of empty updates in there.

before

https://github.com/user-attachments/assets/b5430a38-5686-44fb-bde7-422c80d915a5

after

https://github.com/user-attachments/assets/67838d79-1818-4fbe-a3da-782c6583b787

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -116,7 +116,6 @@ type LanguageServer struct {
 type fileUpdateEvent struct {
 	Reason  string
 	URI     string
-	OldURI  string
 	Content string
 }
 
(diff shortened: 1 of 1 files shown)


anderseknert (approved):
Looks good to me! Some test failure though :P

PR #1083, opened 2024-09-09
lsp: Correct Path to URI encoding

charlieegan3:
Spaces must be encoded as %20. This case causing the state worker to reload contents from disk as it appeared a file did not exist in the cache.

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -211,18 +211,14 @@ func (l *LanguageServer) StartDiagnosticsWorker(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		case evt := <-l.diagnosticRequestFile:
-			success, err := l.processTextContentUpdate(ctx, evt.URI, evt.Content)
+			err := l.processTextContentUpdate(ctx, evt.URI, evt.Content)
 			if err != nil {
 				l.logError(fmt.Errorf("failed to process text content update: %w", err))
 
 				continue
 			}
 
-			if !success {
-				continue
-			}
-
-			// otherwise, lint the file and send the diagnostics
+			// lint the file and send the diagnostics
 			err = updateFileDiagnostics(ctx, l.cache, l.loadedConfig, evt.URI, l.workspaceRootURI)
 			if err != nil {
 				l.logError(fmt.Errorf("failed to update file diagnostics: %w", err))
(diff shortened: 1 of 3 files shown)


internal/lsp/server.go
```diff
@@ -211,18 +211,14 @@ func (l *LanguageServer) StartDiagnosticsWorker(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		case evt := <-l.diagnosticRequestFile:
-			success, err := l.processTextContentUpdate(ctx, evt.URI, evt.Content)
+			err := l.processTextContentUpdate(ctx, evt.URI, evt.Content)
```
charlieegan3, line 214: This has been updated to avoid any special handling of parse errors. diagnostics are always sent from the parent now. This is unrelated to the title fix.

internal/lsp/server.go
```diff
@@ -936,33 +932,24 @@ func (l *LanguageServer) processTextContentUpdate(
 	ctx context.Context,
```
charlieegan3, line 932: See note above, this function now just attempts to update the parse if the content changed.

internal/lsp/server.go
```diff
@@ -2132,7 +2119,7 @@ func (l *LanguageServer) loadWorkspaceContents(ctx context.Context, newOnly bool
 		}
 
 		// if the caller has requested only new files, then we can exit early
-		if _, ok := l.cache.GetModule(fileURI); newOnly && ok {
+		if _, ok := l.cache.GetFileContents(fileURI); newOnly && ok {
```
charlieegan3, line 2122: The best proxy for 'Existence in the cache' is file contents and not a parsed module. This is related to the fix, but is in addition to the required change.

internal/lsp/uri/uri.go
```diff
@@ -19,6 +19,9 @@ func FromPath(client clients.Identifier, path string) string {
 		// Convert Windows path separators to Unix separators
 		path = filepath.ToSlash(path)
 
+		// spaces must be percent-encoded
+		path = strings.ReplaceAll(path, " ", "%20")
```
charlieegan3, line 23: This is the fix, going from paths to URIs when paths contained spaces was broken. This meant files were being reloaded from disk at incorrect times.

anderseknert: Good find!

anderseknert (approved):

PR #1084, opened 2024-09-09
e2e: Sort output from fix hint

charlieegan3:
This list is based on map keys and was causing flakes due to the items not always being in the same order.


anderseknert (approved):

srenatus (commented):
Ah my bad. Thanks!

Issue #1061, opened 2024-09-05
Templating from `directory-package-mismatch` should honor `exclude-test-suffix: false` setting


anderseknert:
The automatic templating from new file -> package path is magical ✨ 🧙 

https://github.com/user-attachments/assets/33eabb93-6c12-488b-a33c-30dbbcc4841f

One potential improvement here though is to honor the user's request to **not** exlude test suffixes in directory structure. So when a user:

- has configured `exclude-test-suffix: false` 
- creates `foo/bar/baz/policy_test.rego`
- we move it to `foo/bar/baz_test/policy_test.rego` and name the package package `foo.bar.baz_test`

PR #1085, opened 2024-09-09
lsp: Rename on template

charlieegan3:
Fixes https://github.com/StyraInc/regal/issues/1061

https://github.com/user-attachments/assets/a6589d87-6885-459d-a0a4-28e5821453a7

I have added some tests for the templating, I was going to determine the desired name for the file in that function too but in the end it was easier to keep the renaming logic separate. Even if this is more verbose, can be revisited tomorrow if we want to do some tidying here. Posting now just to allow testing until I'm next back working on this.


anderseknert (approved):
Very nice, and the thorough tests are excellent 👍

PR #1096, opened 2024-09-10
Add source action for to explore compiler stages

anderseknert:
This PR integrates @srenatus excellent
[opa-explorer](https://github.com/srenatus/opa-explorer/) with Regal, where we now start a web server along with the language server to host the explorer.

Users will now see a "Source Action" in the context menu of any given Rego file, and when clicked will bring them to the explorer view for that file.

Since a command for opening web pages client side is AFAIK only available in VS Code, this feature is sadly limited to that editor currently. Perhaps we can figure out some way to have the link show in Zed, similar to what we did for linter docs?

For now, we're only loading a single file into the explorer for compilation. We could perhaps consider loading all modules in the workspace too, as otherwise compilation will always fail on e.g. unknown functions from outside the policy.


internal/web/server.go
```diff
@@ -0,0 +1,179 @@
...
+type stringResult struct {
+	Show   bool
+	Class  string
+	Stage  string
+	Output string
+}
+
+func (s *Server) Start(_ context.Context) {
+	mux := http.NewServeMux()
+	mux.HandleFunc("GET /compile/", func(w http.ResponseWriter, r *http.Request) {
```
charlieegan3, line 103: I wonder if it's worth registering these under /explorer as we might want to have many things running under the web server further down the road?

I guess that we can always move them later if we add more functions to the server component.

internal/web/server.go
```diff
@@ -0,0 +1,179 @@
...
+	"github.com/styrainc/regal/internal/explorer"
+	"github.com/styrainc/regal/internal/lsp/cache"
+	"github.com/styrainc/regal/internal/lsp/clients"
+)
+
+//go:embed assets/*
+var assets embed.FS
+
+//nolint:gochecknoglobals,lll
+var tpl = template.Must(template.New("main").Parse(`
```
charlieegan3, line 22: Perhaps embed this template too?

internal/web/server.go
```diff
@@ -0,0 +1,179 @@
...
+<!DOCTYPE html>
+<html>
+	<head>
+		<meta charset="utf-8"/>
+		<link rel="stylesheet" href="/assets/missing.min.css">
```
charlieegan3, line 27: It looks like the missing.css file in the PR is unused? only the min.css is referenced here.

charlieegan3 (approved):
This is a really fun feature!

srenatus (approved):
Yay, this is a perfect new home for that code `<3`

internal/web/server.go
anderseknert: Yeah, I doubt people will bookmark / link to the compiler output of their favorite policy :P So we should be free to change names rather liberally. But OTOH I don't mind using "explorer" either, so I can do that.

internal/web/server.go
anderseknert: Ah yes, I guess we need the other one if we ever want to change anything? But we can always get it from @srenatus 😄

README.md
```diff
@@ -3,6 +3,7 @@
 [![Build Status](https://github.com/styrainc/regal/workflows/Build/badge.svg?branch=main)](https://github.com/styrainc/regal/actions)
 ![OPA v0.68.0](https://openpolicyagent.org/badge/v0.68.0)
 [![codecov](https://codecov.io/github/StyraInc/regal/graph/badge.svg?token=EQK01YF3X3)](https://codecov.io/github/StyraInc/regal)
+[![Downloads](https://img.shields.io/github/downloads/styrainc/regal/total.svg)](https://github.com/StyraInc/regal/releases)
```
srenatus, line 6: Hm I wonder if this really is interesting to anyone looking at the README. But I suppose it doesn't hurt -- since going through homebrew also just downloads the github release artifact, the number should be fairly accurate.

internal/web/server.go
srenatus: It's not mine, it's https://missing.style. -- @anderseknert had copied it into the opa-explorer repo at some point when they had changed their URLs or had an outage or something. We could go back to using their import instead and not copy any css here:

```html
<link rel="stylesheet" href="https://unpkg.com/missing.css@1.1.3">
```

internal/web/server.go
```diff
@@ -0,0 +1,179 @@
...
+	<body>
+		<header>
+			<h1>OPA Explorer<v-h>:</v-h><sub-title>Inspect compiler stages</sub-title></h1>
+			<nav>
+				<p class="tool-bar">
+					<img src="https://openpolicyagent.org/badge/v0.68.0"/>
+					<a href="https://github.com/srenatus/opa-explorer">GitHub</a>
+				</p>
+			</nav>
+		</header>
```
srenatus, line 38: Does the header still make sense? I'm not sure. I'd either slim it down to something that matches the source action description, or remove it.

internal/web/server.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+type stringResult struct {
+	Show   bool
+	Class  string
+	Stage  string
+	Output string
+}
+
+func (s *Server) Start(_ context.Context) {
```
srenatus, line 101:
```suggestion
func (s *Server) Start() {
```
If we don't use the context, let's remove it.

internal/web/server.go
```diff
@@ -0,0 +1,179 @@
...
+		</main>
+	</body>
+</html>
+`))
+
+type Server struct {
+	cache        *cache.Cache
+	workspaceURI string
+	client       clients.Identifier
+	baseURL      string
```
srenatus, line 70: I think this could be racy: we're setting the baseURL from a goroutine, but we retrieve it from `GetBaseURL`. It's probably unlikely to be a problem in real world usage, since it involves user interactions, though. OTOH adding a mutex should be simple enough.

internal/web/assets/screenshot.png
srenatus, line 1: This can go away, I think? Is it used?

srenatus (commented):
Some minor comments

internal/web/server.go
anderseknert: I've also embedded htmx now, so that we don't have any online dependencies

README.md
anderseknert: It's mostly just to show that Regal is an established project. 

The downloads should be fairly correct, although they don't account for https://github.com/StyraInc/regal/pkgs/container/regal

PR #1098, opened 2024-09-10
Rule: `comprehension-term-assignment`

### LINKED ISSUE #1073: Rule: `comprehension-term-assignment`
Somewhat similar to the [unconditional-assignment](https://docs.styra.com/regal/rules/style/unconditional-assignment) rule, and also in the style category. I think it would be good to have Regal help simplify comprehensions where the last expression is a simple assignment to a term variable, and suggest moving the value directly to the comprehension head instead.

**Avoid**
```rego
to_string(ref) := concat(".", [path |
    some part in ref
    path := part.value
])
```

**Prefer**
```rego
to_string(ref) := concat(".", [part.value | some part in ref])
```

Note that while it's often possible to move a **lot** of code in that direction, like function calls — I don't think that helps readability at all, and so we should stick to recommend this for simple assignment like the example shown above.

anderseknert:
Flag redundant assignment of simple values that could be used directly as the comprehension term (or key/value).

Fixes #1073

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,92 +197,93 @@ The following rules are currently available:
 
 <!-- RULES_TABLE_START -->
 
-|  Category   |                                                 Title                                                 |                        Description                        |
-|-------------|-------------------------------------------------------------------------------------------------------|-----------------------------------------------------------|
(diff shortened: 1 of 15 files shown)


charlieegan3 (approved):
👏

PR #1100, opened 2024-09-11
lsp: Update contents before diagnosticRequestFile

charlieegan3:
One of the main causes for the 'empty module' error stemmed from the sending of diagnosticRequestFile events with no content.

One solution would have been to fetch the content when processing events on the diagnosticRequestFile channel, however it felt like the publisher had more knowledge of how to do this correctly and processing diagnosticRequestFile was getting unwieldy.

I have removed the content from diagnosticRequestFile, and these events will just update the parse and send diagnostics. This should be less surprising.


internal/lsp/server.go
```diff
@@ -115,9 +115,8 @@ type LanguageServer struct {
 
 // fileUpdateEvent is sent to a channel when an update is required for a file.
 type fileUpdateEvent struct {
-	Reason  string
-	URI     string
-	Content string
```
charlieegan3, line 120: This has been removed so it's no longer possible to send fileUpdateEvents with no content, this reduces the responsibility on the diagnostics worker. Now it can just update parses and send diagnostics.

internal/lsp/server.go
```diff
@@ -212,9 +211,11 @@ func (l *LanguageServer) StartDiagnosticsWorker(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		case evt := <-l.diagnosticRequestFile:
-			err := l.processTextContentUpdate(ctx, evt.URI, evt.Content)
+			// updateParse will not return an error when the parsing failed,
```
charlieegan3, line 214: processTextContentUpdate has been removed and inlined here.

internal/lsp/server.go
```diff
@@ -260,9 +265,42 @@ func (l *LanguageServer) StartHoverWorker(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		case evt := <-l.builtinsPositionFile:
-			err := l.processHoverContentUpdate(ctx, evt.URI, evt.Content)
+			fileURI := evt.URI
```
charlieegan3, line 268: processHoverContentUpdate has also been inlined here.

anderseknert (approved):
I'd lie if I said I understood this fully, but I think I followed along pretty well :)

internal/lsp/server.go
```diff
@@ -813,9 +851,8 @@ func (l *LanguageServer) StartTemplateWorker(ctx context.Context) {
 
 			// finally, trigger a diagnostics run for the new file
 			updateEvent := fileUpdateEvent{
-				Reason:  "internal/templateNewFile",
-				URI:     fileURI,
-				Content: newContents,
+				Reason: "internal/templateNewFile",
```
charlieegan3, line 854: The actual issue this fixes was triggered by 

https://github.com/StyraInc/regal/pull/1100/files#diff-0077cc8440363b2137f8e3e405549e59083242b4281d50b08702a6b7153492a8R751-R756

Where an empty update was sent causing the cache's state to be set to blank for files when they were detected as new, despite them first having their contents loaded from disk.

Issue #1099, opened 2024-09-11
Fix `go test -race ./...`issues and run in CI


anderseknert:
Not sure what the impact of these issues are, but it'd be good to have a clean state here, and add this check to CI (in a new job, ideally not adding time to the total)
```
➜ go test -race ./...
?   	github.com/styrainc/regal	[no test files]
?   	github.com/styrainc/regal/bundle	[no test files]
?   	github.com/styrainc/regal/cmd	[no test files]
?   	github.com/styrainc/regal/internal/compile	[no test files]
?   	github.com/styrainc/regal/internal/dap	[no test files]
?   	github.com/styrainc/regal/internal/docs	[no test files]
?   	github.com/styrainc/regal/internal/embeds	[no test files]
?   	github.com/styrainc/regal/internal/explorer	[no test files]
?   	github.com/styrainc/regal/internal/git	[no test files]
ok  	github.com/styrainc/regal/internal/ast	1.662s
ok  	github.com/styrainc/regal/internal/capabilities	1.345s
?   	github.com/styrainc/regal/internal/lsp/cache	[no test files]
?   	github.com/styrainc/regal/internal/lsp/clients	[no test files]
?   	github.com/styrainc/regal/internal/lsp/examples	[no test files]
ok  	github.com/styrainc/regal/internal/capabilities/embedded	4.739s
ok  	github.com/styrainc/regal/internal/io	1.251s
?   	github.com/styrainc/regal/internal/lsp/opa/oracle	[no test files]
?   	github.com/styrainc/regal/internal/lsp/opa/scanner	[no test files]
?   	github.com/styrainc/regal/internal/lsp/opa/tokens	[no test files]
?   	github.com/styrainc/regal/internal/lsp/rego	[no test files]
?   	github.com/styrainc/regal/internal/lsp/types	[no test files]
?   	github.com/styrainc/regal/internal/lsp/types/completion	[no test files]
?   	github.com/styrainc/regal/internal/lsp/types/symbols	[no test files]
?   	github.com/styrainc/regal/internal/metrics	[no test files]
?   	github.com/styrainc/regal/internal/mode	[no test files]
?   	github.com/styrainc/regal/internal/novelty	[no test files]
?   	github.com/styrainc/regal/internal/test	[no test files]
?   	github.com/styrainc/regal/internal/testutil	[no test files]
?   	github.com/styrainc/regal/pkg/builtins	[no test files]
?   	github.com/styrainc/regal/pkg/report	[no test files]
[...]

PR #1102, opened 2024-09-11
Address races in test cases

charlieegan3:
There were three areas/issues:

- An error var in a config watcher test was accessed in two places
- server loaded config was accessed in a number of places without locking
- a panic caused by early closing of the net.Pipe in the lsp server tests.

This PR addresses these issues and runs the race check on PRs during the Linux build.

Fixes https://github.com/StyraInc/regal/issues/1099

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -56,6 +56,8 @@ jobs:
           path: ~/go/bin/rq
           key: ${{ runner.os }}-${{ runner.arch }}-go-rq-${{ env.RQ_VERSION }}
       - run: build/do.rq pull_request
+      - run: go test -race ./...
+        if: matrix.os.name == 'linux'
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yaml
```diff
@@ -56,6 +56,8 @@ jobs:
           path: ~/go/bin/rq
           key: ${{ runner.os }}-${{ runner.arch }}-go-rq-${{ env.RQ_VERSION }}
       - run: build/do.rq pull_request
+      - run: go test -race ./...
+        if: matrix.os.name == 'linux'
```
anderseknert, line 60: I guess this is tolerable since the intel macos build is going to be slower anyway 😄 But when we get rid of that, we should run this in a parallel job.

anderseknert (approved):
Quite a chore! Well done 👏

PR #1103, opened 2024-09-12
Add `Debug` Code Lens

anderseknert:
Following the excellent work on debugger support by @johanfylling, this PR introduces a code lens for debugging, which triggers a command that simply returns a launch configuration where the package or rule clicked is set as the entrypoint.

The code lens provider is now also rewritten in Rego.

### DIFF
--- a/bundle/regal/lsp/codelens/codelens.rego
+++ b/bundle/regal/lsp/codelens/codelens.rego
@@ -1,0 +1,82 @@
+# METADATA
+# description: |
+#   the code lens provider decides where code lenses should be placed in the given input file
+# schemas:
+#   - input: schema.regal.ast
+package regal.lsp.codelens
+
+import rego.v1
+
+import data.regal.ast
+import data.regal.result
+import data.regal.util
+
+import data.regal.lsp.util.location
+
+# code lenses are displayed in the order they come back in the returned
+# array, and 'evaluate' somehow feels better to the left of 'debug'
+lenses := array.concat(
+	[l | some l in _eval_lenses],
+	[l | some l in _debug_lenses],
+)
+
+_eval_lenses contains {
+	"range": location.to_range(result.ranged_location_from_text(input["package"]).location),
+	"command": {
+		"title": "Evaluate",
+		"command": "regal.eval",
+		"arguments": [
+			input.regal.file.name,
+			ast.ref_to_string(input["package"].path),
+			util.to_location_object(input["package"].location).row,
+		],
+	},
+}
+
+_eval_lenses contains _rule_lens(rule, "regal.eval", "Evaluate") if {
+	some rule in ast.rules
+}
+
+_debug_lenses contains {
+	"range": location.to_range(result.ranged_location_from_text(input["package"]).location),
+	"command": {
+		"title": "Debug",
+		"command": "regal.debug",
+		"arguments": [
+			input.regal.file.name,
+			ast.ref_to_string(input["package"].path),
+			util.to_location_object(input["package"].location).row,
+		],
+	},
+}
+
+_debug_lenses contains _rule_lens(rule, "regal.debug", "Debug") if {
+	some rule in ast.rules
+
+	# no need to add a debug lens for a rule like `pi := 3.14`
+	not _unconditional_constant(rule)
+}
+
+_rule_lens(rule, command, title) := {
+	"range": location.to_range(result.ranged_location_from_text(rule).location),
+	"command": {
+		"title": title,
+		"command": command,
+		"arguments": [
+			input.regal.file.name, # regal ignore:external-reference
+			sprintf("%s.%s", [ast.ref_to_string(input["package"].path), ast.ref_static_to_string(rule.head.ref)]),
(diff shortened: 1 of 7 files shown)


.vscode/settings.json
```diff
@@ -4,6 +4,7 @@
     "OPA_EVAL_CAPABILITIES": "${workspacePath}/build/capabilities.json"
   },
   "opa.roots": [
+    "${workspaceFolder}/lab",
```
srenatus, line 7: Is this related to the PR?

srenatus (approved):
looks good to me

internal/lsp/server.go
```diff
@@ -499,6 +499,52 @@ func (l *LanguageServer) StartCommandWorker(ctx context.Context) { // nolint:mai
...
 				// handle this ourselves as it's a rename and not a content edit
 				fixed = false
+			case "regal.debug":
+				if l.clientIdentifier != clients.IdentifierVSCode {
+					l.logError(errors.New("regal.debug is only supported in VSCode"))
+
+					break
+				}
+
+				if len(params.Arguments) != 3 {
```
johanfylling, line 509: I don't see us reading arg `2` anywhere. Looks like it's the line number for `regal.eval`, which I suppose we don't need for debug. Do we still need to assert its presence?

bundle/regal/lsp/codelens/codelens.rego
```diff
@@ -0,0 +1,82 @@
...
+}
+
+_eval_lenses contains _rule_lens(rule, "regal.eval", "Evaluate") if {
+	some rule in ast.rules
+}
+
+_debug_lenses contains {
+	"range": location.to_range(result.ranged_location_from_text(input["package"]).location),
+	"command": {
+		"title": "Debug",
```
johanfylling, line 43: Does vscode support emojis? If it does, we could prefix this with a '🪲' 😄. 
There's probably plenty of setups where that would break, though.

bundle/regal/lsp/codelens/codelens.rego
```diff
@@ -0,0 +1,82 @@
...
+			ast.ref_to_string(input["package"].path),
+			util.to_location_object(input["package"].location).row,
+		],
+	},
+}
+
+_debug_lenses contains _rule_lens(rule, "regal.debug", "Debug") if {
+	some rule in ast.rules
+
+	# no need to add a debug lens for a rule like `pi := 3.14``
```
johanfylling, line 56: 👍

bundle/regal/lsp/codelens/codelens.rego
```diff
@@ -0,0 +1,82 @@
...
+}
+
+_rule_lens(rule, command, title) := {
+	"range": location.to_range(result.ranged_location_from_text(rule).location),
+	"command": {
+		"title": title,
+		"command": command,
+		"arguments": [
+			input.regal.file.name, # regal ignore:external-reference
+			sprintf("%s.%s", [ast.ref_to_string(input["package"].path), ast.ref_static_to_string(rule.head.ref)]),
```
johanfylling, line 67: Does this strip the dynamic tail off of the rule's ref?

I suppose that it doesn't hurt that vars are transferred to the query, as I think the only side effect is that they will be unified for the result(?). Worst case, the result set will just be large for a large set of enumerated values.
I can even imagine this being a benefit, as it'd possibly include fewer secondary rules with intersecting paths, as these won't satisfy the vars 🤔 .

johanfylling (approved):
LGTM 👍 😃

bundle/regal/lsp/codelens/codelens.rego
anderseknert: It does support that, and since we're only targeting VS Code for now, we could do it. Not sure this is intuitive though. Most people will probably think they got some dirt on their monitor 🤣

bugs

bundle/regal/lsp/codelens/codelens.rego
anderseknert: Yes, `ast.ref_static_to_string` stops at the first dynamic part of a ref. Perhaps we can experiment with dynamic refs later, but I think this is the option that makes the most sense to me.

internal/lsp/server.go
anderseknert: No, you're right, we don't need to assert this here, but we're using the same outputs for both code lenses... and we'll want to check this to avoid getting out of bounds errors in case there aren't at least 2 args.

.vscode/settings.json
anderseknert: No, thanks! Removed. Although it _is_ rather tricky to get it right with the "checkOnSave" feature when you want it outside of the main bundle but not the whole workspace... I'll guess I'll have to do some git magic to make sure this line remains unseen. I've removed it from the change now anyway.

bundle/regal/lsp/codelens/codelens.rego
johanfylling: I was thinking something like `► Evaluate`|`🪲Debug`

Issue #1104, opened 2024-09-12
sprintf-args check false positive when format is a string variable


srenatus:
```rego
package play

import rego.v1

bye if {
	query_fmt := "SELECT name FROM roles INNER JOIN users ON roles.user_id = users.id WHERE users.name = %s"
	_ = sprintf(query_fmt, ["me"])
}
```

This shouldn't be flagged, but it is. [Playground](https://play.openpolicyagent.org/p/wFnWFoOg7L)

anderseknert:
Thanks! Hmm, yeah, that should be easy to account for, I think. I'll take a look.

PR #1106, opened 2024-09-12
More stringent ast.ref_to_string

anderseknert:
While I started looking into #1104 I encountered some cases where `ast.ref_to_string`, and it's `static` counterpart would return wrong representations. We already had a bit of a homegrown format, which was convenient for some things, but would also leak out in places where it shouldn't be. Now use the same format as OPA for representing refs as strings.

Also added the query in error messages where we fail to prepare or eval in a couple of places, as the lack of those made debugging this issue much harder than it had to be.

### DIFF
--- a/bundle/regal/ast/ast.rego
+++ b/bundle/regal/ast/ast.rego
@@ -35,6 +35,8 @@ is_constant(value) if {
 	not has_term_var(value.value)
 }
 
+default builtin_names := set()
+
 builtin_names := object.keys(config.capabilities.builtins)
 
 builtin_namespaces contains namespace if {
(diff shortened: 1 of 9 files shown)


internal/lsp/server.go
```diff
@@ -575,7 +575,7 @@ func (l *LanguageServer) StartCommandWorker(ctx context.Context) { // nolint:mai
 
 				responseParams := map[string]any{
 					"type":        "opa-debug",
-					"name":        "Debug " + path,
```
johanfylling, line 578: 👍

bundle/regal/ast/ast.rego
```diff
@@ -183,28 +185,34 @@ _exclude_arg("assign", 0, _)
...
 # description: |
 #   returns the string representation of a ref up until its first
 #   non-static (i.e. variable) value, if any:
 #   foo.bar -> foo.bar
 #   foo.bar[baz] -> foo.bar
-ref_static_to_string(ref) := ss if {
+ref_static_to_string(ref) := str if {
 	rs := ref_to_string(ref)
-	ss := substring(rs, 0, indexof(rs, ".$"))
+	str := _trim_from_var(rs, regex.find_n(`\[[^"]`, rs, 1))
```
johanfylling, line 209: What if the ref is `foo.bar[7]`? Or this isn't expected?

johanfylling (approved):
👍

PR #1107, opened 2024-09-12
fix: `sprintf-arguments-mismatch`false positive when var used as format

### LINKED ISSUE #1104: sprintf-args check false positive when format is a string variable

anderseknert:
While this rule likely never will catch all violations, as some are just too expensive to find for something very unlikely — it should never report false positives.

Fixes #1104


srenatus (approved):
Thanks

PR #1112, opened 2024-09-16
lsp: Address test concurrency issues

charlieegan3:
This PR is intended to address two issues with the test suite that happen when running tests concurrently:

TestLanguageServerSingleFile fails sometimes as the server logs an error when the context has been cancelled, and it's a panic when logging to t, after the test is over.

https://github.com/StyraInc/regal/actions/runs/10875806798/job/30174833322

TestGetInlayHintsAstTerms will fail when TestLanguageServerSingleFile is running sometimes as [TestLanguageServerSingleFile updates the contents of the map](https://github.com/StyraInc/regal/blob/7bf76856cc72a2d72a1834e5e067c64675c7fc8b/internal/lsp/server.go#L390) while TestGetInlayHintsAstTerms is iterating over it.

https://github.com/StyraInc/regal/actions/runs/10875806798/job/30184826633

Coincidentally, both these issues happened on a recent community PR, which underlines the importance of addressing these as they're a bad contributor experience.

This PR also includes, in a second commit https://github.com/StyraInc/regal/pull/1112/commits/554f6120a575d6cdc9043284261a4eb5afdf12d3, an update to rego.QueryRegalBundle to accept a context. See commit message, but there appears to be a recurring issue here where tests timeout.

* https://github.com/StyraInc/regal/actions/runs/10883075294/job/30195419440?pr=1110
* https://github.com/StyraInc/regal/actions/runs/10883075932/job/30195421353

IO have updated the function an callers to pass a context to this function to allow us to better understand this issue.

### DIFF
--- a/internal/lsp/completions/providers/builtins_test.go
+++ b/internal/lsp/completions/providers/builtins_test.go
@@ -1,11 +1,10 @@
 package providers
 
 import (
+	"context"
 	"slices"
 	"strings"
 	"testing"
-
-	"golang.org/x/net/context"
 
 	"github.com/styrainc/regal/internal/lsp/cache"
 	"github.com/styrainc/regal/internal/lsp/types"
--- a/internal/lsp/completions/providers/builtins_test.go
+++ b/internal/lsp/completions/providers/builtins_test.go
@@ -5,6 +5,8 @@ import (
 	"strings"
 	"testing"
 
+	"golang.org/x/net/context"
+
 	"github.com/styrainc/regal/internal/lsp/cache"
 	"github.com/styrainc/regal/internal/lsp/types"
 )
@@ -32,7 +34,7 @@ allow if c`
 		},
 	}
 
-	completions, err := p.Run(c, completionParams, nil)
+	completions, err := p.Run(context.Background(), c, completionParams, nil)
 	if err != nil {
 		t.Fatalf("Unexpected error: %v", err)
 	}
@@ -67,7 +69,7 @@ allow := c`
 		},
 	}
 
-	completions, err := p.Run(c, completionParams, nil)
+	completions, err := p.Run(context.Background(), c, completionParams, nil)
 	if err != nil {
 		t.Fatalf("Unexpected error: %v", err)
 	}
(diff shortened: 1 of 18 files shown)


internal/lsp/rego/builtins.go
```diff
@@ -23,7 +24,7 @@ func GetBuiltins() map[string]*ast.Builtin {
 	builtInsLock.RLock()
 	defer builtInsLock.RUnlock()
 
-	return builtIns
+	return maps.Clone(builtIns)
```
charlieegan3, line 27: Similar to what we have in the cache, we return a copy instead.

internal/lsp/rego/rego.go
```diff
@@ -56,6 +56,8 @@ func LocationFromPosition(pos types.Position) *ast.Location {
 func AllBuiltinCalls(module *ast.Module) []BuiltInCall {
 	builtinCalls := make([]BuiltInCall, 0)
 
+	bis := GetBuiltins()
```
charlieegan3, line 59: Loading the builtins once also seemed like a good plan here since we're now loading a copy.

internal/lsp/server_test.go
```diff
@@ -1095,17 +1097,39 @@ func createConnections(
...
-	return &testLogger{t: t}
+	tl := &testLogger{t: t, open: true}
+
+	// using cleanup ensure that no goroutines attempt to write to the logger
+	// after the test has been cleaned up
+	t.Cleanup(func() {
+		tl.mu.Lock()
+		defer tl.mu.Unlock()
+		tl.open = false
+	})
```
charlieegan3, line 1114: When the test is over, the logger is closed to stop it accepting more logs and causing a panic when the test is done.

internal/lsp/completions/providers/builtins_test.go
```diff
@@ -5,6 +5,8 @@ import (
 	"strings"
 	"testing"
 
+	"golang.org/x/net/context"
```
srenatus, line 8: This shouldn't be used anymore, probably an auto import

charlieegan3: oops, thanks, have added a new commit to address

charlieegan3:
Thanks Stephan, appreciate the review here 🙏

PR #1114, opened 2024-09-17
docs: correct line endings for GIF files

charlieegan3:
This issue has been fixed in the docs site already, but this change is needed to make sure the next automatic update docs not break.


srenatus (approved):
Thanks

Issue #1119, opened 2024-09-19
fix: Output a better error when attempting to rename files to the same destination


charlieegan3:
When attempting to fix via rename files with the same name and package in different directories, regal fix will fail with an unhelpful error.

```
…/regalbundle main ➜ cat example/example_test.rego 
───────┬───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: example/example_test.rego
───────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ package example_test
   2   │ 
   3   │ import rego.v1
───────┴───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
…/regalbundle main ➜ cat example2/example_test.rego 
───────┬───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: example2/example_test.rego
───────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ package example_test
   2   │ 
   3   │ import rego.v1
───────┴───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
…/regalbundle main ➜ regal fix .
failed to fix: failed to rename file: rename conflict: file /Users/charlieegan3/Desktop/regalbundle/example_test/example_test.rego already exists
```

PR #1120, opened 2024-09-19
fixer: Address rename conflicts with explanation

charlieegan3:
Fixes https://github.com/StyraInc/regal/issues/1119

### DIFF
--- a/cmd/fix.go
+++ b/cmd/fix.go
@@ -313,9 +313,23 @@ func fix(args []string, params *fixCommandParams) error {
 		return fmt.Errorf("failed to create file provider: %w", err)
 	}
 
+	r, err := fixer.ReporterForFormat(params.format, outputWriter)
+	if err != nil {
+		return fmt.Errorf("failed to create reporter for format %s: %w", params.format, err)
+	}
+
 	fixReport, err := f.Fix(ctx, &l, fileProvider)
 	if err != nil {
 		return fmt.Errorf("failed to fix: %w", err)
+	}
+
+	if fixReport.HasConflicts() {
+		err = r.Report(fixReport)
+		if err != nil {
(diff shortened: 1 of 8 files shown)


cmd/fix.go
```diff
@@ -428,13 +442,6 @@ please run fix from a clean state to support the use of git checkout for undo`,
 		}
 	}
 
-	r, err := fixer.ReporterForFormat(params.format, outputWriter)
-	if err != nil {
-		return fmt.Errorf("failed to create reporter for format %s: %w", params.format, err)
-	}
-
-	r.SetDryRun(params.dryRun)
```
anderseknert, line 436: Should this not be set any more?

e2e/cli_test.go
```diff
@@ -938,6 +938,79 @@ test_allow := true
...
+		"foo/foo.rego": `package foo
+
+import rego.v1
+`,
+		// this file should be at foo/foo.rego, but that file already exists
+		"quz/foo.rego": `package foo
+
+import rego.v1
+`,
+		// these thre files should all be at bar/bar.rego, but they cannot all be moved there
```
anderseknert, line 960: thre -> three / 3

pkg/fixer/fileprovider/inmem.go
```diff
@@ -11,6 +11,15 @@ import (
 	"github.com/styrainc/regal/pkg/rules"
 )
 
+type RenameConflictError struct {
```
anderseknert, line 14: Nice! We should do more of this.

anderseknert (approved):
LGTM! I guess one option for later could be to simply create foo1.rego, foo2.rego, and so on, in case of conflicts, instead of failing.. at least if configured to do so?

PR #1122, opened 2024-09-19
Allow `regal lint -` to lint from stdin

### LINKED ISSUE #1118: feature: lint stdin
It would be nice to be able to pipe some Rego into Regal, and have it output anything it could complain about:

```
magic-command-outputting-rego | regal lint -
```

Now, anything that depends on file names would automatically be disabled: we have no file name.

Not a pressing issue in any way. I suppose the user base for this feature is imperceptibly small.

anderseknert:
And print notices for rules ignored as a result of no filename being known.

```
➜ cat lab/p.rego | go run main.go lint -
1 file linted. No violations found. 2 rules skipped:
- directory-package-mismatch: disabled when filename is unknown
- file-missing-test-suffix: disabled when filename is unknown
```

Fixes #1118

### DIFF
--- a/bundle/regal/config/config.rego
+++ b/bundle/regal/config/config.rego
@@ -11,7 +11,9 @@ docs["resolve_url"](url, category) := replace(
 
 merged_config := data.internal.combined_config
 
-capabilities := merged_config.capabilities
+capabilities := object.union(merged_config.capabilities, {"special": special})
+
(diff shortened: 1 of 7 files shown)


srenatus (approved):
Code looks good, I'll try it later today

anderseknert:
Let me know if you find anything to improve!

srenatus:
Works beautifully :sparkles: thanks again!

PR #1127, opened 2024-09-23
fixer: add --on-conflict flag to support renaming

charlieegan3:
This PR adds a new flag to the fix command, `--on-conflict`. This flag can be used to specify the behavior of the fix command when a file's new name conflicts with an existing file or another file's fixed name.

Setting --on-conflict to "rename" will cause the fix command to rename the file to a different name at the target location. The new name will be generated by appending a number to the end of the file name.

### DIFF
--- a/e2e/cli_test.go
+++ b/e2e/cli_test.go
@@ -1011,6 +1011,116 @@ Cannot move multiple files to: bar/bar.rego
 	}
 }
 
+func TestFixWithConflictRenaming(t *testing.T) {
+	t.Parallel()
+
+	stdout := bytes.Buffer{}
+	stderr := bytes.Buffer{}
+	td := t.TempDir()
+
+	initialState := map[string]string{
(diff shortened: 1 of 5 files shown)


e2e/cli_test.go
```diff
@@ -1011,6 +1011,116 @@ Cannot move multiple files to: bar/bar.rego
...
+	// 0 exit status is expected as all violations should have been fixed
+	expectExitCode(t, err, 0, &stdout, &stderr)
+
+	expStdout := fmt.Sprintf(`3 fixes applied:
+In project root: %[1]s
+foo/bar.rego -> bar/bar.rego:
+- directory-package-mismatch
+quz/foo.rego -> foo/foo_1.rego:
+- directory-package-mismatch
+quz/foo_test.rego -> foo/foo_1_test.rego:
```
anderseknert, line 1067: nice 👍

anderseknert (approved):
Looks great!

PR #1129, opened 2024-09-23
tests: Address incorrect built-in function check

charlieegan3:
After seeing https://github.com/StyraInc/regal/actions/runs/10996729351/job/30530568473, I had another look at this one. I spotted the built-in is incorrect here, I think? Not thinking this is the main fix, but this should log to make it clear when we see the next failure.

### DIFF
--- a/internal/lsp/server_test.go
+++ b/internal/lsp/server_test.go
@@ -336,12 +336,13 @@ capabilities:
 		}
 	}
 
-	// NOTE(charles): the configuration is updated asynchronously from the
-	// thread the test is running in. This check prevents a race condition
-	// where the completion test runs before rego.builtIns is updated which
-	// was causing flaky tests in CI.
+	// manually inspect the server's list of builtins to ensure that the EOPA
+	// capabilities were loaded correctly.
 	timeout = time.NewTimer(defaultTimeout)
 	defer timeout.Stop()
+
+	ticker := time.NewTicker(100 * time.Millisecond)
+	defer ticker.Stop()
 
 	for {
 		success := false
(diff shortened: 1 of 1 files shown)


internal/lsp/server_test.go
```diff
@@ -336,30 +336,29 @@ capabilities:
...
 		select {
 		case <-timeout.C:
 			t.Fatalf("timed out waiting for builtins map to be updated")
-		default:
+		case <-ticker.C:
 			bis := rego.GetBuiltins()
 
 			// Search for a builtin we know is only in the EOPA capabilities.
-			if _, ok := bis["startswith"]; ok {
+			if _, ok := bis["neo4j.query"]; ok {
```
charlieegan3, line 357: This is the main update.

internal/lsp/server_test.go
```diff
@@ -336,30 +336,29 @@ capabilities:
...
-	// NOTE(charles): the configuration is updated asynchronously from the
-	// thread the test is running in. This check prevents a race condition
-	// where the completion test runs before rego.builtIns is updated which
-	// was causing flaky tests in CI.
+	// manually inspect the server's list of builtins to ensure that the EOPA
+	// capabilities were loaded correctly.
 	timeout = time.NewTimer(defaultTimeout)
 	defer timeout.Stop()
 
+	ticker := time.NewTicker(100 * time.Millisecond)
```
charlieegan3, line 344: Using a ticker is just for consistency with the below check

anderseknert (approved):

charlieegan3:
Hmm, TestGetInlayHintsAstCall... I hadn't seen that in a while... Will merge for now and see about this one tomorrow.

anderseknert:
Haven't heard that name in years!

PR #1131, opened 2024-09-24
Rule: `missing-metadata`

anderseknert:
This rule is not for everyone, and hence in the `custom` category.

For a public projects, libraries and reusable policies, this should however be considered.

This uncovered a
[bug in OPA](https://github.com/open-policy-agent/opa/issues/7050), which has now been addressed. In order to make progress on this we're now depending on OPA `main`. This should be changed as soon as OPA v0.69.0 is released.

Some unrelated fixes in this PR, fixing annoyances as I worked on it:
- add Regal specific attributes (`input.regal.*`) to the input created by the `eval:use-as-input` directive
- Add another metadata snippet to allow creating a `# METADATA` block with only description and no title. Tbh, I'm still not sure what to use `title` for, while `description` is pretty much mandatory.
- Set `ignore-if-sub-attribute: true` on default config for `prefer-some-in-iteration` rule. This was documented to be the case already, but seemed to have gone missing at some point.
- Add pprof option to language-server command, and pprof handlers to web server


charlieegan3 (approved):
This is great, no comments. pprof is a nice addition :)

PR #1132, opened 2024-09-24
Use betteralign for struct alignment

anderseknert:
Saving us some bytes ;) A few things had to be fixed, but nothing I'd consider breaking.

Not too important though, so happy to close if someone else is concerned :P


pkg/reporter/reporter_test.go
```diff
@@ -255,6 +255,16 @@ func TestJSONReporterPublish(t *testing.T) {
 	expect := `{
   "violations": [
     {
+      "location": {
```
anderseknert, line 258: If we don't like the JSON output changed, we can easily create a custom serializer with jsoniter

charlieegan3 (approved):
LGTM, more concerned about keeping these orders correct. Is this something that we can lint for?

anderseknert:
@charlieegan3 there used to be an `malign` linter to do this, but it seems dead. Frankly though, for anything that isn't in an absolute hot path where millions of objects are created... this is mostly just of academic interest / benefit. I would not want to have external contributor PRs fail for this, for example. If we run this check ourselves once a year or so, that's fine with me :)

PR #1135, opened 2024-09-25
Completions: don't suggest loop vars as locals on same line

### LINKED ISSUE #1124: Locals provider should not suggest iteration over `some` bindings in the same `some .. in`
The variables bound in `some .. in` should not be suggested at the location of the item being looped over.

Screenshot 2024-09-22 at 13 06 21

anderseknert:
Fixes #1124


charlieegan3 (approved):
👏

PR #1136, opened 2024-09-25
bundle: Load bundle once

charlieegan3:
Previously, loading of the bundle happened many times, each time it takes around 20ms. This improves the performance of the language server by loading the bundle once and reusing it since the bundle is never changed once a binary is running.

before

https://github.com/user-attachments/assets/7a23a92c-e489-436d-96c7-e67b9ce9b91d

after

https://github.com/user-attachments/assets/ca1d2c5f-b043-44b0-abd1-aea3a4909ecf

### DIFF
--- a/bundle/bundle.go
+++ b/bundle/bundle.go
@@ -2,6 +2,11 @@ package bundle
 
 import (
 	"embed"
+	"sync"
+
+	"github.com/open-policy-agent/opa/bundle"
+
+	rio "github.com/styrainc/regal/internal/io"
 )
 
 // Bundle FS will include the tests as well, but since that has negligible impact on the size of the binary,
@@ -9,3 +14,18 @@ import (
 //
 //go:embed *
 var Bundle embed.FS
+
+//nolint:gochecknoglobals
+var (
+	loadedBundle     bundle.Bundle
+	loadedBundleOnce sync.Once
+)
+
+// LoadedBundle returns the loaded Regal bundle for this version of Regal.
+func LoadedBundle() *bundle.Bundle {
+	loadedBundleOnce.Do(func() {
+		loadedBundle = rio.MustLoadRegalBundleFS(Bundle)
+	})
+
+	return &loadedBundle
+}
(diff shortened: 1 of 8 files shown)


pkg/linter/linter.go
```diff
@@ -102,8 +100,8 @@ func (l Linter) WithInputModules(input *rules.Input) Linter {
 }
 
 // WithAddedBundle adds a bundle of rules and data to include in evaluation.
-func (l Linter) WithAddedBundle(b bundle.Bundle) Linter {
-	l.ruleBundles = append(l.ruleBundles, &b)
+func (l Linter) WithAddedBundle(b *bundle.Bundle) Linter {
```
charlieegan3, line 103: This is an API change. We can do it without this, but it feels right to do it given how this function is used.

charlieegan3: This is not used in the Rego playground or in any public code on https://sourcegraph.com/search?q=context:global+WithAddedBundle&patternType=keyword&sm=0

bundle/bundle.go
```diff
@@ -2,10 +2,30 @@ package bundle
...
 //go:embed *
 var Bundle embed.FS
+
+//nolint:gochecknoglobals
+var (
+	loadedBundle     bundle.Bundle
+	loadedBundleOnce sync.Once
+)
+
+// LoadedBundle returns the loaded Regal bundle for this version of Regal.
```
srenatus, line 24: I take it you'd like to keep it being lazily loaded? Because if it was going to be used anyways, often times sync.Once etc can be replaced with a package var, like
```go
var loadedBundle = io.MustLoadRegalBundleFS(Bundle)
```

charlieegan3: Perhaps we should just do this 👍 I was remembering a discussion about how inits in the ls implementation were slowing down the linting cmd, but that's not the case here.

charlieegan3: Fixed in 9c2e409614028150e2ac19835d89466fccc006a1

srenatus (approved):

PR #1143, opened 2024-09-26
Fix `detached-metadata` issues

### LINKED ISSUE #1138: False positive in `detached-metadata` when both package and subpackages scope in use
This is reported as detached, but is valid, as the package can have both `package` scoped annotations and `subpackages`.
```rego
# METADATA
# title: foo

# METADATA
# scope: subpackages
# title: bar
package p
```

anderseknert:
Build a more robust system for determining this regardless of scope. Lean in on the OPA parser failing anyway when scope is invalid.

Fixes #1138

### DIFF
--- a/bundle/regal/rules/style/detached-metadata/detached_metadata.rego
+++ b/bundle/regal/rules/style/detached-metadata/detached_metadata.rego
@@ -9,7 +9,7 @@ import data.regal.result
 import data.regal.util
 
 report contains violation if {
-	some block in ast.comments.blocks
+	some i, block in ast.comments.blocks
 
 	startswith(trim_space(block[0].text), "METADATA")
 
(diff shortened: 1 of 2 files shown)


bundle/regal/rules/style/detached-metadata/detached_metadata.rego
```diff
@@ -29,3 +28,16 @@ _annotation_at_row(row) := annotation if {
 
 	util.to_location_object(annotation.location).row == row
 }
+
+_allow_detached(last_row, i, blocks, lines) if {
```
charlieegan3, line 32: I understand this now, but a comment might be helpful in a few months time! 😅

charlieegan3 (approved):

bundle/regal/rules/style/detached-metadata/detached_metadata.rego
anderseknert: Sure! :) I'll add one 👍

Issue #1133, opened 2024-09-24
Redesign handling of aggregate rules in language server


anderseknert:
Working on #1131 I hit some severe performance issues around aggregate rule violations, frequently leading to the language server to either hang, or heavily tax the CPU. The reason for this is that aggregate violations currently result in the entire workspace to be linted. Any change later in that package will again lead to the entire workspace to be linted, and so on... instead, I think we should do the following in the context of LSP (and only in that context):

1. Separate aggregation from linting. I think we do this already internally, but calling `linter.Lint()` will obviously do both.
2. On server startup, collect aggregate data from all aggregate rules, and store in the server cache.
3. On file changes, collect aggregate data only from that file, and replace existing data for that file.
4. On file changes, request linting of both normal rules and aggregate rules, using the aggregated data as input for the latter.

CC @charlieegan3 for ideas.

charlieegan3:
Yeah this sounds like a good plan to me. As I understand it, the aggregates are already set on the report, just never shown. So when we complete linting, we can stash these in the cache pretty easily.

I guess the core part that needs to be adjusted is

```go
	if len(input.FileNames) > 1 {
		aggregateReport, err := l.lintWithRegoAggregateRules(ctx, regoReport.Aggregates, regoReport.IgnoreDirectives)
		if err != nil {
			return report.Report{}, fmt.Errorf("failed to lint using Rego aggregate rules: %w", err)
		}

		finalReport.Violations = append(finalReport.Violations, aggregateReport.Violations...)
	}

```

Where we might want to have a new field in the input containing all the cached aggregates?

anderseknert:
Anything more to do here, @charlieegan3 ?

charlieegan3:
There is always more we can do, but I think we can close this one until we have a new performance issue we want to focus on.

PR #1146, opened 2024-09-27
lsp: Update LSP linting to run incrementally after file change

charlieegan3:
Fixes to https://github.com/StyraInc/regal/issues/1133

This PR is a large refactoring of the language server diagnostics loops. Changes:

- when the contents of a file with aggregate violations changes, this no longer triggers a full workspace lint
- when the contents of any file changes, the aggregate rule state is now cached on the server and the results of this update, in addition to the full global state, is used to perform a linting run of all files but only linting for aggregate rules.
- new tests have been added to cover this functionality and the underlying functionality which these server changes depend on in the linter.
- server tests moved to new files since they are very long.
- increased timeouts for go test when running in go test -race https://go.dev/doc/articles/race_detector#Runtime_Overheads
- updated the cache to allow partial updating of diagnostics based on the relevant rules.

### DIFF
--- a/internal/lsp/cache/cache.go
+++ b/internal/lsp/cache/cache.go
@@ -9,6 +9,7 @@ import (
 	"github.com/open-policy-agent/opa/ast"
 
 	"github.com/styrainc/regal/internal/lsp/types"
+	"github.com/styrainc/regal/pkg/report"
 )
 
 // Cache is used to store: current file contents (which includes unsaved changes), the latest parsed modules, and
@@ -26,11 +27,14 @@ type Cache struct {
 	// modules is a map of file URI to parsed AST modules from the latest file contents value
 	modules map[string]*ast.Module
 
+	// aggregateData stores the aggregate data from evaluations for each file.
+	// This is used to cache the results of expensive evaluations and can be used
+	// to update aggregate diagostics incrementally.
+	aggregateData   map[string][]report.Aggregate
+	aggregateDataMu sync.Mutex
+
 	// diagnosticsFile is a map of file URI to diagnostics for that file
 	diagnosticsFile map[string][]types.Diagnostic
-
-	// diagnosticsAggregate is a map of file URI to aggregate diagnostics for that file
-	diagnosticsAggregate map[string][]types.Diagnostic
 
 	// diagnosticsParseErrors is a map of file URI to parse errors for that file
 	diagnosticsParseErrors map[string][]types.Diagnostic
@@ -54,8 +58,6 @@ type Cache struct {
 
 	diagnosticsFileMu sync.Mutex
 
-	diagnosticsAggregateMu sync.Mutex
-
 	diagnosticsParseMu sync.Mutex
 
 	builtinPositionsMu sync.Mutex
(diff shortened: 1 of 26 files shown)


internal/lsp/cache/cache.go
```diff
@@ -180,6 +189,53 @@ func (c *Cache) SetModule(fileURI string, module *ast.Module) {
...
+func (c *Cache) GetFileAggregates(fileURI string) ([]report.Aggregate, bool) {
+	c.aggregateDataMu.Lock()
+	defer c.aggregateDataMu.Unlock()
+
+	data, ok := c.aggregateData[fileURI]
+	if !ok {
+		return nil, false
+	}
+
+	return data, true
```
srenatus, line 220: 💭 Is that significantly different from this?
```go
return c.aggregateData[fileURI]
```

charlieegan3: Erm, no! Will fix, though there is plenty more to address here too 😅

charlieegan3: Hmm, that doesn't seem to work... 

```
internal/lsp/cache/cache.go:215:9: not enough return values
        have ([]report.Aggregate)
        want ([]report.Aggregate, bool)

```

charlieegan3: But this is fine:

```
	data, ok := c.aggregateData[fileURI]

	return data, ok
```

srenatus: Ah, right, it's not using the two-return value magic from simple returns. Well anything is fine, just nitpicking. I wasn't even completely sure if I had sent that comment or abandoned it 🙈

charlieegan3: all good! this PR has a way to go however, still!

internal/lsp/server.go
```diff
@@ -122,12 +122,23 @@ type LanguageServer struct {
...
 	URI    string
 }
 
+// lintWorkspaceJob is sent to lintWorkspaceJobs when a full workspace
+// diagnostic update is needed.
+type lintWorkspaceJob struct {
+	Reason string
+	// OverwriteAggregates for a workspace is only run once at start up. All
+	// later updates to aggregate state is made as files are changed.
+	OverwriteAggregates bool
```
anderseknert, line 138: 👍

anderseknert (approved):
Code looks good to me! Only complaint would be all the changes not directly related to solving the issue, but they are also *good* changes, so why not! Might have been better off in separate PRs though :)

I'm excited to try this out!

anderseknert:
I've tried this now and it's a huge improvement 👏 No more lag or even freezes, but most things just show up right away, and the aggregate violations a little later. This is great!

charlieegan3:
> aggregate violations a little later

Yeah, I actually think that we are in a position to correct that, but I'd need to do some more testing. This is because for a per-file run, only the non agg rules are run to improve performance. Since we know that a full workspace agg rule run will take place immediately after with the updated aggregate state.

I think the solution is to provide all aggregate data but for the current file, and then run all rules in the per file lint run. It should be simple enough, but something I think might approached in another PR....

PR #1150, opened 2024-09-28
fix: fix var name for commit hash used for `version` subcommand

rinx:
I found that `regal version` subcommand does not show commit hash.

```bash
❯ ./regal version             
Version:    0.27.0
Go Version: go1.22.5
Platform:   linux/amd64
Commit:     unknown
Timestamp:  2024-09-17T15:00:44Z
```

This is because the build time information is passed to `pkg/version.Commit` variable, but it does not exist there.
https://github.com/StyraInc/regal/blob/e205cd9e4a9f27c52d7d0a6e7ac250c12113def8/.goreleaser.yaml#L24

I fixed this problem by renaming `pkg/version.Vcs` to `pkg/version.Commit`.

```bash
❯ go build -ldflags='-X github.com/styrainc/regal/pkg/version.Commit=my-commit'

❯ ./regal version
Version:    unknown
Go Version: go1.22.7
Platform:   linux/amd64
Commit:     my-commit
Timestamp:  unknown
Hostname:   unknown

```

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -11,7 +11,7 @@ var Version = ""
 
 // Additional Regal metadata to be injected at build time.
 var (
-	Vcs       = ""
+	Commit    = ""
 	Timestamp = ""
 	Hostname  = ""
 )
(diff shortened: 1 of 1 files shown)


srenatus (approved):
Thanks

PR #1153, opened 2024-10-01
Adapt to new location format

anderseknert:
Wow, this took quite some time to finish! But looking at the end result, I'd say it was worth it. Having Roast represent locations as `row:col:end_row:end_col` means quite a reduction in AST size, and Regal's own JSON AST goes from 12 Mb to 10 Mb (in OPA AST that is 24 Mb for comparison).

Running `regal lint bundle` is also ~100 ms faster, and improvement of about 5%. Nice!

But this wasn't only about performance. Having end locations available for any node means we can now easily show e.g. violations with a precise location in LSP editors, and won't have to default to end-of-line, as we previously did for most rules. This also meant that we were able to remove all code in Regal that tried to calculate end position based on the text attribute. Nice!

NOTE: this should be merged after https://github.com/anderseknert/roast/pull/11 so that we can depend on a tagged roast version.

### DIFF
--- a/bundle/regal/ast/ast_test.rego
+++ b/bundle/regal/ast/ast_test.rego
@@ -133,17 +133,17 @@ allow := true
 `
 
 	module := regal.parse_module("p.rego", policy)
-	blocks := ast.comment_blocks(module.comments)
+	blocks := ast.comment_blocks(module.comments) with input as module
 	blocks == [
 		[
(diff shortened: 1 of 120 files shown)


mem.pprof
charlieegan3, line 1: Presume this was not meant to be committed.

charlieegan3 (approved):
Great work! this is a really nice improvement

PR #1156, opened 2024-10-01
internal/lsp/hover: "fix" codeql finding

srenatus:
> Incorrect conversion of an unsigned 64-bit integer from to a lower bit size type uint without an upper bound check.


charlieegan3 (approved):
Thanks for this!

PR #1159, opened 2024-10-01
Fix some minor issues reported by IntelliJ

anderseknert:
Would be good to know why golangci-lint fails to report these, as some were fairly obvious. But oh well!

And

### DIFF
--- a/pkg/fixer/reporter.go
+++ b/pkg/fixer/reporter.go
@@ -50,9 +50,7 @@ func (r *PrettyReporter) ReportConflicts(fixReport *Report) error {
 	if len(roots) > 0 {
 		fmt.Fprintln(r.outputWriter, "Source file conflicts:")
 
-		i := 0
-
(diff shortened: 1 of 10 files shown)


pkg/fixer/reporter.go
```diff
@@ -50,9 +50,7 @@ func (r *PrettyReporter) ReportConflicts(fixReport *Report) error {
 	if len(roots) > 0 {
 		fmt.Fprintln(r.outputWriter, "Source file conflicts:")
 
-		i := 0
-
-		for _, rootKey := range roots {
```
srenatus, line 55: Looks like this just didn't work before? `i` was never increased...

srenatus (approved):

pkg/fixer/reporter.go
anderseknert: Indeed. Not too disastrous though as it was just for printing an extra newline between groups.

Issue #1145, opened 2024-09-26
Regal LSP should support import aliases


charlesdaniels:
Consider the following example:

```rego
# foo/bar/bar.rego
package foo.bar

hello := "world"

# a/b/b.rego
package a.b
import data.foo.bar as baz

allow if {
    baz.h # tab complete here
}
```

Currently, this will not work, as the LSP is not import alias aware. However, right now if the `as baz` was removed from the import statement, then `bar.h` would correctly complete to `bar.hello`. 

It would be desirable for Regal to also support import aliases, in addition to normal imports which are already supported.

## Background

I have investigated a little in terms of what would be required to support this, and my findings are outlined below.

`bundle/regal/lsp/completion/providers/rulerefs/rulerefs.rego` is where the relevant code lives. Currently, the list of imports for the current file is determined by `_current_file_imports`. This line:

```rego
	some imp in _parsed_current_file.imports
```

Will iterate over values that look similar to these:

```json
{"location": "3:1:aW1wb3J0", "path": {"location": "3:8:cmVnby52MQ==", "type": "ref", "value": [{"location": "3:8:cmVnbw==", "type": "var", "value": "rego"}, {"location": "3:13:djE=", "type": "string", "value": "v1"}]}}
```
```json
{"alias": "s", "location": "4:1:aW1wb3J0", "path": {"location": "4:8:ZGF0YS5mb28uYmFyLnE=", "type": "ref", "value": [{"location": "4:8:ZGF0YQ==", "type": "var", "value": "data"}, {"location": "4:13:Zm9v", "type": "string", "value": "foo"}, {"location": "4:17:YmFy", "type": "string", "value": "bar"}, {"location": "4:21:cQ==", "type": "string", "value": "q"}]}}
```

> [!TIP]
> To be able to see print statements from this policy during testing, you can add `regoArgs = append(regoArgs, rego.EnablePrintStatements(true))` into `prepareQuery` in `internal/lsp/completions/providers/policy.go`. This method is how the samples above were obtained.

In the case of imports which have an `as` alias, a top level `alias` field will be present in the import object. A naive solution might be to simply mangle the value passed to `ast.ref_to_string(pathval)` to reflect the alias values, perhaps like this:

```rego
_current_file_imports contains ref if {
	some imp in _parsed_current_file.imports

	pathval := [
		v
		|
		some i, elem in imp.path.value ;
		v := {
			false: elem,
[...]

PR #1160, opened 2024-10-02
experiment: allow sideloading files from fs.FS

srenatus:
1. add a layer of abstraction, fs.FS, into the code path for loading files
2. expose a new package global, `virtualInput`
3. setting virtualInput will load the provided fs.FS directly, and return an input
4. all calls to NewInput (all commands incl. LSP) will merge with the "virtual input", if provided

The reason for this: in EOPA, we've got a bunch of rego helper functions, and we'd like the LSP to be aware of them for code completion. This would seem like a decent approach to _make it aware of them_.

(The actual code completion UX still depends on #1145, because that is how the utility methods are advertised to be used.)

### DIFF
--- a/pkg/rules/rules.go
+++ b/pkg/rules/rules.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"sort"
 	"sync"
+	"sync/atomic"
 
 	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/loader"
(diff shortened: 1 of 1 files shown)


pkg/rules/rules.go
```diff
@@ -79,14 +83,13 @@ func InputFromPaths(paths []string) (Input, error) {
 		go func(path string) {
 			defer wg.Done()
 
-			result, err := loader.RegoWithOpts(path, parse.ParserOptions())
```
srenatus, line 82: What this helper did for us isn't much, so I've inlined it below, added the indirection to read from fs.FS.

internal/util/util.go
```diff
@@ -22,13 +23,7 @@ func Keys[K comparable, V any](m map[K]V) []K {
...
 // Contains checks if slice contains element.
 func Contains[T comparable](s []T, e T) bool {
-	for _, v := range s {
-		if v == e {
-			return true
-		}
-	}
-
-	return false
+	return slices.Contains(s, e)
```
anderseknert, line 26: We could just remove this helper entirely since it's in stdlib now.

srenatus:
Not today 🧹

Issue #1164, opened 2024-10-02
Templating doesn't work reliably without a `.regal` directory


anderseknert:
It seems like the templating of new files imposed by `directory-structure-mismatch` doesn't work very well unless there is a `.regal` directory to pin the workspace root to. Since this is going to be the first impression many will get of Regal, it should be fixed before next release.

anderseknert:
#1141 certainly related, so might be a good idea to look at both at the same time.

Issue #1141, opened 2024-09-26
Skip templating when creating policy file in workspace root. Allow disabling templating.


anderseknert:
We currently create a "main" directory for this and move it there. This has rightly been perceived as confusing, and there's nothing in this process that provides a clue as to why this happened. We could add a comment or something like that in the template, but a simpler approach is to just skip templating in the workspace root. This will instead flag a violation, giving users a chance to learn about it, and disable it if they so want.

On the topic of disabling, we should also allow disabling templating altogether.

anderseknert:
I don't think there's much more to be done here. Will you can't disable templating separately, you can always disable the directory-package-mismatch rule.

PR #1171, opened 2024-10-03
lsp/templating: gracefully handle unknown root

charlieegan3:
This also disables templating for files in the root.

Fixes https://github.com/StyraInc/regal/issues/1164

Related to https://github.com/StyraInc/regal/issues/1141, but this needs more work

Sorry for the low res video... 

https://github.com/user-attachments/assets/18c95fce-4981-4e8b-a1c0-45418aac9573

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -986,12 +986,16 @@ func (l *LanguageServer) StartTemplateWorker(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		case job := <-l.templateFileJobs:
+			// disable the templating feature for files in the workspace root.
+			if filepath.Dir(uri.ToPath(l.clientIdentifier, job.URI)) ==
+				uri.ToPath(l.clientIdentifier, l.workspaceRootURI) {
+				continue
+			}
+
 			// determine the new contents for the file, if permitted
 			newContents, err := l.templateContentsForFile(job.URI)
 			if err != nil {
-				if !errors.Is(err, &templatingInRootError{}) {
-					l.logError(fmt.Errorf("failed to template new file: %w", err))
-				}
+				l.logError(fmt.Errorf("failed to template new file: %w", err))
 
 				continue
 			}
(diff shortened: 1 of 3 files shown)


internal/lsp/server.go
```diff
@@ -356,10 +356,10 @@ func (l *LanguageServer) StartDiagnosticsWorker(ctx context.Context) {
...
 				return
 			case job := <-workspaceLintRuns:
-				// if there are no files in the cache, then there is no need to
-				// run the aggregate report. This can happen if the server is
-				// very slow to start up.
-				if len(l.cache.GetAllFiles()) == 0 {
+				// if there are no parsed modules in the cache, then there is
+				// no need to run the aggregate report. This can happen if the
+				// server is very slow to start up.
+				if len(l.cache.GetAllModules()) == 0 {
```
charlieegan3, line 362: This is unrelated, but is something that I noticed when working with empty files, a better check here is to test if there are modules rather than if there are files since empty files have no parse.

internal/lsp/server_template_test.go
```diff
@@ -124,6 +116,73 @@ func TestTemplateContentsForFile(t *testing.T) {
 	}
 }
 
+func TestTemplateContentsForFileInWorkspaceRoot(t *testing.T) {
```
charlieegan3, line 119: Root is not allowed and is an error which can be silently ignored if the caller understands it.

internal/lsp/server_template_test.go
```diff
@@ -124,6 +116,73 @@ func TestTemplateContentsForFile(t *testing.T) {
...
+	if err == nil {
+		t.Fatalf("expected error")
+	}
+
+	if !errors.Is(err, &templatingInRootError{}) {
+		t.Fatalf("expected error to be templatingInRootError, got %T", err)
+	}
+}
+
+func TestTemplateContentsForFileWithUnknownRoot(t *testing.T) {
```
charlieegan3, line 153: When the root is unknown, but there is a path, we can attempt to use the server's root URI as a root instead.

internal/lsp/server.go
```diff
@@ -989,7 +989,9 @@ func (l *LanguageServer) StartTemplateWorker(ctx context.Context) {
 			// determine the new contents for the file, if permitted
 			newContents, err := l.templateContentsForFile(job.URI)
 			if err != nil {
-				l.logError(fmt.Errorf("failed to template new file: %w", err))
+				if !errors.Is(err, &templatingInRootError{}) {
+					l.logError(fmt.Errorf("failed to template new file: %w", err))
```
anderseknert, line 993: Is it really an error unless it's something the user asked for and didn't get? If I saw this in the logs, I would probably think I did something wrong and that I should look into how to fix it.

internal/lsp/server_template_test.go
```diff
@@ -124,6 +116,73 @@ func TestTemplateContentsForFile(t *testing.T) {
...
+	td := t.TempDir()
+
+	err := os.MkdirAll(filepath.Join(td, ".regal"), 0o755)
+	if err != nil {
+		t.Fatalf("failed to create directory %s: %s", filepath.Join(td, ".regal"), err)
+	}
+
+	err = os.WriteFile(filepath.Join(td, ".regal/config.yaml"), []byte{}, 0o600)
+	if err != nil {
+		t.Fatalf("failed to create directory %s: %s", filepath.Join(td, ".regal"), err)
```
anderseknert, line 131: This is not creating a directory

anderseknert (approved):
LGTM, with one comment consider. Even if we use an error for "control flow", I don't think we'll want to log this as such

internal/lsp/server.go
charlieegan3: I've opted to not use the error for control flow here now: 0936ca7b7827ed2fc91a93a3b00bbccf7556e3be


While it's still an error, the callers of the function are now expected to test if the file is in the root and avoid calling the function accordingly.

PR #1175, opened 2024-10-05
feat: enable `regal/startDebugging` LSP method in Neovim

rinx:
Refs: neovim/nvim-lspconfig#3354

I created a pull request to nvim-lspconfig, it adds an LSP handler for `regal/startDebugging` method to regal default config.
Then, in this pull request, I'd like to enable `regal/startDebugging` method in Neovim to use `Debug` codelens feature.

Demo:
nvim-debug-codelens

I'm also planning to write an Neovim handler for `regal/showEvalResult`. I'll send a PR about it later.

---

Update:

I closed neovim/nvim-lspconfig#3354 because the handler depends on nvim-dap plugin.

Then, I updated nvim-dap-rego to bundle this handler into it.
https://github.com/rinx/nvim-dap-rego/commit/7316f3fa47266dbb28a82f77a0e1e9adc83ab4a3

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -711,8 +711,8 @@ func (l *LanguageServer) StartCommandWorker(ctx context.Context) { // nolint:mai
 				// handle this ourselves as it's a rename and not a content edit
 				fixed = false
 			case "regal.debug":
-				if l.clientIdentifier != clients.IdentifierVSCode {
-					l.logError(errors.New("regal.debug command is only supported in VSCode"))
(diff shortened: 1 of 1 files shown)


anderseknert:
Wow, that looks amazing! 😻

rinx:
In this pull request, I added Neovim to allow list of debug codelens feature.
However, I think it is better to allow users to control which feature is enabled.
I'm going to close this pull request and send another pull request to allow users to pass initialization options to enable debug codelens and select evaluate codelens output methods.

PR #1176, opened 2024-10-06
feat(lsp): add initialization options about codelenses

rinx:
Refs: #1175 

This pull request adds two options to LSP's initialization options.

- `enableDebugCodelens`: to enable/disable debug codelens
- `evalCodelensDisplayInline`: to select result output methods of evaluate codelens

In the current implementation, `Debug` codelens feature requires the client is VSCode.
However like I showed in #1175, it is possible to use this feature if the client implements specific handlers for the Regal's `regal/startDebugging` request.
In this pull request, I propose to add the options to control the Regal's behavior about codelenses.

This change will be breaking to vscode-opa, but just adding two parameters to https://github.com/open-policy-agent/vscode-opa/blob/f49ab2ec0e72516749f7416daea4a246ddfe7b6b/src/ls/clients/regal.ts#L203.
I'll send a pull request to handle this issue.

For Neovim users, nvim-dap-rego provides `regal/startDebugging` handler. rinx/nvim-dap-rego@7316f3fa47266dbb28a82f77a0e1e9adc83ab4a3

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -711,12 +711,6 @@ func (l *LanguageServer) StartCommandWorker(ctx context.Context) { // nolint:mai
 				// handle this ourselves as it's a rename and not a content edit
 				fixed = false
 			case "regal.debug":
-				if l.clientIdentifier != clients.IdentifierVSCode {
-					l.logError(errors.New("regal.debug command is only supported in VSCode"))
-
-					break
-				}
-
 				if len(params.Arguments) != 3 {
 					l.logError(fmt.Errorf("expected three arguments, got %d", len(params.Arguments)))
 
@@ -854,7 +848,8 @@ func (l *LanguageServer) StartCommandWorker(ctx context.Context) { // nolint:mai
 					target = strings.TrimPrefix(path, currentModule.Package.Path.String()+".")
 				}
 
-				if l.clientIdentifier == clients.IdentifierVSCode {
+				if l.clientInitializationOptions.EvalCodelensDisplayInline != nil &&
+					*l.clientInitializationOptions.EvalCodelensDisplayInline {
 					responseParams := map[string]any{
 						"result": result,
 						"line":   line,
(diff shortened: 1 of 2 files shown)


charlieegan3:
Hi Rinx, thanks for taking a look into this one. I was wondering if we might be able to make use of the [CodeLensWorkspaceClientCapabilities](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeLensWorkspaceClientCapabilities) for this?

If these were set and compatible, then we could use those without the need for custom option? Just a thought.

charlieegan3:
I just watched the demo in https://github.com/StyraInc/regal/pull/1175, looks amazing!

anderseknert:
Re: @charlieegan3 

> I was wondering if we might be able to make use of the [CodeLensWorkspaceClientCapabilities](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeLensWorkspaceClientCapabilities) for this?

I guess the idea here is that even if the client supports showing code lenses, both of the code lenses execute commands that aren't defined by the standard. Showing the code lenses in a client which has no way of handling e.g. a debugging is probably a worse option that not showing it at all.

Any code lens-capable client should be able to handle eval though. It would just have the output found in `output.json`? rather than printed inline, if this option wasn't set. And that seems OK to me.

internal/lsp/server.go
```diff
@@ -1631,7 +1626,25 @@ func (l *LanguageServer) handleTextDocumentCodeLens(
 		return nil, nil // return a null response, as per the spec
 	}
 
-	return rego.CodeLenses(ctx, params.TextDocument.URI, contents, module) //nolint:wrapcheck
+	ls, err := rego.CodeLenses(ctx, params.TextDocument.URI, contents, module)
+	if err != nil {
+		return nil, err
```
anderseknert, line 1631: Just use the `//nolint:wrapcheck` comment here if you don't think wrapping the error adds anything.

internal/lsp/server.go
```diff
@@ -1631,7 +1626,25 @@ func (l *LanguageServer) handleTextDocumentCodeLens(
...
+	}
+
+	if l.clientInitializationOptions.EnableDebugCodelens != nil &&
+		*l.clientInitializationOptions.EnableDebugCodelens {
+		return ls, nil
+	}
+
+	// filter out `regal.debug` codelens
+	lenses := make([]types.CodeLens, 0, len(ls))
+	for _, lens := range ls {
```
anderseknert, line 1641: The linter issue here is basically just saying that these two lines aren't immediately related, and as such shouldn't be connected. So just add a newline in between, e.g:

```go
// filter out `regal.debug` codelens
lenses := make([]types.CodeLens, 0, len(ls))
	
for _, lens := range ls {
```

anderseknert (approved):
Looks good to me! Just some linter issues to address, and then we'll need that change in the VS Code extension too as you suggested.

charlieegan3:
> And that seems OK to me.

👍 sounds good to me too.

In which case, I'd expect `EnableDebugCodelens` only to be added to the initialisation options?

anderseknert:
`evalCodelensDisplayInline` seems like a good way to differentiate clients capable of handling the eval response (i.e. inlined) vs those who don't, no?

charlieegan3:
Ahh I see, yeah true!

charlieegan3:
Hey @rinx, we're going to try and do a release this afternoon so I'm going to make a few adjustments to make the Go linter happy and get this in. I will keep your commit here 🙂

charlieegan3:
0.18.0 of vs code is out now, https://github.com/open-policy-agent/vscode-opa/blob/main/CHANGELOG.md#0180

So will proceed with the release.

rinx:
Sorry for belated response. I'm happy to see it is released!
Thank you for reviewing and releasing it. :smile:

PR #1188, opened 2024-10-08
Lsp logging

charlieegan3:
We have been having a hard time debugging test suite issues like this one: https://github.com/StyraInc/regal/actions/runs/11233013451/job/31225814604 where the order of events doesn't appear to follow the order expected based on the implementation.

This PR makes an improvement to the Language Server to support logging of messages at different levels.

This retains the old behaviour in a new model, but then also adds a number of debug messages for requests and diagnostics jobs to aid in debugging tests.

In future, we may wish for more log levels, but for now, I have used the three available  in the LSP TraceValues.

### DIFF
--- a/cmd/languageserver.go
+++ b/cmd/languageserver.go
@@ -10,6 +10,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/styrainc/regal/internal/lsp"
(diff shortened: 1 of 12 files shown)


anderseknert (approved):
Nice!

PR #1191, opened 2024-10-09
Add UNIwise company to adopters.md

Graloth:
I work for [UNIwise](https://uniwise.eu/) and we use Regal for all our OPA policies, and would like to show our support for it here.


anderseknert (approved):
Thank you! 😃

PR #1203, opened 2024-10-15
io: Address path input.json separator issue

charlieegan3:
When searching for an input.json file to use, the existing code will use the path library, rather than filepath which is separator aware.

This causes issues on Windows where the path separator is \ and not /.


anderseknert (approved):
Good find!

PR #1206, opened 2024-10-15
Make `prefer-snake-case` check package name

### LINKED ISSUE #1205: Package name not considered by `prefer-snake-case` rule
I expected this to be a `prefer-snake-case` violation, but apparently it's not.

```
➜ echo "package fooBar" | regal lint -
Rule:         	use-rego-v1
Description:  	Use `import rego.v1`
Category:     	imports
Location:     	stdin:1:1
Text:         	package fooBar
Documentation:	https://docs.styra.com/regal/rules/imports/use-rego-v1

1 file linted. 1 violation found. 2 rules skipped:
- directory-package-mismatch: disabled when filename is unknown
- file-missing-test-suffix: disabled when filename is unknown
```

anderseknert:
Previously, only rules and vars were checked.

Fixes #1205

### DIFF
--- a/bundle/regal/rules/style/prefer-snake-case/prefer_snake_case.rego
+++ b/bundle/regal/rules/style/prefer-snake-case/prefer_snake_case.rego
@@ -9,8 +9,17 @@ import data.regal.result
 import data.regal.util
 
 report contains violation if {
+	some part in input["package"].path
+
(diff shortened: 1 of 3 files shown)


charlieegan3 (approved):
👏

PR #1207, opened 2024-10-16
Handle `file://` URLs in exclusion policy

### LINKED ISSUE #1204: LSP: ignore config doesn't fully work for some rules
I'm not yet sure what the cause is or why it doesn't happen consistently, but for some rules, like [missing-metadata](https://docs.styra.com/regal/rules/custom/missing-metadata), a configuration like this:

```yaml
rules:
  custom:
    missing-metadata:
      level: error
      ignore:
        files:
          - "*_test.rego"
          - "bundle/example/**"
```

Ignores test files as expected, but **not** files under `bundle/example`. This works fine for `regal lint`, so there seems to be some inconsistency as to how the language server parses the configuration. Obviously the ideal would be if we had only one way of doing this, but either way, they should at least work the same.

anderseknert:
This is not an ideal solution, as we should not be sending URLs to the linter in the first place. But as fixing that is a much broader issue, we'll have to settle for a hack in the meantime.

Fixes #1204

### DIFF
--- a/bundle/regal/config/exclusion.rego
+++ b/bundle/regal/config/exclusion.rego
@@ -3,46 +3,61 @@ package regal.config
 import rego.v1
 
 # METADATA
-# description: determines if file should be excluded by the given rule
+# description: |
+#   determines if file should be excluded, either because it's globally
+#   ignored, or because the specific rule configuration excludes it
 excluded_file(category, title, file) if {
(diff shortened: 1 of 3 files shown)


charlieegan3 (approved):
LGTM, it's a shame that the rego code needs to know about the URIs. Thanks for adding a test.

PR #1210, opened 2024-10-16
Use filepath.WalkDir instead of filepath.Walk

anderseknert:
This has the benefit of not having to stat all files along the way.

Ideally, we'd be able to build the data bundle in the same function more efficiently too, but that's a more complex operation.


charlieegan3 (approved):
LGTM

Issue #1211, opened 2024-10-18
lint: Suggestion - issues with error levels in output


geirs73:
Context: We have a build pipeline where we check our bundle in many ways before accepting it for upload to our Styra systems.

Problem: Pretty-formatted output on console is colored red for errors, but when running with `--no-colors` they are indistinguishable from warnings. Of course, that is the way we see it in our build server logs.  

Mitigation: My workaround is to use `--format github` instead, since level is visible there. 

Suggestion: Maybe the combination `--format pretty --no-colors` should have some extra information outputted, either another line with level, or a bracket after the description, [warning] or [error]

Suggestion: Another possibility is to filter output based on level, by specifying `--level (warning|error)`.  That way, you can prioritize to only see errors. Alternatively, it could be `--level (ignore|warning|error)` so that you can see how many ignored rules are affected without having to cheat with different config files. However, when developing rules, we work in vscode-opa extension, so I'm not sure this is needed, as they show up in the Problems panel view. Yes, it would also solve my build server problem, but I would like to see both warnings and errors there.

Regal version info:
```text
Version:    0.28.0
Go Version: go1.22.5
Platform:   windows/amd64
Commit:     9503967b3cd1d3725d078b04dd55ee9ec807b3f3
Timestamp:  2024-10-07T15:17:16Z
Hostname:   github.actions.local
````

charlieegan3:
Hey @geirs73, lmk if the impl at https://github.com/StyraInc/regal/pull/1213 looks suitable for your use case!

geirs73:
Thanks @charlieegan3 ! This looks great. Just tested it on a locally built version.

charlieegan3:
That's great, glad to hear it! we should be getting a release out within the next week or so.

PR #1213, opened 2024-10-21
lint/rpt: Handle no-color ttys

charlieegan3:
This fixes https://github.com/StyraInc/regal/issues/1211 by supporting the NO_COLOR environment variable and checking if the output is a tty before using color.

The color.NoColor global is used as described here: https://github.com/fatih/color?tab=readme-ov-file#disableenable-color

### DIFF
--- a/pkg/reporter/reporter.go
+++ b/pkg/reporter/reporter.go
@@ -238,6 +238,13 @@ func buildPrettyViolationsTable(violations []report.Violation) string {
 		}
 
 		table.Append([]string{yellow("Rule:"), violation.Title})
+
+		// if there is no support for color, then we show the level in addition
+		// so that the level of the violation is still clear
+		if color.NoColor {
+			table.Append([]string{"Level:", violation.Level})
+		}
+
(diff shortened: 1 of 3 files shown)


anderseknert (approved):

pkg/reporter/reporter.go
```diff
@@ -238,6 +238,13 @@ func buildPrettyViolationsTable(violations []report.Violation) string {
 		}
 
 		table.Append([]string{yellow("Rule:"), violation.Title})
+
+		// if there is no support for color, then we show the level in addition
+		// so that the level of the violation is still clear
+		if color.NoColor {
+			table.Append([]string{yellow("Level:"), violation.Level})
```
srenatus, line 245: Feels a little odd to use `yellow()` here, doesn't it? 🤔

charlieegan3: 😬 hmm indeed!

charlieegan3: Have updated this now

srenatus (approved):
Thanks!

PR #1215, opened 2024-10-21
Rule: `defer-assignment`

### LINKED ISSUE #1147: Rule: `defer-assignment`
Example somewhat exaggerated, but hopefully well suited for demonstration.

**Avoid**
```rego
allow if {
    resp := http.send({"method": "get", "url": "http://example.org"})
    
    "developer" in input.user.roles # if this fails, `resp` assignment was for nothing
    resp.status.code == 200
}
``` 

**Prefer**
```rego
allow if {
    "developer" in input.user.roles

    resp := http.send({"method": "get", "url": "http://example.org"})
    resp.status.code == 200
}
``` 

The basic idea would be to recommend moving assignments next to where they're used, as having checks in between where the value of assignment isn't used may render the (potentially costly) assignment redundant.

anderseknert:
Finally a new rule in the `performance` category! And this is quite a fun one. I was curious to see how many violations this would find in the Regal codebase, if any. It did find a few, and good ones to fix!

In order to keep it simple, this first implementation is perhaps a bit too cautious about what it considers deferrable. Check the implementation to see what exactly, but pretty much only assignments where the next line is a not an assignment, a potential loop, or something else that *might* make deferring the assignment undesirable.

Fixes #1147

### DIFF
--- a/bundle/regal/rules/performance/defer-assignment/defer_assignment.rego
+++ b/bundle/regal/rules/performance/defer-assignment/defer_assignment.rego
@@ -1,0 +1,97 @@
+# METADATA
+# description: Assignment can be deferred
+package regal.rules.performance["defer-assignment"]
+
+import rego.v1
+
+import data.regal.ast
+import data.regal.result
+
+report contains violation if {
+	some i, rule in input.rules
+	some j, expr in rule.body
+
+	[var, rhs] := ast.assignment_terms(expr)
+
+	not _ref_with_vars(rhs)
+
+	# for now, only simple var assignment counts.. later we can
+	# consider checking the contents of arrays here
+	var.type == "var"
+
+	next := rule.body[j + 1]
+
+	not ast.is_assignment(next)
+	not ast.var_in_head(rule.head, var.value)
+	not _var_used_in_expression(var, next)
+	not _iteration_expression(next)
+	not _print_call(next)
+
+	violation := result.fail(rego.metadata.chain(), result.location(expr))
+}
+
+_ref_with_vars(node) if {
+	node.type == "ref"
+
+	some i, term in node.value
+
+	i > 0
+	term.type == "var"
+}
+
+_var_used_in_expression(var, expr) if {
+	not expr.terms.symbols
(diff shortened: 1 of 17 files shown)


bundle/regal/rules/performance/defer-assignment/defer_assignment.rego
```diff
@@ -0,0 +1,97 @@
...
+
+report contains violation if {
+	some i, rule in input.rules
+	some j, expr in rule.body
+
+	[var, rhs] := ast.assignment_terms(expr)
+
+	not _ref_with_vars(rhs)
+
+	# for now, only simple var assignment counts.. later we can
```
charlieegan3, line 18:
```suggestion
	# for now, only simple var assignment counts, later we can
```

docs/rules/performance/defer-assignment.md
```diff
@@ -0,0 +1,79 @@
...
+
+    # more done with response here
+}
+```
+
+## Rationale
+
+Assignments are normally cheap, but certainly not always. If the right-hand side of an assignment is expensive,
+deferring the assignment to where it's needed can save a considerable amount of time. Even for less expensive
+assignments, code tends to be more readable when assignments are placed close to where they're used.
```
charlieegan3, line 49: > code tends to be more readable when assignments are placed close to where they're used.

I think this is the number one reason, cheap/expensive is the secondary reason for this rule since it does not always apply / is more complicated.

Something like this might work:

> Placing assignments close to where they’re used improves code readability. Following this rule is also particularly important when the right-hand side of an assignment is expensive, such as making an HTTP call, as deferring the assignment can also save significant time.

charlieegan3 (approved):
This is a great rule 👏

docs/rules/performance/defer-assignment.md
anderseknert: This rule is in the performance category, which is why it's emphasized here. I think there are good examples of both really, and there are of course examples of both where location doesn't matter much.

PR #1217, opened 2024-10-22
lsp/eval: Load capabilities and config into eval

charlieegan3:
When developing Regal rules it's useful to be able to access the server's capabilities and configuration. This change adds these to the rego evaluation context.

### DIFF
--- a/internal/lsp/eval.go
+++ b/internal/lsp/eval.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
 	"strings"
 
 	"github.com/anderseknert/roast/pkg/encoding"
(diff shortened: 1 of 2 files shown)


internal/lsp/eval.go
```diff
@@ -138,21 +145,57 @@ func (l *LanguageServer) EvalWorkspacePath(
...
+		},
+		Data: map[string]any{
+			"internal": map[string]any{
+				"combined_config": config.ToMap(evalConfig),
+				"capabilities":    caps,
+			},
+		},
+	}
+
+	fmt.Fprintln(os.Stderr, "Loaded internal bundle")
```
anderseknert, line 194: I guess this was left from testing :) I can have it removed in the next PR though.

anderseknert (approved):
Thanks a lot! 😃

internal/lsp/eval.go
charlieegan3: oops, indeed! thanks!

PR #1219, opened 2024-10-23
Rule: `walk-no-path`

### LINKED ISSUE #1154: Rule: `walk-no-path`
**Avoid**
```rego
allow if {
    # path is bound but never referenced
    walk(users, [path, value])

    value.type == "user"
    value.is_admin == true
}
```

**Prefer**
```rego
allow if {
    # wildcard used for path
    walk(users, [_, value])

    value.type == "user"
    value.is_admin == true
}
```

The latter is optimized in the evaluation, as the path does not need to be constructed in each iteration. This saves a ton of allocations when traversing large nodes.

anderseknert:
Suggest replacing the path variable with a wildcard where possible, as the walk built-in is able to optimize such calls. More info in the docs included in the PR.

Also:
- Add `ast.is_wildcard` convenience function and use it consistently

Fixes #1154

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,6 +254,7 @@ The following rules are currently available:
 | imports     | [unresolved-import](https://docs.styra.com/regal/rules/imports/unresolved-import)                       | Unresolved import                                         |
 | imports     | [use-rego-v1](https://docs.styra.com/regal/rules/imports/use-rego-v1)                                   | Use `import rego.v1`                                      |
 | performance | [defer-assignment](https://docs.styra.com/regal/rules/performance/defer-assignment)                     | Assignment can be deferred                                |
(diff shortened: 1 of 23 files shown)


charlieegan3 (approved):
This is great! No comments from me on this one.

charlieegan3:
The only thing I wondered about was showing the results of some testing to motivate this rule, e.g. `for a deeply nested walk of 1mb of json this can make the walk 50% faster` or similar. It might be good for the other performance category rules too.

anderseknert:
@charlieegan3 added!

> More concretely, `walk`ing without generating the path array cuts down evaluation time by about 33%, and reduces the
number of allocations by about 40%.

PR #1227, opened 2024-10-30
lsp: Fix inconsistent processing of ignores

charlieegan3:
In the LS linting, pattern/* would not work when it would in the CLI. This was related to how abs paths were resolved and is fundamentally caused by and issue with URIs being used as file names in the linter.

Now, all files from the LS lint are ignored based on their path with the workspace root URI trimmed.

https://github.com/user-attachments/assets/30118754-61d5-4265-b68b-9fac99c204d9

### DIFF
--- a/bundle/regal/config/exclusion_test.rego
+++ b/bundle/regal/config/exclusion_test.rego
@@ -91,21 +91,6 @@ test_excluded_file_cli_overrides_config if {
 		with data.eval.params as object.union(params, {"ignore_files": [""]})
 }
 
-test_excluded_file_using_uri if {
-	conf := {"rules": {"test": {"rule": {
-		"level": "error",
-		"ignore": {"files": ["foo/**/p.rego"]},
-	}}}}
-
-	config.excluded_file("test", "rule", "file:///workspace/folder/foo/bar/p.rego") with config.merged_config as conf
-}
-
-test_not_excluded_file_using_uri if {
-	conf := {"rules": {"test": {"rule": {"level": "error"}}}}
-
-	not config.excluded_file("test", "rule", "file:///workspace/folder/foo/bar/p.rego") with config.merged_config as conf
-}
-
 test_trailing_slash if {
 	config._trailing_slash("foo/**/bar") == {"foo/**/bar", "foo/**/bar/**"} # regal ignore:leaked-internal-reference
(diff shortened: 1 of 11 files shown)


bundle/regal/config/exclusion_test.rego
```diff
@@ -91,21 +91,6 @@ test_excluded_file_cli_overrides_config if {
...
-	}}}}
-
-	config.excluded_file("test", "rule", "file:///workspace/folder/foo/bar/p.rego") with config.merged_config as conf
-}
-
-test_not_excluded_file_using_uri if {
-	conf := {"rules": {"test": {"rule": {"level": "error"}}}}
-
-	not config.excluded_file("test", "rule", "file:///workspace/folder/foo/bar/p.rego") with config.merged_config as conf
-}
```
anderseknert, line 107: I did not expect these tests to have to be removed, but perhaps moved to the tests of main. Wr could certainly use some coverage here.

anderseknert (approved):
Fixing this is great!  👏 

Passing URIs to a function that at least judging by its name expects a directory.. doesn't feel great. But your hack for dealing with that is much better than mine!

charlieegan3:
> Passing URIs to a function that at least judging by its name expects a directory

I would actually like to rename the function as it's being used with URIs elsewhere already and is even mentioned in the comment for the function.

I will take another pass at this today.

bundle/regal/config/exclusion_test.rego
charlieegan3: I think that 72cc35b7af3b995c09e0f08a5bd432f9aca302d2 is a good improvement here. I've brought back some tests, though they are now in main.rego as config is not meant to know about URIs any more.

I have also renamed the rootDir refs to pathPrefix to make it clearer that we're supporting URI or path prefixes with these.

bundle/regal/main/main_test.rego
```diff
@@ -191,6 +191,78 @@ test_exclude_files_rule_config if {
 	count(report) == 0
 }
 
+test_exclude_files_rule_config_with_path_prefix_relative_name if {
+	cfg := {"rules": {"testing": {"test": {"level": "error"}}}}
+
+	# regal ignore:leaked-internal-reference
```
anderseknert, line 197: I guess you won't be needing this if you rebase from main? :)

anderseknert (approved):

PR #1228, opened 2024-10-30
rule: Disable leaked_internal_reference for test files

charlieegan3:
Option added to turn back on if needed

### DIFF
--- a/docs/rules/bugs/leaked-internal-reference.md
+++ b/docs/rules/bugs/leaked-internal-reference.md
@@ -47,10 +47,11 @@ mean that the benefits listed above won't apply to your project.
 
 ## Exceptions
 
-In test files and packages, it can be useful to reference internal rules and
-functions to ensure good test coverage. If you want to run this rule for test
-files too, you can set `include-test-files: true` in the configuration for this
-rule.
+This rule is not enabled by default for test files. In tests, it can be useful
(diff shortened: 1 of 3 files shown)


docs/rules/bugs/leaked-internal-reference.md
```diff
@@ -44,6 +45,13 @@ Do note that if you disagree with this rule, you don't need to disable it unless
 something else. If you don't use underscore prefixes, nothing will be reported by this rule anyway. It does however
 mean that the benefits listed above won't apply to your project.
 
+## Exceptions
+
+In test files and packages, it can be useful to reference internal rules and
+functions to ensure good test coverage. If you want to run this rule for test
+files too, you can set `include-test-files: true` in the configuration for this
```
anderseknert, line 52: It isn't super clear here that this rule is disabled by default for tests. I'd put more emphasis on that, given that it's probably the only rule where that is the case.

anderseknert (approved):

docs/rules/bugs/leaked-internal-reference.md
charlieegan3: Yeah that's fair, I have updated in 1abaa961d24e6a312dc6e9a1b5d25f6da02c9a19

PR #1230, opened 2024-10-31
linter: Reuse lint prepared query between lint runs

charlieegan3:
This allows aggregating linting and standard linting to use the same prepared query. This saves around 150ms on lint runs on my laptop.

This is done by using a single entrypoint with some input 'switches' to turn things on and off depending on the linter needs at that time and to stop things being run when they aren't required in the result.

This does not make a change to the LSP integration where this could be cached even between linter instances, but sharing this PR now to check the direction etc.

I get around 3%-4% total improvement for linting the regal bundle

```sh
…/regal main ➜ hyperfine --warmup 1 --runs 10 './regal-main lint bundle' './regal-new lint bundle'
Benchmark 1: ./regal-main lint bundle
  Time (mean ± σ):      3.974 s ±  0.073 s    [User: 22.025 s, System: 0.542 s]
  Range (min … max):    3.883 s …  4.096 s    10 runs

Benchmark 2: ./regal-new lint bundle
  Time (mean ± σ):      3.847 s ±  0.108 s    [User: 21.776 s, System: 0.521 s]
  Range (min … max):    3.735 s …  4.086 s    10 runs

Summary
  ./regal-new lint bundle ran
    1.03 ± 0.03 times faster than ./regal-main lint bundle
```


anderseknert (approved):
Amazing! 👏 150 ms shaved off from `regal lint` 😍

PR #1232, opened 2024-10-31
lsp: format root directory files

charlieegan3:
This had been disabled by mistake, now files in the project root will still be formatted using the configured formatter but will still not be templated.


anderseknert (approved):

anderseknert:
I didn't even know we did that :)

PR #1236, opened 2024-11-04
workflow: read OPA version from go.mod

srenatus:
Just reducing the manual work required when updating the OPA version... 🤞


srenatus:
Whenever I trust auto-complete, it's wrong.

anderseknert (approved):
Very nice!

PR #1238, opened 2024-11-05
Fetch .regal.json capabilites files for Enterprise OPA

charlieegan3:
Soon, as part of the Enterprise OPA release process, complete built-ins data will be pushed to the public repo. This will allow Regal to generate rich hover messages and show the correct links to documentation of additional built-ins.

### DIFF
--- a/build/do.rq
+++ b/build/do.rq
@@ -287,7 +287,7 @@ fetch_eopa_caps if {
 		not eopa_caps_tree[p]
 
 		# construct the URL to fetch the content from
-		r := rq.template("https://raw.githubusercontent.com/StyraInc/enterprise-opa/main/capabilities/{{.tag}}.json", {"tag": t})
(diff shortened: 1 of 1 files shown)


srenatus (approved):

build/do.rq
```diff
@@ -287,7 +287,7 @@ fetch_eopa_caps if {
 		not eopa_caps_tree[p]
 
 		# construct the URL to fetch the content from
-		r := rq.template("https://raw.githubusercontent.com/StyraInc/enterprise-opa/main/capabilities/{{.tag}}.json", {"tag": t})
+		r := rq.template("https://raw.githubusercontent.com/StyraInc/enterprise-opa/main/capabilities/{{.tag}}.regal.json", {"tag": t})
```
srenatus, line 290: This will fail for old tasks, iirc

charlieegan3: It should not have, since the test to fetch an old tag is based on the file on disk which is renamed to the format as before. However, sounds like we won't be needing now!

charlieegan3:
Closing as no longer needed.

PR #1243, opened 2024-11-06
Rule: `rule-assigns-default`

### LINKED ISSUE #1018: Rule: `rule-assigns-default`
This is likely not common, but then I've said so about pretty much all violations that have turned out to be common, lol

Anyway, having a rule assign the same value as the `default` assignment should be flagged as a bug.

```rego
default allow := false

allow := false if {
	input.foo
}
```

anderseknert:
Flags rules where the default value is used as the assigned value in any other incremental definition.

Fixes #1018

(unrelated changes mostly adding some ignore directives in Go code is from me having update golangci-lint which led to a few new issues getting reported)

### DIFF
--- a/docs/rules/bugs/rule-assigns-default.md
+++ b/docs/rules/bugs/rule-assigns-default.md
@@ -1,0 +1,61 @@
+# rule-assigns-default
+
+**Summary**: Rule assigned its default value
+
+**Category**: Bugs
+
+**Avoid**
+```rego
+package policy
+
+import rego.v1
+
+default allow := false
+
+# this rule assigns the same value as the default
+# and the policy would work the same without it
+allow := false if {
+    not "admin" in input.user.roles
+}
+```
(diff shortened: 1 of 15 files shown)


docs/rules/bugs/rule-assigns-default.md
```diff
@@ -0,0 +1,61 @@
...
+
+**Prefer**
+```rego
+package policy
+
+import rego.v1
+
+default allow := false
+
+# or just `allow if {` as `true` is implicit
```
charlieegan3, line 30: I wonder if a non boolean example would be a good addition here too? E.g.

```
default cpu_shares := 0.1

cpu_shares := 0.5 if input.role == "developer"
cpu_shares := 0.1 if input.role == "intern"
```

Or something similar?

charlieegan3 (approved):

PR #1244, opened 2024-11-06
lsp/hover: Use link overrides if set

charlieegan3:
Other engines may expose custom builtins with does not on openpolicyagent.org, [see here](https://github.com/StyraInc/enterprise-opa/blob/v1.29.1/capabilities/v1.29.1.json#L5110).

This PR supports loading them from the Categories for a builtin.

Thanks @srenatus for the help on this one!

Screenshot 2024-11-06 at 14 36 14

### DIFF
--- a/internal/lsp/hover/hover_test.go
+++ b/internal/lsp/hover/hover_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	"github.com/open-policy-agent/opa/ast"
+	"github.com/open-policy-agent/opa/types"
 )
 
 func TestCreateHoverContent(t *testing.T) {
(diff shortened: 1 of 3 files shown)


internal/lsp/hover/hover_test.go
```diff
@@ -26,6 +27,18 @@ func TestCreateHoverContent(t *testing.T) {
 			ast.JSONFilter,
 			"testdata/hover/jsonfilter.md",
 		},
+		{
+			&ast.Builtin{
+				Name:        "foo.bar",
+				Description: "Description for Foo Bar",
+				Decl: types.NewFunction(
+					types.Args(),
+					types.Named("bar", types.A),
```
anderseknert, line 36: Why aren't the return value name in the rendered markdown? Would be good to add a named arg too.

charlieegan3: Rebased to include this in a3e42835e2b300999a316508b5c8e7ebab6ccba4

srenatus (approved):
Thank you!

internal/lsp/hover/hover_test.go
anderseknert: Nice, thanks!

PR #1247, opened 2024-11-08
lsp/test: Wait for aggregate state to be set

charlieegan3:
This updates the server_multi_file_test to wait for the aggregate state to be set after the server starts up. This state is required to correctly lint the authz.rego file in the following step. This was sometimes happening too late, causing the test to fail due to the state not being present at the time of linting.


srenatus (approved):
Thanks, the single test run now works for me, too.

PR #1248, opened 2024-11-08
internal/lsp/*_test.go: replace for+success pattern

srenatus:
It's no functional change, just makes the code a little more concise.

Hope you don't mind me messing with your test code like this. Originally, I had attempted to use labels, like

```go
outer:
    for {
      select {
        case ... :
          // stuff
        break outer
    }
```
but labels need to be unique within a file, and having to go with `outer1`, `outer2`, ... etc would have been awkward.

### DIFF
--- a/internal/lsp/server_multi_file_test.go
+++ b/internal/lsp/server_multi_file_test.go
@@ -74,8 +74,7 @@ ignore:
 
 	// wait for the aggregate data to be set, required for correct lint in next
 	// step
-	for {
-		var success bool
+	for success := false; !success; {
 		select {
 		default:
 			uri := "file://" + filepath.Join(tempDir, "admins.rego")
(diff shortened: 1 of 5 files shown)


internal/lsp/server_multi_file_test.go
```diff
@@ -74,32 +74,27 @@ ignore:
...
-		var success bool
+	for success := false; !success; {
 		select {
 		default:
 			uri := "file://" + filepath.Join(tempDir, "admins.rego")
 
 			aggs := ls.cache.GetFileAggregates(uri)
 			if len(aggs) > 0 {
 				success = true
+				break // don't sleep
```
srenatus, line 85: that's the only non-automatic change here

charlieegan3 (approved):
This looks great to me!

anderseknert (approved):

srenatus:
I'll fix the linter finding after lunch...

PR #1252, opened 2024-11-11
Fix false positive given arity mismatch in `inconsistent-args`

### LINKED ISSUE #1250: inconsistent-args rule does not handle wildcard variables correctly
:wave:

[NOTE: I originally opened this because I thought there was a bug with wildcard handling, but it looks like arity mismatches are the problem.]

It seems that arity mismatches cause the rule to report incorrectly.

```
matches_category(pod, category, app_labels) if {
    category != "none"
    pod.metadata.labels.app in app_labels
}

# regal reports a problem on this line
matches_category(pod, "none", _) if {
    not pod.metadata.labels.app
}

matches_category(pod, app_labels) if {
    not pod.metadata.labels.app in app_labels
}
```

anderseknert:
While a mismatched arg is technically still a violation, the arity mismatch should probably be dealt with first, so let's have the compiler handle that.

Fixes #1250

### DIFF
--- a/bundle/regal/rules/bugs/inconsistent-args/inconsistent_args.rego
+++ b/bundle/regal/rules/bugs/inconsistent-args/inconsistent_args.rego
@@ -24,6 +24,9 @@ report contains violation if {
 
 	some name, args_list in function_args_by_name
 
+	# leave that to the compiler
+	not _arity_mismatch(args_list)
+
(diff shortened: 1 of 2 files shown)


charlieegan3 (commented):
LGTM

charlieegan3 (approved):

PR #1253, opened 2024-11-11
ci: Allow race detector to fail

charlieegan3:
This step is unreliable and needs more investigation. It is currently running in optional mode.


anderseknert:
Thank you!

PR #1269, opened 2024-11-18
Add support for using input.yaml in Evaluate code lens

anderseknert:
This addresses a request filed in the VS Code extension: https://github.com/open-policy-agent/vscode-opa/issues/308

Sadly this doesn't yet work for the debug feature as OPA currently only will do JSON decoding in that path, so next step is to submit a fix for that there.

### DIFF
--- a/internal/lsp/completions/providers/policy.go
+++ b/internal/lsp/completions/providers/policy.go
@@ -2,10 +2,8 @@ package providers
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
 	"os"
 
 	"github.com/open-policy-agent/opa/ast"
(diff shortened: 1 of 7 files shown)


internal/lsp/completions/providers/policy.go
```diff
@@ -70,20 +68,15 @@ func (p *Policy) Run(
 	inputContext["path_separator"] = string(os.PathSeparator)
 
 	workspacePath := uri.ToPath(opts.ClientIdentifier, opts.RootURI)
-	inputDotJSONPath, inputDotJSONReader := rio.FindInput(
+
+	inputDotJSONPath, inputDotJSONContent := rio.FindInput(
```
charlieegan3, line 72: Not sure if it's worth saving some future confusion with the names here.
```suggestion
	inputDocumentFilePath, inputDocumentFileContent := rio.FindInput(
```

charlieegan3 (approved):
LGTM!

PR #1271, opened 2024-11-20
Less frequent, and grouped dependabot updates

anderseknert:
Monthly may seem like an excaggeration, but our only real dependency is OPA, and we update that ourselves anyway. The rest are just peripheral, and security related updates will be suggested immediately regardless.


charlieegan3 (approved):

charlieegan3:
Agreed, the main pain is the rebase time, not the unwanted GHA updates as things are today.

Issue #1275, opened 2024-11-25
go install regal does not provide any version information


sbp-bvanb:
install regal
`go install github.com/styrainc/regal@v0.29.2`

issue `regal version`

```
Version:    unknown
Go Version: go1.23.3
Platform:   darwin/arm64
Commit:     unknown
Timestamp:  unknown
Hostname:   unknown
```

while
```
Version:    0.29.2
Go Version: go1.22.5
Platform:   darwin/amd64
Commit:     20a5cfaf7a721cc4d54b82f2f4ffc6c09b0b07cb
Timestamp:  2024-11-15T09:12:15Z
Hostname:   github.actions.local
```

is expected

sbp-bvanb:
I still think that go install should be supported including versioning when this option is chosen to install regal.

anderseknert:
Like you allude to, `go install` isn't "supported" in the sense that we list that as a recommended way to install Regal, but it's not like we'd stop anyone from improving that experience should they want that 🙂

Considering the huge number of golng tools around, "more than five" isn't exactly a huge number. But if you drop some links to projects that does this, perhaps that would be helpful for whoever wants to work on this.

anderseknert:
By all means, please do! 👍

sbp-bvanb:
PR has been created. Please review.

anderseknert:
@sbp-bvanb did you find a satisfiable solution to this eventually?

Given that this would require a larger effort to our build process, I don't think we're going to change this in the near future. I'll close this, but if someone is interested enough in this to put in the work, I'm happy to re-open.

anderseknert:
Thanks! In case you missed it, and you're using GitHub Actions, you can use https://github.com/StyraInc/setup-regal/ for some added convenience.

PR #1276, opened 2024-11-25
fix: [#1275] Return version if regal installed using go install

sbp-bvanb:
Return version if regal is installed using `go install`.

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -51,7 +51,7 @@ func (vi Info) String() string {
 
 func New() Info {
 	return Info{
-		Version:   unknownString(Version),
+		Version:   unknownVersionString(Version),
 		GoVersion: goVersion,
 		Platform:  platform,
 		Commit:    unknownString(Commit),
@@ -63,6 +63,14 @@ func New() Info {
 func unknownString(s string) string {
 	if s == "" {
 		return "unknown"
+	}
+
+	return s
+}
+
+func unknownVersionString(s string) string {
+	if s == "" {
+		return "v.29.3"
 	}
 
 	return s
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -3,8 +3,6 @@
     "OPA_CHECK_CAPABILITIES": "${workspacePath}/build/capabilities.json",
     "OPA_EVAL_CAPABILITIES": "${workspacePath}/build/capabilities.json"
   },
-  "opa.roots": [
-    "${workspaceFolder}/bundle"
-  ],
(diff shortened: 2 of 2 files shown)


pkg/version/version.go
```diff
@@ -67,3 +67,11 @@ func unknownString(s string) string {
 
 	return s
 }
+
+func unknownVersionString(s string) string {
+	if s == "" {
+		return "v.29.3"
```
anderseknert, line 73: This would need to be updated continuously with each new release, and I don't think that's feasible.

Looking at one of the links you provided, and as @srenatus previously mentioned, this method looks more appealing https://github.com/vektra/mockery/blob/master/pkg/logging/logging.go#L36-L45

Would you want to give that a shot?

sbp-bvanb: I tried, but I wonder whether it will work: https://github.com/golang/go/issues/50603#issuecomment-2300983645

Locally I have tried:

```
func buildVersion() string {
	version, ok := debug.ReadBuildInfo()
	fmt.Println("----------", version.Main.Version)
	if ok && version.Main.Version != "(devel)" && version.Main.Version != "" {
		return version.Main.Version
	}
	return "unknown"
}
```

```
func New() Info {
	return Info{
		Version:   buildVersion(),
		GoVersion: goVersion,
		Platform:  platform,
		Commit:    unknownString(Commit),
		Timestamp: unknownString(Timestamp),
		Hostname:  unknownString(Hostname),
	}
}
```

and 

go build -ldflags "-s -w -X github.com/styrainc/regal/pkg/version.Version=1.2.3"

./regal version

```
---------- (devel)
Version:    unknown
Go Version: go1.23.3
Platform:   darwin/arm64
Commit:     unknown
Timestamp:  unknown
Hostname:   unknown
```

anderseknert:
@sbp-bvanb yeah, having looked into this tonight, I don't think I've seen a single project where this works without having some external script setting the version somewhere in a file, and then have that read for this data.

From what I now gather, the "module version" we thought would work here is merely the version that could possibly be extracted from the module path, like "v2" in "github.com/styrainc/regal/v2" if that was a thing. Sorry if I led you astray here.

Like I said before, I think we can rule out any system that requires us to update some file manually each time we push a release. That's prone to being forgotten about, or be done wrong. So the alternative that remains would be to have a github action or similar do this for us. Or perhaps if Goreleaser can do something like that. I'll leave your issue open for ideas, but I think we can close this PR for now as neither your or my own attempts to get it working this way got where we wanted.

PR #1278, opened 2024-11-27
rules: Add InputFromTextWithOptions for Go users

charlieegan3:
This creates a helper function that creates an `Input` given a module as a string in a given version of Rego. This is being added for the rego-playground primarily.


srenatus (approved):

charlieegan3:
Thanks Stephan!

anderseknert:
Nice, thanks! Later we might want to support both if there are multiple bundles in a workspace, and manifest files that declare the Rego version. But that’s for another day 🙂

charlieegan3:
Yeah for sure, this was mostly just added as a better version for the playground for starters.

PR #1279, opened 2024-11-30
Have `redundant-existence-check` include function args

### LINKED ISSUE #1226: Improve `redundant-existence-check` to include function args
I caught myself writing something like this today.

```rego
response := http.send(url, opts) if url
```

Before I realized `if url` is a [redundant-existence-check](https://docs.styra.com/regal/rules/bugs/redundant-existence-check). Would be nice to have Regal tell me that.

anderseknert:
Fixes #1226

### DIFF
--- a/bundle/regal/result/result.rego
+++ b/bundle/regal/result/result.rego
@@ -196,7 +196,7 @@ _with_text(loc_obj) := loc if {
 # METADATA
 # description: |
 #   returns a "normalized" location object from the location value found in the AST.
-#   new code should most often use one of the ranged_ location functions instea, as
+#   new code should most often use one of the ranged_ location functions instead, as
 #   that will also include an `"end"` location attribute
 # scope: document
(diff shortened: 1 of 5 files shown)


srenatus (approved):
Code looks good. Not sure if we care about `fun(false)`, where the "existence check" isn't really redundant.

anderseknert:
@srenatus not sure I understand. Where's the existence check in `fun(false)`? 🙂

srenatus:
Excuse my brevity, the test in this PR included this rego:
```rego
fun(foo) if {
  foo # line 2
}
```
the "existence check" in line 2 doesn't seem redundant to me if I feed `false` into `fun()`.

anderseknert:
@srenatus ah, gotcha. I'll amend the docs for the rule to mention that case, and how being explicit for that case (e.g. `foo == false`/`foo != false` is a better solution anyway.

PR #1286, opened 2024-12-03
Fix false positive in `pointless-reassignment`

### LINKED ISSUE #1285: False positive in `pointless-reassignment` when array on left hand side
Originally reported in #1277, but the two issues turned out to be unrelated, so I'm creating a separate issue for this. 

This is currently reported as a pointless reassignment, but that should only apply when there is a single var on either side of the assignment:

```rego
parts := split(input.parts, ".")

s contains [b, a] if {
	[a, b] := parts
}
```

anderseknert:
That would happen despite the left hand side not being a var

Fixes #1285

### DIFF
--- a/bundle/regal/rules/style/pointless-reassignment/pointless_reassignment.rego
+++ b/bundle/regal/rules/style/pointless-reassignment/pointless_reassignment.rego
@@ -26,7 +26,10 @@ report contains violation if {
 
 	not expr["with"]
 
-	ast.assignment_terms(expr)[1].type == "var"
+	[lhs, rhs] := ast.assignment_terms(expr)
+
+	lhs.type == "var"
+	rhs.type == "var"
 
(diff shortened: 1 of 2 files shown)


srenatus (approved):

bundle/regal/rules/style/pointless-reassignment/pointless_reassignment.rego
```diff
@@ -26,7 +26,10 @@ report contains violation if {
 
 	not expr["with"]
 
-	ast.assignment_terms(expr)[1].type == "var"
+	[lhs, rhs] := ast.assignment_terms(expr)
+
+	lhs.type == "var"
+	rhs.type == "var"
```
srenatus, line 32: 💭 
```rego
every s in ast.assignment_terms(expr) { s.type == "var" }
```
Nope, not worth it.

PR #1289, opened 2024-12-06
Avoid allocations by not calling custom functions on hot path

anderseknert:
This is absolutely wild 🙃

**Before**
```
BenchmarkRegalLintingItself-10	1	3245668375 ns/op	6689411656 B/op	127195706 allocs/op
```

**After**
```
BenchmarkRegalLintingItself-10  1	2777580459 ns/op	5235852824 B/op	107686435 allocs/op
```

And
```
hyperfine -i --warmup 1 'regal lint bundle' 'regal-new lint bundle'
Benchmark 1: regal lint bundle
  Time (mean ± σ):      3.232 s ±  0.046 s    [User: 22.535 s, System: 0.623 s]
  Range (min … max):    3.169 s …  3.296 s    10 runs

Benchmark 2: regal-new lint bundle
  Time (mean ± σ):      2.754 s ±  0.046 s    [User: 18.641 s, System: 0.472 s]
  Range (min … max):    2.689 s …  2.813 s    10 runs

Summary
  regal-new lint bundle ran
    1.17 ± 0.03 times faster than regal lint bundle
```

No @srenatus, I will not use benchstat! 🤣

### DIFF
--- a/bundle/regal/ast/search.rego
+++ b/bundle/regal/ast/search.rego
@@ -111,13 +111,6 @@ _find_vars(value, last) := {"term": find_term_vars(function_ret_args(fn_name, va
 	function_ret_in_args(fn_name, value)
 }
 
-_find_vars(value, last) := {"assign": _find_assign_vars(value)} if {
-	last == "terms"
-	value[0].type == "ref"
-	value[0].value[0].type == "var"
-	value[0].value[0].value == "assign"
-}
-
 # `=` isn't necessarily assignment, and only considering the variable on the
 # left-hand side is equally dubious, but we'll treat `x = 1` as `x := 1` for
 # the purpose of this function until we have a more robust way of dealing with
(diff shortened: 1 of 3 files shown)


bundle/regal/ast/search.rego
```diff
@@ -126,11 +119,9 @@ _find_vars(value, last) := {"assign": _find_assign_vars(value)} if {
 	last == "terms"
 	value[0].type == "ref"
 	value[0].value[0].type == "var"
-	value[0].value[0].value == "eq"
+	value[0].value[0].value in {"assign", "eq"}
```
srenatus, line 122: That's a function call, too, isn't it? It'll become `internal.member_1(value[0].value[0].value, {"assign", "eq"})` or some such thing IIRC

srenatus (approved):

Issue #1300, opened 2024-12-13
regal fix with directory-package-mismatch assumes root as base


mrueg:
My repository looks like this

```
.git
policy/my-opa-rule/src.rego
README.md
```

in src.rego the package is called `my_opa_rule`
when I run `rego fix policy` it detects the directory-package-mismatch 

instead of 
```
.git
policy/my_opa_rule/src.rego
README.md
```

it creates the following structure:
```
.git
my_opa_rule/src.rego
policy
README.md
```

possible solution:

`rego fix $path` should include $path when fixing directory package mismatches.

anderseknert:
Hi there! And thanks for raising this.

I think it’s referenced in the docs for regal fix, but we should add a link from the docs of that rule too. You can set the project roots either via configuration, or by using .manifest files https://docs.styra.com/regal#project-roots

mrueg:
Thanks for the quick response! It was kinda unexpected when running the fix. I assumed base/root defaults to the path supplied, good to know!

anderseknert:
Yeah, that would work well for some commands, and others not so much, as e.g the Regal config may be in a directory above. So we tend to default to where the .regal directory is… and if that’s not found, I think the behavior differs some between commands, and  if invoked as a tool or a server, etc. Certainly room for improvement here.. :)

anderseknert:
Reopening this to make sure we add docs about project roots (or at least a link to the docs) in the directory-package-mismatch rule docs.

PR #1303, opened 2024-12-16
docs: add note about roots in directory-package-mismatch rule

charlieegan3:
Fixes https://github.com/StyraInc/regal/issues/1300


anderseknert (approved):
Doesn’t have to be many sub-projects, but just that the policy root is somewhere else than in the project root :) But this is good!

anderseknert (approved):

PR #1315, opened 2025-01-06
dependabot: Update GHA dependencies in groups

charlieegan3:
This should reduce the number of PRs we have when there are many updates to GHA deps, not just for go.mod deps.

Screenshot 2025-01-06 at 11 19 14


srenatus (approved):
Lgtm

charlieegan3:
thanks Stephan!

PR #1317, opened 2025-01-07
Prepare for OPA 1.0

anderseknert:
Go:
- Use OPA v1 packages
- Add function to parse Rego of unknown version
- Allow providing the Rego version via config or .manifest files
- Make `regal fix` formatter consider version to format for

Rego:
- Remove `import rego.v1` (via `opa fmt`)
- Add `config.capabilities.is_opa_v1`
- Use that to disable `use-if`, `use-contains`, `use-rego-v1`
- Update tests for 1.0 compliance

Docs:
- TODO remove `import rego.v1` from all examples
- `regal fix` doesn't work well
- Many failing e2e tests, despite some effort
- Test LSP functionality


.vscode/launch.json
```diff
@@ -0,0 +1,29 @@
...
+      "mode": "auto",
+      "program": "${workspaceFolder}",
+      "args": [
+        "fix",
+        "--dry-run",
+        "bundle"
+      ]
+    }
+  ]
+}
```
charlieegan3, line 29: nit: missing new line

charlieegan3 (approved):
Gosh, big one! great work. Few small comments. Happy to pick up the fix and LS work on this one.

pkg/linter/linter.go
```diff
@@ -451,6 +463,7 @@ func (l Linter) DetermineEnabledRules(ctx context.Context) ([]string, error) {
 
 	queryStr := `[rule |
         data.regal.rules[cat][rule]
+		count([true | data.regal.rules[cat][rule].notices]) == 0
```
charlieegan3, line 466: I think this is meant to check that there are no notices. But what it seems to do is check that if the notices set rule is defined.

If there is a notices rule, but the set is empty, we still want to enable the rule right?

charlieegan3: Might 		`object.get(data.regal.rules[cat][rule], "notices", set()) == set()` do the trick?

anderseknert: 👍

anderseknert: Yeah that was a late night addition 😄

internal/parse/parse.go
```diff
@@ -18,15 +18,104 @@ import (
...
+	// Iterate over RegoV1 and RegoV0 in that order
+	// If `import rego.v1`` is present in module, RegoV0CompatV1 is used
+	for i := range attemptVersionOrder {
+		version := attemptVersionOrder[i]
+
+		opts.RegoVersion = version
+
+		mod, err = ast.ParseModuleWithOpts(filename, policy, opts)
+		if err == nil {
+			if hasRegoV1Import(mod.Imports) {
```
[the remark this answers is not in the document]
anderseknert: That's still Rego v0 AFAICS.. i.e. you can import contains without using it, while you can't import rego.v1 without using contains, etc.

pkg/linter/linter.go
```diff
@@ -290,7 +290,19 @@ func (l Linter) Lint(ctx context.Context) (report.Report, error) {
 	l.stopTimer(regalmetrics.RegalFilterIgnoredFiles)
 	l.startTimer(regalmetrics.RegalInputParse)
 
-	inputFromPaths, err := rules.InputFromPaths(filtered)
+	var versionsMap map[string]ast.RegoVersion
+
+	// TODO: How should we deal with this in the language server?
+	// AllRegoVersions will call WalkDir on the root to find manifests, but that's obviously not
```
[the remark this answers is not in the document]
anderseknert: I suppose this is address by now, right?

PR #1318, opened 2025-01-08
Use concurrent map implementation

anderseknert:
To avoid mutex handling cluttering up the code, use a concurrent wrapper for maps where we need it.


charlieegan3 (approved):
👏 this is great!

PR #1321, opened 2025-01-08
e2e: Fix CLI test for 1.0 preparations and defaults

charlieegan3:
Mainly, this is ensuring the configs for each test are correct.

PR #1322, opened 2025-01-08
golangci-lint: Update and address linter issues

charlieegan3:
Addresses linter issues found in https://github.com/StyraInc/regal/pull/1321

### DIFF
--- a/pkg/linter/linter_test.go
+++ b/pkg/linter/linter_test.go
@@ -773,7 +773,7 @@ func BenchmarkRegalLintingItself(b *testing.B) {
 
 	var rep report.Report
 
-	for i := 0; i < b.N; i++ {
+	for range b.N {
 		rep, err = linter.Lint(context.Background())
 		if err != nil {
 			b.Fatal(err)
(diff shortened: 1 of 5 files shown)


pkg/linter/linter_test.go
```diff
@@ -773,7 +773,7 @@ func BenchmarkRegalLintingItself(b *testing.B) {
 
 	var rep report.Report
 
-	for i := 0; i < b.N; i++ {
+	for range b.N {
```
anderseknert, line 776: This doesn't work in Go 1.22 which is what we target.. but I guess since these are test files, it doesn't matter.

anderseknert (approved):

pkg/linter/linter_test.go
charlieegan3: Hmm, I wonder why golangcilint wasn't able to work that out? perhaps because we have a more recent toolchain version?

anderseknert: Yeah, that sounds reasonable, and smart :) If it did this generally, we'd have a lot of places to update I reckon.

PR #1325, opened 2025-01-09
lsp/server: Cache AllRegoVersions at config load

charlieegan3:
This does not reload when a .manifest file is changed yet.

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -88,20 +88,21 @@ func NewLanguageServer(ctx context.Context, opts *LanguageServerOptions) *Langua
 
 	var ls *LanguageServer
 	ls = &LanguageServer{
-		cache:                    c,
-		regoStore:                store,
-		logWriter:                opts.LogWriter,
-		logLevel:                 opts.LogLevel,
-		lintFileJobs:             make(chan lintFileJob, 10),
-		lintWorkspaceJobs:        make(chan lintWorkspaceJob, 10),
-		builtinsPositionJobs:     make(chan lintFileJob, 10),
-		commandRequest:           make(chan types.ExecuteCommandParams, 10),
-		templateFileJobs:         make(chan lintFileJob, 10),
-		configWatcher:            lsconfig.NewWatcher(&lsconfig.WatcherOpts{LogFunc: ls.logf}),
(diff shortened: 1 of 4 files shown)


internal/lsp/server.go
```diff
@@ -114,10 +115,13 @@ type LanguageServer struct {
...
 
-	configWatcher                        *lsconfig.Watcher
-	loadedConfig                         *config.Config
+	configWatcher *lsconfig.Watcher
+	loadedConfig  *config.Config
+	// this is also used to lock the updates to the cache of enabled rules
+	loadedConfigLock                     sync.Mutex
 	loadedConfigEnabledNonAggregateRules []string
 	loadedConfigEnabledAggregateRules    []string
+	loadedConfigAllRegoVersions          *concurrent.Map[string, ast.RegoVersion]
```
anderseknert, line 124: Neat :)

internal/lsp/server.go
```diff
@@ -1096,6 +1113,32 @@ func (l *LanguageServer) StartWebServer(ctx context.Context) {
 	l.webServer.Start(ctx)
 }
 
+func (l *LanguageServer) determineVersionForFile(fileURI string) ast.RegoVersion {
+	var versionedDirs []string
+
+	// if we have no information, then we can return the default
+	if l.loadedConfigAllRegoVersions == nil || l.loadedConfigAllRegoVersions.Len() == 0 {
```
anderseknert, line 1120: Perhaps we should have Len() always return 0 when its parent is nil? It would make the check here a little cleaner.

internal/lsp/server.go
```diff
@@ -1185,7 +1228,13 @@ func (l *LanguageServer) templateContentsForFile(fileURI string) (string, error)
...
 	}
 
-	return fmt.Sprintf("package %s\n\nimport rego.v1\n", pkg), nil
+	version := l.determineVersionForFile(fileURI)
+
+	if version == ast.RegoV0 {
+		return fmt.Sprintf("package %s\n\nimport rego.v1\n", pkg), nil
+	}
+
+	return fmt.Sprintf("package %s\n\n", pkg), nil
```
anderseknert, line 1237: Very good!

anderseknert (approved):
Terrific!

Issue #1324, opened 2025-01-09
Have Regal the language server print its location and version on startup


anderseknert:
Since it's currently rather opaque to users where editors like VS Code get their Regal language server from and which versoon is in use, a modest improvement would be to simply print this information when the language server starts up. Users could then find this information by checking the output of the server.

PR #1326, opened 2025-01-09
cmd/languageserver: Log path & version at startup

charlieegan3:
Fixes https://github.com/StyraInc/regal/issues/1324

### DIFF
--- a/cmd/languageserver.go
+++ b/cmd/languageserver.go
@@ -5,12 +5,14 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"path/filepath"
 	"syscall"
 
 	"github.com/spf13/cobra"
 
 	"github.com/styrainc/regal/internal/lsp"
 	"github.com/styrainc/regal/internal/lsp/log"
+	"github.com/styrainc/regal/pkg/version"
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


cmd/languageserver.go
```diff
@@ -25,6 +27,23 @@ func init() {
...
+				fmt.Fprintln(os.Stderr, "error getting executable:", err)
+			} else {
+				absPath, err := filepath.Abs(exe)
+				if err != nil {
+					fmt.Fprintln(os.Stderr, "error getting executable path:", err)
+				} else {
+					v := version.Version
+					if v == "" {
+						v = "Unknown"
+					}
```
srenatus, line 41: Do you know [`cmp.Or`](https://pkg.go.dev/cmp#Or)? We could write this as
```suggestion
					v := cmp.Or(version.Version, "Unknown")
```

anderseknert (approved):

srenatus (approved):

cmd/languageserver.go
anderseknert: Oh, nice! Had no idea that was a stdlib thing. I had already merged this but we could easily sprinkle that in lots of places I'm sure.

PR #1328, opened 2025-01-09
Prepare docs for OPA 1.0 compatibility

anderseknert:
- Remove `import rego.v1` from examples
- Add new page covering Regal and OPA 1.0
- Update logic of any rules obsolete in 1.0 to make sure they're automatically disabled when 1.0 is targeted, and continue to work when older versions are linted
- Update a few tests to remove `import rego.v1`
- Many minor fixes related to this theme

### DIFF
--- a/docs/opa-one-dot-zero.md
+++ b/docs/opa-one-dot-zero.md
@@ -1,0 +1,76 @@
+# OPA 1.0 and Regal
+
+While we always recommend using the latest version of OPA, we're well aware that there may be situations where — for
+one reason or another — that might not be possible. As we want everyone to benefit from Regal, we do our very best to
+ensure it works seamlessly with OPA versions both before and after 1.0, and even projects that use a mix of both! While
+this should mostly work out of the box and without additional configuration, it's good to be aware of how Regal parses
+and lints policies of different versions of Rego, and how you can tell Regal to target only a specific version.
+
+**Note:** This document does not cover the specifics of OPA 1.0, but rather how Regal works with it. If you want to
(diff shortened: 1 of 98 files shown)


docs/opa-one-dot-zero.md
```diff
@@ -0,0 +1,76 @@
...
+
+See the documentation covering Regal's [configuration](https://docs.styra.com/regal#configuration) for more information
+on [configuring Rego version](https://docs.styra.com/regal#configuring-rego-version) for your project.
+
+Finally, Regal will automatically parse and lint any file with a `_v0.rego` suffix as Rego v0. This is intended only
+for testing and development, where you sometimes may want to try something out using and older Rego version without
+configuration. Note that this has lower precedence than Rego versions set by other means, and should not be considered
+as anything but a convenience for testing.
+
+## Rules disabled with OPA 1.0
```
charlieegan3, line 47: Some rules are labelled `Notice: Rule made obsolete by OPA 1.0`, others `Notice: Rule disabled with OPA 1.0` - is there a subtle difference I'm missing?

charlieegan3 (approved):
LGTM!

docs/opa-one-dot-zero.md
anderseknert: The`import-shadows-import` rule should have the "obsolete" header, and I'll fix that. The only other one that has "rule disabled" header is `deprecated-builtin`. That rule isn't really made obsolete — it's just that we currently don't have any deprecated built-ins in OPA 1.0. But that may very well change, and the rule will be enabled once that happens.

PR #1336, opened 2025-01-14
docs: opa-fmt v1/v0 notes

charlieegan3:
This explains the different behaviour of `opa fmt` between v0 and v1 when working with v0/v1 files in Rego.


anderseknert (approved):
Thanks!

PR #1337, opened 2025-01-14
tidy: Use cmp.Or in a few places

charlieegan3:
As suggested here: https://github.com/StyraInc/regal/pull/1326#discussion_r1908895712

Sure there are more, but these are the first ones I found


anderseknert (approved):
Neat!

Issue #1092, opened 2024-09-10
lsp: Stop server templating files that are no longer Rego


charlieegan3:
When a file is renamed from .rego to .foo, the server will continue to template the file until it's opened again as the formatting requests are still sent to it.

https://github.com/user-attachments/assets/7f4d7415-5713-44a1-b63d-a34147573110

PR #1340, opened 2025-01-16
lsp: ignore after rename if not Rego

charlieegan3:
There were two places where a file would be processed after a rename incorrectly:

- inlay hints were still requests if the file wasn't rego
- new files were still formatting if the file wasn't rego any more

Fixes https://github.com/StyraInc/regal/issues/1092


anderseknert (approved):
Thanks!

PR #1345, opened 2025-01-16
fix: filter out `regal.debug` codelens

jglasovic:
This PR fixes a bug with filtering out `regal.debug` codelens when `enableDebugCodelens != true`
I ran into it a few days ago and it seems like it has been here since the implementation (https://github.com/StyraInc/regal/pull/1176)


anderseknert (approved):
Thanks! Tbh, I don't remember we even made it an option to disable that 😅

PR #1347, opened 2025-01-17
Add workspace folder type, and "support" that capability

anderseknert:
Ironic quotes on support since this seems to be a client feature only, but I figured we'd still document that in the code rather than elsewhere, for reference next time we look into that.

Next up to look into supporting this in the OPA VS Code extension.


charlieegan3 (approved):
Odd one, but LGTM

PR #1349, opened 2025-01-18
Language server refactoring

anderseknert:
While there's more to do here, we've got to start somewhere.

- Simplify handlers by extracting shared logic
- Use "global" vars for static slices to avoid needless allocations
- No named return values
- Add GetContentAndModule convenience function to reduce boilerplate
- Various fixes

This should not change behavior of anything, just nake server.go a little more manageable.


charlieegan3 (approved):
This is really, really nice. Thanks for taking some time to brush it up!

charlieegan3:
I am going to merge as I want to work on an issue with the parse errors and would like to be up to date.

PR #1351, opened 2025-01-20
perf: move config check out of loop

anderseknert:
In main.rego there are two places where we extract a relative path from the input policy file. This was however done in a loop despite only using variables from outside of the loop. Turns out there was close to a million allocations hiding there, lol.

**BenchmarkRegalLintingItself-10 Before**
```
1875658375 ns/op    3539151104 B/op    70020076 allocs/op
```
**BenchmarkRegalLintingItself-10 After**
```
1821361667 ns/op    3499082704 B/op    69062781 allocs/op
```


charlieegan3 (approved):
🔍

Issue #1344, opened 2025-01-16
lsp: fix parse error violations


charlieegan3:
It appears this line is no longer getting a list of parse errors like we used to, and so we get no parse error violations any more.

https://github.com/StyraInc/regal/blob/main/internal/lsp/lint.go#L79

I guess something might have changed with our using Roast now?

charlieegan3:
Fixing this would aid in the solution for https://github.com/StyraInc/regal/issues/1123

anderseknert:
Roast only serializes the parsed modules to JSON. The parsing is still done in Regal (via OPA, naturally), so it’s likely something we’ve changed here recently, perhaps with all the things that went into support Rego v1.

charlieegan3:
> Roast only serializes the parsed modules to JSON. 

🤦 of course.

anderseknert:
Well, not entirely impossibe after all — we use Roast to serialize all the things we can in Regal now. And I guess that includes errors. Still, I'd be surprised if that was the issue here... but who knows :)

Issue #1123, opened 2024-09-21
Code lenses look weird when syntax errors are reported


anderseknert:
We can't evaluate or debug files with syntax errors, but it looks like we keep sending these from the server in that case, presumably from the last known good state, which sometimes makes them appear in weird locations. We should simply not return anything in this case. 

<img width="495" alt="Screenshot 2024-09-21 at 20 37 54" src="https://github.com/user-attachments/assets/a03bdb45-dd7c-422e-89a2-0596a73b00ea">

Perhaps later on we can have those actions located before any syntax errors in the file remain and when clicked help the user understand more about the error... but  it's likely better to show syntax errors in-lined like linter issues anyway. So let's just hide the actions for now.

charlieegan3:
This is partly blocked by https://github.com/StyraInc/regal/issues/1344, as we get 0 parse diags set now, so we can't tell if the module is stale or not.

charlieegan3:
I think the title is referring to Code Lenses rather than actions?

anderseknert:
Indeed! Fixed.

PR #1352, opened 2025-01-20
lsp: Ensure parse errors are saved as diagnostics correctly

charlieegan3:
Fixes https://github.com/StyraInc/regal/issues/1344

Fixes https://github.com/StyraInc/regal/issues/1123

(in fact, the final two lsp bugs we have in the tracker: https://github.com/StyraInc/regal/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22language%20server%20protocol%22%20label%3Abug)

### DIFF
--- a/internal/lsp/lint.go
+++ b/internal/lsp/lint.go
@@ -2,7 +2,6 @@ package lsp
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"strings"
@@ -30,6 +29,7 @@ func updateParse(
 	store storage.Store,
 	fileURI string,
 	builtins map[string]*ast.Builtin,
+	version ast.RegoVersion,
 ) (bool, error) {
 	content, ok := cache.GetFileContents(fileURI)
 	if !ok {
@@ -43,8 +43,7 @@ func updateParse(
 		module *ast.Module
 	)
 
-	// TODO: IF not set in config, or from .manifest
-	module, err = rparse.ModuleUnknownVersionWithOpts(fileURI, content, rparse.ParserOptions())
+	module, err = rparse.ModuleWithOpts(fileURI, content, ast.ParserOptions{RegoVersion: version})
 	if err == nil {
 		// if the parse was ok, clear the parse errors
 		cache.SetParseErrors(fileURI, []types.Diagnostic{})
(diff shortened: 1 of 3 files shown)


internal/lsp/lint.go
```diff
@@ -76,29 +75,30 @@ func updateParse(
 		return true, nil
 	}
 
-	unwrappedError := errors.Unwrap(err)
```
charlieegan3, line 79: I think that something changed such that this was the issue.

internal/lsp/lint.go
```diff
@@ -76,29 +75,30 @@ func updateParse(
...
-	var parseError *ast.Error
-
-	if errors.As(unwrappedError, &parseError) {
-		astErrors = append(astErrors, astError{
-			Code:     parseError.Code,
-			Message:  parseError.Message,
-			Location: parseError.Location,
-		})
-	} else {
-		jsonErrors, err := json.Marshal(unwrappedError)
```
charlieegan3, line 95: But i have done away with all this stuff from before.

internal/lsp/server.go
```diff
@@ -1659,6 +1652,11 @@ func (l *LanguageServer) handleTextDocumentInlayHint(params types.TextDocumentIn
 }
 
 func (l *LanguageServer) handleTextDocumentCodeLens(ctx context.Context, params types.CodeLensParams) (any, error) {
+	parseErrors, ok := l.cache.GetParseErrors(params.TextDocument.URI)
+	if ok && len(parseErrors) > 0 {
+		return []types.CodeLens{}, nil
```
charlieegan3, line 1657: This is the fix for the code lenses issue.

internal/lsp/server.go
```diff
@@ -2004,11 +1998,7 @@ func (l *LanguageServer) handleTextDocumentFormatting(
 	switch formatter {
 	case "opa-fmt", "opa-fmt-rego-v1":
 		opts := format.Opts{
-			RegoVersion: rules.RegoVersionFromVersionsMap(
-				l.loadedConfigAllRegoVersions.Clone(),
-				strings.TrimPrefix(params.TextDocument.URI, l.workspaceRootURI),
-				ast.RegoUndefined,
-			),
+			RegoVersion: l.regoVersionForURI(params.TextDocument.URI),
```
charlieegan3, line 2001: This is just DRYing up with some more use of regoVersionForURI and is otherwise unrelated.

internal/lsp/lint.go
```diff
@@ -43,8 +43,7 @@ func updateParse(
 		module *ast.Module
 	)
 
-	// TODO: IF not set in config, or from .manifest
-	module, err = rparse.ModuleUnknownVersionWithOpts(fileURI, content, rparse.ParserOptions())
+	module, err = rparse.ModuleWithOpts(fileURI, content, ast.ParserOptions{RegoVersion: version})
```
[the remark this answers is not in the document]
charlieegan3: Hmm, this should perhaps still use ModuleUnknownVersionWithOpts when RegoVersion is undefined....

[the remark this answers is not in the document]
charlieegan3: Ahh, that's already handled in ModuleWithOpts - if the version is undefined. Then ModuleUnknownVersionWithOpts is used. I will add a test for this either way.

PR #1354, opened 2025-01-20
Fix missing annotations in LSP parsing

anderseknert:
Also, update the missing-metadata rule to report violation locations on the rule head ref and not the whole rule head.


charlieegan3 (approved):
LGTM, the main change here is using `rparse.ParserOptions()` instead right?

anderseknert:
@charlieegan3 — yes, that's fixing the bug that cause annotations to be dropped by the parser

PR #1356, opened 2025-01-21
fix: Remove mandatory fixes

charlieegan3:
This functionality is no longer used.


srenatus (approved):
Deleted code is the best code

PR #1358, opened 2025-01-23
refactor: better use of test helpers

anderseknert:


.lsp/.cache/db.transit.json
srenatus, line 1: Cute. Can we remove that? 😉

srenatus (approved):

.clj-kondo/.cache/v1/lock
charlieegan3, line 1: this is from some vs code ext right? I seem to remember this popping up at some point too.

e2e/cli_test.go
```diff
@@ -1083,7 +1072,7 @@ Cannot move multiple files to: bar/bar.rego
 	}
 
 	for file, expectedContent := range initialState {
-		bs := testutil.Must(os.ReadFile(filepath.Join(td, file)))(t)
+		bs := testutil.MustReadFile(t, filepath.Join(td, file))
```
charlieegan3, line 1075: We could even consider some custom linters for this kind of thing... 🤔

charlieegan3 (approved):
yeah this all looks good to me, would be nice to lint for some of these if it's not too much work as I fear I might forget...

e2e/cli_test.go
srenatus: Just remove `testutil.Must` and the compiler is your linter 🙃 On a more productive note, https://github.com/quasilyte/go-ruleguard might be your friend.

charlieegan3: Ooh, that looks like a nice project!

.clj-kondo/.cache/v1/lock
anderseknert: Yeah it's from Calva I believe. Incredibly annoying that they do that in projects without Clojure files simply because I mistakenly pressed siome menu option that belongs to their extension 🤦

.lsp/.cache/db.transit.json
anderseknert: Yep..

e2e/cli_test.go
anderseknert: Pretty sure the compiler enforces types for imported testutil packages as well ;)

PR #1359, opened 2025-01-23
Remove custom Ref stringer

anderseknert:
I can't remember why we have this in the first place, but in case some new issue reminds us, we can deal with it then.


charlieegan3 (approved):
"I do not recall"

PR #1363, opened 2025-01-23
Test getting rid of timeout in flakey test

anderseknert:
YOLO! But really, this feels more like an issue with the timeout than with the actual logic... let's see.


anderseknert:
Turns out I was wrong. Who would have thought! Oh well, one step closer, and I'll keep this open for more things to try. As this never fails locally, I don't know of a better way to test this.

Issue #1075, opened 2024-09-08
Allow "global" (user level) Regal configuration file


anderseknert:
While it's well-documented and easy to ignore linter rule violations via a project's `.regal/config.yaml` file, some users have told us that issues reported by Regal in the context of authoring policies in a supported (LSP) editor — can be somewhat annoying when:

- working with code that belongs to someone else (like a customer, client, someone you support)
- just messing around with some Rego in a temporary scratch directory

It's possible to achieve this today already, as Regal will traverse the directory tree upwards looking for configuration, using the first one it finds. But in addition to this, we could easily support a user-level `~/.config/regal/config.yaml` too. This configuration would only apply when none of the above methods found a config, and config files should not be "merged". I.e. a project-level config always takes precedence.

PR #1378, opened 2025-01-29
config: Use global user config if required

charlieegan3:
If a config file is not found in a parent of the workspace, then we attempt to load the user's global default file instead.

Fixes https://github.com/StyraInc/regal/issues/1075

Screenshot 2025-01-29 at 14 23 56

LS will also log the location of the file used.
Screenshot 2025-01-29 at 14 22 10

### DIFF
--- a/README.md
+++ b/README.md
@@ -403,8 +403,8 @@ project:
 Regal will automatically search for a configuration file (`.regal/config.yaml`
 or `.regal.yaml`) in the current directory, and if not found, traverse the
 parent directories either until either one is found, or the top of the directory
-hierarchy is reached. If no configuration file is found, Regal will use the
-default configuration.
+hierarchy is reached. If no configuration file is found, and no file is found at
+`~/.config/regal/config.yaml` either, Regal will use the default configuration.
 
 A custom configuration may be also be provided using the `--config-file`/`-c`
 option for `regal lint`, which when provided will be used to override the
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -403,8 +403,8 @@ project:
 Regal will automatically search for a configuration file (`.regal/config.yaml`
 or `.regal.yaml`) in the current directory, and if not found, traverse the
 parent directories either until either one is found, or the top of the directory
-hierarchy is reached. If no configuration file is found, Regal will use the
-default configuration.
+hierarchy is reached. If no configuration file is found, and no file is found at
+`~/.config/regal/config.yaml` either, Regal will use the default configuration.
```
anderseknert, line 407: This is good, but I'd probably add a section on its own about "User-level configuration" or something. Where we also explain when we think it's appropriate to use one, and that regular projects should rather check in their Regal config along with their policies.

anderseknert (approved):
Thanks! Perhaps expand a little on the docs, but LGTM 👍

README.md
```diff
@@ -403,13 +403,26 @@ project:
...
 
 A custom configuration may be also be provided using the `--config-file`/`-c`
 option for `regal lint`, which when provided will be used to override the
 default configuration.
 
+### User-level Configuration
```
anderseknert, line 413: This is excellent!

anderseknert (approved):

Issue #1214, opened 2024-10-21
Have `regal lint` fail when unknown category or rule is provided via `--enable`/`--disable` options


anderseknert:
```
regal lint --disable-all --enable foo .
```

Should have regal fail loudly unless the `foo` rule was actually found. Currently it'll just report this as a success, which while technically true, is not likely to be what the user wanted. Note that we'll also need to take custom rules into account, so this is probably a check we'll have to do some time into initialization.

PR #1381, opened 2025-01-30
linter: Show error if rule or category is invalid

charlieegan3:
Fixes https://github.com/StyraInc/regal/issues/1214

### DIFF
--- a/internal/util/set.go
+++ b/internal/util/set.go
@@ -1,0 +1,68 @@
+package util
+
+// Set is a generic set implementation.
+type Set[T comparable] struct {
+	elements map[T]struct{}
+}
+
+// NewSet creates a new set from the supplied items.
+func NewSet[T comparable](items ...T) *Set[T] {
+	s := &Set[T]{elements: make(map[T]struct{})}
+
+	s.Add(items...)
+
+	return s
+}
+
+// Add adds one or more items to the set.
+func (s *Set[T]) Add(items ...T) {
+	for _, item := range items {
+		s.elements[item] = struct{}{}
+	}
+}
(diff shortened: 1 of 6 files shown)


internal/util/set.go
```diff
@@ -0,0 +1,68 @@
...
+
+// Remove removes one or more items from the set.
+func (s *Set[T]) Remove(items ...T) {
+	for _, item := range items {
+		delete(s.elements, item)
+	}
+}
+
+// Contains checks if all given items are in the set.
+func (s *Set[T]) Contains(items ...T) bool {
```
anderseknert, line 32: Neat!

anderseknert (approved):
This is great! 

Did check that this failure propagates cleanly to the CLI? Like `regal lint --enable unknown` gives a friendly message?

The Set implementation is really nice compared to the map version, I think, and there's likely more places in Regal where we can replace the latter with the former? But that can come later. Thinking of it, we might even want to move this over to OPA if we can benefit from it there too. And shared utilities are always nice :)

Speaking of next steps, we could probably send this info over to evaluation and skip some needless processing in Rego as we already know which rules to evaluate now? But OTOH, it feels wrong to do less in Rego and more in Go... so let's see.

charlieegan3:
> Did check that this failure propagates cleanly to the CLI?

I think this is clear enough:


```
…/regal linter-rule-config-check ➜ go run main.go lint bundle/ --enable=woza
error(s) encountered while linting: validation failed: unknown rules: [woza]
exit status 1
```

Will have a think about some more places to bin struct{}{} s

Issue #1380, opened 2025-01-30
Add fixer for `non-raw-regex-pattern`


anderseknert:
Should be a fairly low-hanging fruit, but I guess we'll see :)

It'd be great if `regal fix` (and the language server) could rewrite:

```rego
regex.match("foo\\.bar", str)
```

to

```rego
regex.match(`foo\.bar`, str)
```

PR #1382, opened 2025-01-30
fix: non-raw regex pattern

charlieegan3:
Also update the lsp commands to support the end locations so that we can operate more accurately rather than guessing the end of the target range.

Fixes https://github.com/StyraInc/regal/issues/1380

### DIFF
--- a/internal/lsp/command.go
+++ b/internal/lsp/command.go
@@ -1,8 +1,13 @@
 package lsp
 
-import "github.com/styrainc/regal/internal/lsp/types"
+import (
+	"encoding/json"
 
-func toAnySlice(a []string) *[]any {
+	"github.com/styrainc/regal/internal/lsp/types"
+	"github.com/styrainc/regal/internal/util"
+)
+
+func toAnySlice(a ...string) *[]any {
 	b := make([]any, len(a))
 	for i := range a {
 		b[i] = a[i]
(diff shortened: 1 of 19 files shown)


internal/lsp/command.go
```diff
@@ -1,8 +1,13 @@
 package lsp
 
-import "github.com/styrainc/regal/internal/lsp/types"
+import (
+	"encoding/json"
```
anderseknert, line 4: Doesn't matter much in this context, but in most of Regal we use this now https://github.com/anderseknert/roast/blob/main/pkg/encoding/json.go

pkg/fixer/fixes/nonrawregexpattern.go
```diff
@@ -0,0 +1,64 @@
...
+			line[startIdx] = '`'
+			fileChanged = true
+		}
+
+		if line[endIdx] == '"' {
+			line[endIdx] = '`'
+			fileChanged = true
+		}
+
+		lines[loc.Row-1] = string(line)
```
anderseknert, line 50: We need to account for escaping here too, unless I missed something. I.e. 
```
"\\." -> `\.`
```
ast the \\. itself doesn't have to be escaped in a raw string.

pkg/fixer/fixes/nonrawregexpattern_test.go
```diff
@@ -0,0 +1,229 @@
...
+			fc: &FixCandidate{Filename: "test.rego", Contents: `package test
+
+all_digits if {
+    regex.match(` + "`" + `[\\d]+` + "`" + `, "12345")
+}
+`},
+			contentAfterFix: `package test
+
+all_digits if {
+    regex.match(` + "`" + `[\\d]+` + "`" + `, "12345")
```
anderseknert, line 28: So as per above comment, this should be:

```
`[\d]+`
```

anderseknert (changes requested):
Awesome! One issue to fix, then good to merge.

anderseknert:
Also just reminded that there are some places in the docs where we list the supported fixers, and this should be added to those locations too. And I guess the docs for the rule itself :)

pkg/fixer/fixes/nonrawregexpattern_test.go
charlieegan3: Ahh, very good spot. I have fixed this in 32c8f22b0d46a1ea2823bf6e1eb26f64becdb99a!

anderseknert (approved):
LGTM! Thanks for bearing with me :)

charlieegan3:
tbf, the \\ issues was fairly fundamental 🤦 I was just so focused on not changing other chars to ` etc etc.

PR #1383, opened 2025-01-30
Consistently use new `Set` type

anderseknert:
This was added by @charlieegan3 yesterday, and I much prefer the ergonomics of this compared to using `map[T]struct{}` directly. So let's do it!

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,5 +7,9 @@
     "${workspaceFolder}/bundle"
   ],
   "opa.strictMode": true,
-  "go.lintTool": "golangci-lint"
+  "go.lintTool": "golangci-lint",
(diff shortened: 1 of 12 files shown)


.vscode/settings.json
```diff
@@ -7,5 +7,9 @@
     "${workspaceFolder}/bundle"
   ],
   "opa.strictMode": true,
-  "go.lintTool": "golangci-lint"
+  "go.lintTool": "golangci-lint",
+  "go.buildFlags": [
+    "-tags",
+    "e2e"
+  ]
```
anderseknert, line 14: This was needed to have the cli_test.go file compiled / checked by VS Code.

charlieegan3 (approved):
This looks really nice!

Issue #821, opened 2024-06-10
Be specific about number of errors vs warnings in `regal lint` summary


anderseknert:
The summary presented by `regal lint` currently says something like:

```
58 files linted. 16 violations found in 2 files.
```

It would be good to differentiate configured errors vs warnings here rather than just say "violations".

anderseknert:
This was fixed in #821

PR #1387, opened 2025-02-03
Be specific about number of errors vs warnings in regal lint summary #821

anaypurohit0907:
before it used to look like this :
image

and now it would look like this:
image


pkg/reporter/reporter.go
```diff
@@ -105,6 +105,20 @@ func NewJUnitReporter(out io.Writer) JUnitReporter {
...
+	for i, violation := range r.Violations {
+		if violation.Level == "warning" {
+			numsWarning++
+		}
+		if violation.Level == "error" {
+			numsError++
+		}
+		if i+1 > numsWarning {
+
+		}
```
anderseknert, line 120: This doesn't look right..

pkg/reporter/reporter.go
```diff
@@ -116,12 +130,34 @@ func (tr PrettyReporter) Publish(_ context.Context, r report.Report) error {
...
+
+			pluralError := ""
+			if numsError == 0 {
+				pluralError = "s"
+			}
+			if numsError > 0 {
+				pluralError = "s"
+			}
 
+			footer += fmt.Sprintf("(%d warning%s,%d Error%s) found", numsWarning, pluralWarnings, numsError, pluralError)
```
anderseknert, line 157: Good!

pkg/reporter/reporter.go
anaypurohit0907: oh lol, i forgot about that

anderseknert:
Thanks for contributing to Regal!

PR #1389, opened 2025-02-03
lsp/test: Error in test should be log instead

charlieegan3:


charlieegan3:
I ran this 1000x in https://github.com/StyraInc/regal/actions/runs/13116621278/job/36592398229?pr=1389

anderseknert (approved):

PR #1392, opened 2025-02-04
Remove unused fixer code

anderseknert:
Since we no longer have mandatory fixes, we don't need code to deal with that. If we do later, it'll be in git history :)


charlieegan3 (approved):
Thanks!

PR #1393, opened 2025-02-04
Rego-based opa-fmt

### LINKED ISSUE #1298: Add a built-in function for formatting, and sunset the Go linter
When I started hacking on Regal, I was thinking there would be a lot of things that couldn't be done using Rego alone. One obvious thing was formatting, but the plan was to add more Go linters for things that turned out to be impossible to do in Rego. I no longer think that we'll need linters written in Go, and if we find problems like that going forward, we can implement custom functions for that. Let's make one for formatting and have the `opa-fmt` rule use that instead. With the only Go linter gone, we can decomission that code.

anderseknert:
Well, with the help of a new built-in function, anyway.

This allows us to ditch the system for Go-based rules entirely, which feels like a big step for this project!

Fixes #1298


anderseknert:
Submitting this with a few known caveats for discussion:
* Since we split the file into lines [as part of building the input ](https://github.com/StyraInc/regal/blob/a283bee6b638b82b5d7c15d2929ea196c88417c0/internal/parse/parse.go#L142), this formatter won't be able to identify files where `\r\n` is used for linebreaks, which is a formatting violation. Not sure how to best deal with that, but I guess we could add something to the input in those (very rare) cases. 
* One e2e test has been modified to success (hehe) as it relates to how the fixer handles v0/v1 Rego, and we'll want to look at that separately. 
* I have not yet tested the performance difference of this vs the Go implementation.

e2e/cli_test.go
```diff
@@ -919,13 +919,13 @@ foo/main_test.rego -> wow/main_test.rego:
 
 In project root: %[1]s/v0
 main.rego:
-- opa-fmt
+- use-rego-v1
```
anderseknert, line 922: This is is incorrect, and should be fixed. I'm only not sure where the issue is yet.

pkg/linter/linter.go
```diff
@@ -1209,70 +1048,6 @@ func (l Linter) GetConfig() (*config.Config, error) {
 	return l.combinedCfg, nil
 }
 
-func (l Linter) enabledGoRules() ([]rules.Rule, error) {
```
charlieegan3, line 1212: this is great!

charlieegan3 (commented):
this looks great to me!

charlieegan3 (approved):

e2e/cli_test.go
anderseknert: It's not incorrect actually, but we'll want to find out why there is a difference :)

PR #1399, opened 2025-02-06
Rule: `use-object-keys`

### LINKED ISSUE #1044: Rule: `use-object-keys`
Another rule in the "let's use a built-in for that" series. Whenever a set comprehension is used to collect the keys from an object, we should recommend using `object.keys` instead, as it's easier to read, and less "imperative".

**Avoid**
```rego
package policy

keys := {k | some k, _ in input.object}

# or

keys := {k | some k; input.object[k]}
```

**Prefer**
```rego
package policy

keys := object.keys(input.object)
```

Most often, we can't really tell the type of the input used in a comprehension like this statically. Meaning that something like:

```
{user | some user, _ in users}
```

..._could_ be collecting indices of an array, or "key" values from a set.  However, collecting array indices into a set seems unlikely / pointless, and collecting values from a set into another set with no further processing is equally pointless. Obviously, we'll need to ensure that only the simples form matches, not when more things happen in the comprehension.

anderseknert:
Fixes #1044

### DIFF
--- a/docs/rules/idiomatic/use-object-keys.md
+++ b/docs/rules/idiomatic/use-object-keys.md
@@ -1,0 +1,51 @@
+# use-object-keys
+
+**Summary**: Prefer to use `object.keys`
+
+**Category**: Idiomatic
+
+**Avoid**
+```rego
+package policy
+
+keys := {k | some k, _ in input.object}
+
+# or
+
+keys := {k | some k; input.object[k]}
+```
+
+**Prefer**
+```rego
+package policy
+
+keys := object.keys(input.object)
+```
+
+## Rationale
+
+Instead of using a set comprehension to collect keys from an object, prefer to use the built-in function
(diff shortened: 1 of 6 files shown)


docs/rules/idiomatic/use-object-keys.md
```diff
@@ -0,0 +1,53 @@
+# use-object-keys
+
+**Summary**: Prefer to use `object.keys`
+
+**Category**: Idiomatic
+
+**Automatically fixable**: Yes/No
```
charlieegan3, line 7:
```suggestion
**Automatically fixable**: No
```

charlieegan3 (approved):
LGTM!

docs/rules/idiomatic/use-object-keys.md
anderseknert: Thanks! I removed that entirely as we don't normally show this when fixing is unsupported. Adding a fix for this seems simple though, so might be that I look into that next.

Issue #93, opened 2023-04-16
Rule: Detect expressions that may be moved outside of loop


anderseknert:
I've seen a few policies where expressions are evaluated in each iteration of a loop even though they don't make use of values bound in the loop, i.e.
```rego
allow {
    some user in data.users
    endswith(input.email, "acmecorp.com") # this should be moved out of the loop
    # more things here
}
```
Warning on this would be great.

PR #1401, opened 2025-02-06
bundle: Add non-loop-expression rule

charlieegan3:
Fixes https://github.com/StyraInc/regal/issues/93

TODO:

- [x] Docs for the rule
- [x] Optimizations on the rego / checking performance
- [x] ignore `print` statements

### DIFF
--- a/bundle/regal/rules/performance/non-loop-expression/non_loop_expression.rego
+++ b/bundle/regal/rules/performance/non-loop-expression/non_loop_expression.rego
@@ -1,0 +1,100 @@
+# METADATA
+# description: Non loop expression in loop
+package regal.rules.performance["non-loop-expression"]
+
+import data.regal.ast
+import data.regal.result
+import data.regal.util
+
+_exprs[sprintf("%d", [rule_index])][loc.row] contains expr if {
+	some rule_index, rule in input.rules
+	some expr in rule.body
+	loc := util.to_location_object(expr.location)
+}
+
+_exprs[sprintf("%d", [rule_index])][loc.row] contains expr if {
+	some rule_index, rule in input.rules
+	some rule_expr in rule.body
+	some expr in rule_expr.terms.body
+	loc := util.to_location_object(expr.location)
+}
+
+_loop_start_points[rule_index][loc.row] contains var if {
+	some rule_index
+	some var in ast.found.vars[rule_index].assign
+	some expr in _exprs[rule_index][_]
+	some term in expr
+	ast.is_wildcard(regal.last(regal.last(term).value))
+	loc := util.to_location_object(var.location)
+	loc.row == util.to_location_object(term[0].location).row
+	# no need to ignore vars here in comprehensions, since we are only looking
+	# for top level wildcards in the final term.
+}
+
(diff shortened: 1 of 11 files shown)


bundle/regal/rules/performance/non-loop-expression/non_loop_expression.rego
```diff
@@ -0,0 +1,107 @@
+# METADATA
+# description: Non loop expression in loop
```
anderseknert, line 2: Should it be "Non-loop expression"?

bundle/regal/rules/performance/non-loop-expression/non_loop_expression_test.rego
```diff
@@ -0,0 +1,389 @@
...
+allow if {
+	some k, v in input.foo
+	k == 1
+	some value in v
+	v == 2
+}
+`)
+
+	sps == {"0": {
+		5: {
```
anderseknert, line 32: Interesting how the first level variable can't be numeric (due to that type checker bug), but seems to work for this next level. That should be fixed of course, but certainly out of scope for this PR :)

bundle/regal/rules/performance/non-loop-expression/non_loop_expression_test.rego
```diff
@@ -0,0 +1,389 @@
...
+test_every if {
+	r := rule.report with input as ast.policy(`
+allow if {
+    every role in data.required_roles {
+        endswith(input.email, "acmecorp.com")
+        role in input.roles
+    }
+}`)
+
+	r == {{
```
anderseknert, line 369: We're repeating these large objects frequently here. Might want to look at how we use helpers in other tests to only pass the values that change between tests.

docs/rules/performance/non-loop-expression.md
```diff
@@ -0,0 +1,99 @@
...
+'Loops' in Rego refers to anywhere a rule branches, for example:
+
+- `some foo, bar in data.baz`
+- `foo := data.baz[_]` (prefer using `some`)
+- `walk(data.baz, [path, value])`
+- ...
+
+## Exceptions
+
+This rule cannot yet detect the following cases.
```
anderseknert, line 46: Great!

anderseknert (approved):
Awesome work here! Left a couple of comments but nothing blocking.

bundle/regal/rules/performance/non-loop-expression/non_loop_expression_test.rego
anderseknert: I'm hoping that @johanfylling can fix that in OPA soon :)

bundle/regal/rules/performance/non-loop-expression/non_loop_expression_test.rego
charlieegan3: I've used a with_location helper like we have in some other rules.

bundle/regal/rules/performance/non-loop-expression/non_loop_expression.rego
charlieegan3: Updated!

PR #1407, opened 2025-02-11
lsp: Correctly parse eval and debug params

charlieegan3:
These had been missed in https://github.com/StyraInc/regal/pull/1382


anderseknert (approved):
Very nice!

PR #1409, opened 2025-02-12
docs: Use strings for non-loop expression role

charlieegan3:
This example role is clearer as a string rather than some random var.


anderseknert (approved):
Agreed!

PR #1410, opened 2025-02-12
Bump Go to 1.23, and replace archived dependency

anderseknert:
Although tbh, I'm not sure why we rely on the mapstructure dependency in the first place. It's only used in a single location, and it looks like we could easily do what it does there without it..

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -63,7 +63,7 @@ jobs:
       - uses: golangci/golangci-lint-action@2226d7cb06a077cd73e56eedd38eecad18e5d837 # v6.5.0
         if: matrix.os.name == 'linux'
         with:
-          version: v1.63.4
+          version: v1.64.5
       - uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
         with:
           name: regal-${{ matrix.os.name }}
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -20,7 +20,7 @@ linters:
     - ireturn
     - funlen
     - gomoddirectives # need replacements for wasip1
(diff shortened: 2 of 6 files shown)


charlieegan3:
At the same time, there are some places where we might be able to unmarshal our rego results into structs with it?

e.g. https://github.com/StyraInc/regal/blob/dcecdabf079e7bbcb954a9dd8047c7092bc17a04/pkg/linter/linter.go#L545-L568 or https://github.com/StyraInc/regal/blob/dcecdabf079e7bbcb954a9dd8047c7092bc17a04/internal/lsp/completions/providers/policy.go#L102 ?

anderseknert:
What would the benefit of that be compared to the JSON roundtrip option? Not that I particularly like _that_ option either, but... Go 🤷 

As for the Rego result, I think I'd prefer if we avoided that to be turned into a map in the first place! We have a chance to turn the term pointer directly into the result if we provide a custom implementation via https://pkg.go.dev/github.com/open-policy-agent/opa@v1.1.0/v1/rego#GenerateJSON where we would _not_ generate JSON :)

charlieegan3:
yeah that's fair enough, there are better ways than both mapstructure and what we have.

internal/lsp/examples/index.json
```diff
@@ -1,22 +0,0 @@
-{
```
charlieegan3, line 1: This file is automatically updated by this workflow: https://github.com/StyraInc/regal/blob/main/.github/workflows/update-example-index.yaml so that when new docs are added, we link to them automatically from Regal.

anderseknert: Ah, I had missed that! Will revert and switch to use a load once function. Thanks!

charlieegan3 (approved):

Issue #1415, opened 2025-02-18
Root bundle folders in Docker break folder ignore paths


nevumx:
Mounting a bundle folder on the root of a regal docker container breaks ignore paths that are folder-wide.

For example, running regal in docker like:
```bash
docker run \
    --rm \
    --mount "type=bind,src=${PWD}/policies,dst=/policies,readonly" \
    --mount "type=bind,src=${PWD}/.regal,dst=/.regal,readonly" \
    ghcr.io/styrainc/regal:latest \
    lint \
    /policies
```
Will not allow folder rules like:
```yaml
rules:
  performance:
    non-loop-expression:
      ignore:
        files:
          - polices/role/*
```
but will allow file rules like:
```yaml
rules:
  performance:
    non-loop-expression:
      ignore:
        files:
          - some_file.rego
```
A workaround is to use a nonroot folder like:
```bash
docker run \
    --rm \
    --mount "type=bind,src=${PWD}/policies,dst=/home/nonroot/policies,readonly" \
    --mount "type=bind,src=${PWD}/.regal,dst=/home/nonroot/.regal,readonly" \
    ghcr.io/styrainc/regal:latest \
    lint \
    /home/nonroot/policies
```

PR #1418, opened 2025-02-24
lint: Handle root filesystem root prefix

charlieegan3:
When the root is also the root of the filesystem, the linter will  now handle ignore patterns correctly.

Fixes https://github.com/StyraInc/regal/issues/1415
Screenshot 2025-02-24 at 14 44 36


anderseknert (approved):
Nice!

PR #1419, opened 2025-02-25
lsp: URI/Path handling fixes

charlieegan3:
There was an issue where a colon in a file name would cause the wrong path to be generated from a URI and vice versa. This caused the server to reload the file every second as it was missing, but really it was just registered under the wrong URI in the cache.


internal/lsp/uri/uri.go
```diff
@@ -15,21 +16,36 @@ func FromPath(client clients.Identifier, path string) string {
 	path = strings.TrimPrefix(path, "file://")
 	path = strings.TrimPrefix(path, "/")
 
-	if client == clients.IdentifierVSCode {
-		// Convert Windows path separators to Unix separators
-		path = filepath.ToSlash(path)
+	drivePattern := regexp.MustCompile(`^([A-Za-z]):`)
```
anderseknert, line 19: Let's make this a package level var to avoid it being compiled with each call.

internal/lsp/uri/uri.go
```diff
@@ -48,10 +64,17 @@ func ToPath(client clients.Identifier, uri string) string {
 	}
 
 	if client == clients.IdentifierVSCode {
+		path = strings.TrimPrefix(path, "/")
 		// handling case for windows when the drive letter is set
-		if strings.Contains(path, ":") || strings.Contains(path, "%3A") {
-			path = strings.Replace(path, "%3A", ":", 1)
-			path = strings.TrimPrefix(path, "/")
+		drivePattern := regexp.MustCompile(`^([A-Za-z])(%3[aA]|:)`)
```
anderseknert, line 69: Same.

anderseknert (approved):
Thanks! LGTM with two suggested fixes.

charlieegan3:
Thanks, that makes sense! Have fixed now.

PR #1427, opened 2025-03-10
lsp: Correct path when fixing dir package mismatch

charlieegan3:
This had not been updated to the new arg format causing it to look up the wrong file path.

Also only log the use of the global file if a file is found.


anderseknert (approved):

anderseknert:
We should really have some tests for these controllers! But obviously, this fixes something broken, so let's get it in first.

PR #1429, opened 2025-03-10
OPA v1.2.0

anderseknert:
Also:
- Use oracle from OPA instead of vendored copy
- Use shared BaseCache for `regal lint`, which is now an option OPA provides
- Use `go tool` for `rq` dependency. Not `golangci-lint` for the time being, as it introduces a million indirect dependencies.
- Remove listed requirement of `gci` and `gofumpt` as these are both included in `golangci-lint` and don't need to be installed separately.
- Try out parameterized tests! Very nice, but will have to wait until later for a more complete transition to those. Still commiting the few tests I converted. Also filed: https://github.com/open-policy-agent/opa/issues/7407
- Update build/capabilities.json

Notes
- `rq` is still installed as a binary in GH actions. We should fix that as a next step.


anderseknert:
Backing out of parameterized tests until the next OPA release as we're running into issues with unmarshaling, and traces being all over the JSON output format. @johanfylling is taking a look on this from the OPA side.

charlieegan3 (approved):
LGTM!

anderseknert:
Tracing being in the JSON output format was a Regal mistake, not one from OPA. That filter was added in 2023 in OPA and our code was forked before that, so there was the issue.

PR #1430, opened 2025-03-10
build: Make novelty package opt-in

jalseth:
This allows Regal to be built without the go-asciisprite dependency.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -18,6 +18,7 @@ builds:
         goarch: arm64
     tags:
       - regal_standalone
+      - regal_enable_novelty
     ldflags:
       - -s -w
       - -X github.com/styrainc/regal/pkg/version.Version={{ .Version }}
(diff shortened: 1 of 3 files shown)


srenatus:
I'd like to understand the problem here. Is there anything at odds with the go-asciisprite dependency?

jalseth:
For internal reasons, it is more work to have more dependencies, so I was hoping to remove this as a hard dependency for the tool. I spoke a bit with @anderseknert offline and and it sounds like the preference is to only have behavior like this as opt-out rather than opt-in, which doesn't help my use case. Going to close this for now.

anderseknert:
Eh, change of heart. If this makes things a little easier for you @jalseth , we can live with the fact that `go build` won't give you the festive output format. It's a secret easter egg anyway, and this way it'll be even more secret 😎

charlieegan3 (approved):

Issue #1431, opened 2025-03-11
lsp: Bundle root detection causes bad renames when root is parent


charlieegan3:
When working in /foo/bar as the workspace root. And there is a config in /foo, the server will find /foo as the root. This is mostly fine, but causes issues when renaming files relative to this root as it means files can be moved outside the workspace.

While this is not a common way to use Regal LS, we need to protect the alarming behaviour where newly created files are teleported to the parent dir in such cases.

PR #1432, opened 2025-03-11
lsp: Ensure rename target is within workspace root

charlieegan3:
Fixes https://github.com/StyraInc/regal/issues/1431


anderseknert (approved):
Thanks!

PR #1433, opened 2025-03-12
test: fix deadline exceeded test issue in GitHub runners

anderseknert:
This was caused by a single test that by the way it evaluated rules would have all the tests from the bundle evaluated inside of its body!

This simple fix takes evaluation time of that test down frm 800 ms on my machine down to 4 ms.


srenatus (approved):
Good return for that little a change 😎

PR #1438, opened 2025-03-13
Fix non-loop-expression false positive

### LINKED ISSUE #1436: Incorrect non-loop-expression violation?
I have a rule that is checking that an HTTP requests matches a parameterized path (eg `/accounts/:id`). The (simplified) rule looks like

```rego
package service.authz

import data.http_request.groups
import data.http_request.path

allow if {
	path = ["accounts", _]
	groups.admins
}
```

and regal flags
```
Rule:         	non-loop-expression                                                                           	
Description:  	Non-loop expression                                                                           	
Category:     	performance                                                                                   	
Location:     	services/ui-storage.rego:8:2                                                                  	
Text:         	groups.admins                                                                                 	
Documentation:	https://docs.styra.com/regal/rules/performance/non-loop-expression   
```

As far as I'm aware (though please correct me if I'm wrong) this is not doing loop iteration, but is just unifying `path` against a tuple where the second element is a wildcard.

It's worth calling out that
```rego
package service.authz

import data.http_request.path

allow if {
	path = ["accounts", _]
}
```
does not flag the violation, presumably because the rule sees there is nothing after the use of the wildcard.

Apologies if this is a correct flag and I'm misunderstanding the mechanics of how this rule is actually evaluated!

anderseknert:
Thanks @drewcorlin1 for reporting the issue1

Fixes #1436

### DIFF
--- a/bundle/regal/rules/performance/non-loop-expression/non_loop_expression.rego
+++ b/bundle/regal/rules/performance/non-loop-expression/non_loop_expression.rego
@@ -43,12 +43,18 @@ _exprs[sprintf("%d", [rule_index])][row] contains expr if {
 	row := to_number(substring(expr.location, 0, indexof(expr.location, ":")))
 }
 
+# cover iteration in the form of:
+# x := foo.bar[_]
+# x = foo.bar[_]
 _loop_start_points[rule_index][loc.row] contains var if {
 	some rule_index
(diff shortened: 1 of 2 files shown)


charlieegan3 (commented):
LGTM!

charlieegan3 (approved):

PR #1457, opened 2025-03-25
Rule: time-now-ns-twice

### LINKED ISSUE #1456: Rule: time-now-ns-twice
Every once in a while, developers new to Rego get confused about `time.now_ns()` not working as they thought it would:

```rego
allow if {
    start := time.now_ns()

    # do something

    print(time.now_ns() - start)
}
```

This will of course print `0` no matter how much work is done in between the calls, as the "now value" is set once when evaluation is initiated. A Regal rule to flag multiple calls to `time.now_ns` in a body would be a good way to help these users understand what's going on, and why that is so. It would also be a good opportunity to link to relevant resources about how actually measure performance.

Not sure if there are even more effective ways to detect accidental "misuse" of the function than check for it being called twice ... let me know if you can think of any!

anderseknert:
Fixes #1456


srenatus (approved):
Nice extensive rationale!

PR #1469, opened 2025-04-07
Add benchmark for testing all rules individually

anderseknert:
This is currently very expensive/slow, but it's still a good benchmark to keep around for knowing the cost of some rule relative to others.


charlieegan3 (approved):
Sounds good to me!

PR #1472, opened 2025-04-08
perf: various improvements

anderseknert:
This started out as simply improving the performance of ref comparisons, where we should avoid stringifying if possible. But eventually led to a series of smallish optimizations, mostly related to iteration, and to keep the scope of _what_ is iterated down to smaller components. This saves about 300k allocations in `regal lint bundle`.

### DIFF
--- a/bundle/regal/rules/bugs/constant-condition/constant_condition.rego
+++ b/bundle/regal/rules/bugs/constant-condition/constant_condition.rego
@@ -9,36 +9,31 @@ import data.regal.result
 # Additionally, there are several other conditions that could be considered
 # constant, or if not, redundant... so this rule should be expanded in time
 
-_operators := {"equal", "gt", "gte", "lt", "lte", "neq"}
-
-_rules_with_bodies := [rule |
-	some rule in input.rules
-	rule.body
-]
-
 # METADATA
 # description: single scalar value, like a lone `true` inside a rule body
 # scope: rule
 report contains violation if {
-	expr := _rules_with_bodies[_].body[_]
+	terms := input.rules[_].body[_].terms
 
 	# We could probably include arrays and objects too, as a single compound value
 	# is not very useful, but it's not as clear cut as scalars, as you could have
 	# something like {"a": foo(input.x) == "bar"} which is not a constant condition,
 	# however meaningless it may be. Maybe consider for another rule?
-	expr.terms.type in ast.scalar_types
+	terms.type in ast.scalar_types
 
-	violation := result.fail(rego.metadata.chain(), result.location(expr.terms))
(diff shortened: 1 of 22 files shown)


bundle/regal/rules/bugs/constant-condition/constant_condition.rego
```diff
@@ -9,36 +9,31 @@ import data.regal.result
...
 }
 
 # METADATA
 # description: two scalar values with a "boolean operator" between, like 1 == 1, or 2 > 1
 # scope: rule
 report contains violation if {
-	expr := _rules_with_bodies[_].body[_]
+	operators := {"equal", "gt", "gte", "lt", "lte", "neq"}
+
+	expr := input.rules[_].body[_]
```
srenatus, line 33: Hm! I'd have expected the linter to scream "use 'some .. in ..' here"

pkg/linter/linter_test.go
```diff
@@ -727,8 +727,8 @@ import data.unresolved`,
 	}
 }
 
-// 1133583042 ns/op	3185226192 B/op	61392363 allocs/op
-// 1111341875 ns/op	3181965048 B/op	61390130 allocs/op
+// 1173608750 ns/op	3293593016 B/op	63366367 allocs/op
+// 1157190875 ns/op	3285082192 B/op	63014329 allocs/op
```
srenatus, line 731: What is that? 🔍

srenatus: You can use gobenchdata to block PRs  that spoil your performance too much, you know? 😉

srenatus (approved):

pkg/linter/linter_test.go
anderseknert: That would prevent us from adding new rules ;) Which is what these numbers reflect. New rules are "expensive" as even a simple one will scan 245 files or so..

bundle/regal/rules/bugs/constant-condition/constant_condition.rego
anderseknert: There's an exception made for "nested" iteration. It can even be configured for those that want it 🙂

pkg/linter/linter_test.go
anderseknert: The recent improvements in OPA provides plenty of headroom for new rules to be added until we reach previous levels with less rules! But ultimately, I don't think we can continue to add, say 100 more rules without changing something more fundamental in the model, like "bundling" similar rules together to avoid each one starting from scratch, etc.... But a change like that most likely requires further improvements in OPA too. Like how we can't really use custom functions in hot paths (like `walk`) currently without a hefty penalty.

PR #1485, opened 2025-04-14
Fix unnecessary-some rule not working after OPA 1.0+

### LINKED ISSUE #1483: No violation reported for `unnecessary-some` on OPA 1.0+
This would previously have been flagged, but it looks like that's no longer the case for 1.0+ policies.

```rego
fail if {
    some 1 in [1]
}
```

Looking at the code, it looks like we're only considering imported keywords, and will need to consider 1.0 keywords always being available here. Luckily it doesn't seem like other rules share this deficiency.

anderseknert:
And a few tweaks to the import checking functions.

Fixes #1483

### DIFF
--- a/bundle/regal/ast/imports.rego
+++ b/bundle/regal/ast/imports.rego
@@ -1,5 +1,7 @@
 package regal.ast
 
+import data.regal.capabilities
+
 default imports := []
 
 # METADATA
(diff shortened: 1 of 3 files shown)


bundle/regal/ast/imports.rego
```diff
@@ -46,26 +48,31 @@ resolved_imports[identifier] := path if {
 #   returns true if provided path (like ["data", "foo", "bar"]) is in the
 #   list of imports (which is commonly ast.imports)
 imports_has_path(imports, path) if {
-	some imp in imports
+	pv := imports[_].path.value
+
+	count(pv) == count(path)
 
-	_arr(imp) == path
+	every i, part in path {
```
charlieegan3, line 55: Does this run any faster/slower than building an array and using `==`?

anderseknert: Potentially, as `every` can benefit from early exit. But it's mostly custom functions that is a win to avoid when they're not strictly needed. 

See https://github.com/open-policy-agent/opa/issues/7266

anderseknert: (although this is not a hot path.. I just thought the change made the code nicer :)

PR #1486, opened 2025-04-15
create new rule to detect if there is a reference to a rule that does…

HookFirebolt:
… not exist

### DIFF
--- a/bundle/regal/config/provided/data.yaml
+++ b/bundle/regal/config/provided/data.yaml
@@ -1,6 +1,4 @@
-features:
-  remote:
-    check-version: true
+capabilities_url: regal:///capabilities/default
 rules:
   bugs:
     annotation-without-metadata:
@@ -125,6 +123,8 @@ rules:
     redundant-alias:
       level: error
     redundant-data-import:
+      level: error
+    unknown-reference:
       level: error
     unresolved-import:
       level: error
--- a/bundle/regal/rules/imports/unknown-reference/unknown_reference.rego
+++ b/bundle/regal/rules/imports/unknown-reference/unknown_reference.rego
@@ -1,0 +1,109 @@
+# METADATA
+# description: Invalid reference to import.
+package regal.rules.imports["unknown-reference"]
+
+import data.regal.ast
+import data.regal.config
+import data.regal.result
+import data.regal.util
+
+aggregate contains content if {
+	exported_rules := object.keys(ast.rule_head_locations)
+
+	refs := {ref |
+		_ref := ast.found.refs[_][_].value
+		_name := ast.ref_static_to_string(_ref)
+		not _name in ast.builtin_names
+		ref := {
+			"name": _name,
+			"path": split(_name, "."),
+			"location": _ref[0].location,
+		}
+	}
+
+	import_aliases := {alias_key: string_value |
+		some alias_key, value in ast.resolved_imports
+		string_value := concat(".", value)
+	}
+
+	data_refs := {{"full_name": ref.name, "ref": ref} |
+		some ref in refs
+		ref.path[0] == "data"
+	}
+
+	aliased_refs := {{"full_name": expanded_ref, "ref": ref} |
+		some ref in refs
+		full_source_prefix := import_aliases[ref.path[0]]
+
+		full_path_array := array.concat([full_source_prefix], util.rest(ref.path))
+
+		expanded_ref := concat(".", full_path_array)
+	}
+
+	# Combine 'aliased_refs' with 'data_refs'.
+	# Make it a map from full path to locations this endpoint is used for performance
+	all_refs_full_path := {ref.full_name |
+		some ref in (data_refs | aliased_refs)
+	}
+
+	all_full_path_refs := {full_name: _refs |
+		some full_name in all_refs_full_path
+		_refs := {ref.ref |
+			some ref in (data_refs | aliased_refs)
+			ref.full_name == full_name
+		}
+	}
+
+	content := result.aggregate(rego.metadata.chain(), {
+		"exported_rules": exported_rules,
+		"expanded_refs": all_full_path_refs,
+	})
+}
+
+aggregate_report contains violation if {
+	all_exports_in_bundle := {export |
+		some entry in input.aggregate
+		some export in entry.aggregate_data.exported_rules
+	}
+
+	some entry in input.aggregate
+	some ref_full_name, ref_locations in entry.aggregate_data.expanded_refs
+
+	# TODO: is pre-fix tree a posibility here?
+	# rough calculation of running it a few times in terminal, this loop is almost 50% of the runtime
+	every endpoint in all_exports_in_bundle {
+		not is_valid_ref(ref_full_name, endpoint)
+	}
+
+	some ref in ref_locations
+
+	# todo: this should probably be done with result.location but it doesnt work
+	loc := {"location": {
+		"file": sprintf("%s:%s", [entry.aggregate_source.file, ref.location]),
+		"text": sprintf("%s (%s) does not exist", [ref.name, ref_full_name]),
+	}}
+
+	violation := result.fail(rego.metadata.chain(), loc)
+}
+
+# METADATA
+# description: a reference is valid with respect to a rule if
+# 1: it is the prefix of a rule
+# 2: it indexes into a rule - we do not consider the possible data
+# 3: the reference is ignored in the config
+default is_valid_ref(_, _) := false
+
+is_valid_ref(ref_full_name, rule) if {
+	startswith(rule, ref_full_name)
+}
+
+is_valid_ref(ref_full_name, rule) if {
+	startswith(ref_full_name, rule)
+}
+
+is_valid_ref(ref_full_name, _) if {
+	cfg := config.for_rule("imports", "unknown-ref")
+
+	some exception in cfg["except-imports"]
+	glob.match(exception, [], ref_full_name)
+}
--- a/README.md
+++ b/README.md
@@ -261,6 +261,7 @@ The following rules are currently available:
 | imports     | [prefer-package-imports](https://docs.styra.com/regal/rules/imports/prefer-package-imports)             | Prefer importing packages over rules                      |
 | imports     | [redundant-alias](https://docs.styra.com/regal/rules/imports/redundant-alias)                           | Redundant alias                                           |
 | imports     | [redundant-data-import](https://docs.styra.com/regal/rules/imports/redundant-data-import)               | Redundant import of data                                  |
+| imports     | [unknown-reference](https://docs.styra.com/regal/rules/imports/unknown-reference)                       | Add description of rule here!                             |
 | imports     | [unresolved-import](https://docs.styra.com/regal/rules/imports/unresolved-import)                       | Unresolved import                                         |
 | imports     | [use-rego-v1](https://docs.styra.com/regal/rules/imports/use-rego-v1)                                   | Use `import rego.v1`                                      |
 | performance | [defer-assignment](https://docs.styra.com/regal/rules/performance/defer-assignment)                     | Assignment can be deferred                                |
--- a/bundle/regal/rules/imports/unknown-reference/unknown_reference_test.rego
+++ b/bundle/regal/rules/imports/unknown-reference/unknown_reference_test.rego
@@ -1,0 +1,33 @@
+package regal.rules.imports["unknown-reference_test"]
+
+import data.regal.ast
+import data.regal.config
+
+import data.regal.rules.imports["unknown-reference"] as rule
+
+# Example test, replace with your own
+# test_rule_named_foo_not_allowed if {
+# 	module := ast.policy("foo := true")
+# 	r := rule.report with input as module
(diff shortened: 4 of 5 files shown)


bundle/regal/rules/imports/unknown-reference/unknown_reference.rego
```diff
@@ -0,0 +1,109 @@
...
+			"name": _name,
+			"path": split(_name, "."),
+			"location": _ref[0].location,
+		}
+	}
+
+	import_aliases := {alias_key: string_value |
+		some alias_key, value in ast.resolved_imports
+		string_value := concat(".", value)
+	}
```
anderseknert, line 27: Neither `refs` and `import_aliases` depends on anything else in this rule, so I think making them set/map rules instead of comprehensions would be a good idea here, as the `aggregate rule is pretty long.

bundle/regal/rules/imports/unknown-reference/unknown_reference.rego
```diff
@@ -0,0 +1,109 @@
...
+	all_exports_in_bundle := {export |
+		some entry in input.aggregate
+		some export in entry.aggregate_data.exported_rules
+	}
+
+	some entry in input.aggregate
+	some ref_full_name, ref_locations in entry.aggregate_data.expanded_refs
+
+	# TODO: is pre-fix tree a posibility here?
+	# rough calculation of running it a few times in terminal, this loop is almost 50% of the runtime
```
anderseknert, line 73: That's a great observation 👍  And one we can save until later :) let's make it all work first.

bundle/regal/rules/imports/unknown-reference/unknown_reference.rego
```diff
@@ -0,0 +1,109 @@
...
+	every endpoint in all_exports_in_bundle {
+		not is_valid_ref(ref_full_name, endpoint)
+	}
+
+	some ref in ref_locations
+
+	# todo: this should probably be done with result.location but it doesnt work
```
anderseknert, line 80: Yeah, that's surprising. What did you pass into result.location? `ref`?
Should be fairly easy to replicate in a test.

bundle/regal/rules/imports/unknown-reference/unknown_reference.rego
```diff
@@ -0,0 +1,109 @@
...
+
+	violation := result.fail(rego.metadata.chain(), loc)
+}
+
+# METADATA
+# description: a reference is valid with respect to a rule if
+# 1: it is the prefix of a rule
+# 2: it indexes into a rule - we do not consider the possible data
+# 3: the reference is ignored in the config
+default is_valid_ref(_, _) := false
```
anderseknert, line 94: Function looks fine, but `is_known_ref` would be a better name, as that matches the name of this rule.

anderseknert (commented):
Awesome start! 

A few unit tests would be a great next step. Well done 👍

anderseknert:
I know this is just a start, but I guess we will need to consider local refs as well, i.e.

```rego
function(obj) if {
    obj.foo.bar == 4
}
```

Where obviously, `obj` "resolves" to the argument, and so on. 
You really jumped straight into the deep end of the pool with this rule 😆 

There's some code dealing with refs in the LSP policies too. Might be worth a look to see if there's something there that's already done, and may be reused between both.

bundle/regal/rules/imports/unknown-reference/unknown_reference.rego
HookFirebolt: Once refs and import_aliases were extracted, all_full_path_refs could also be extracted while removing data_refs, aliased_refs and all_refs_full_path.

HookFirebolt:
> ```rego
> obj
> ```
Just to make sure that we understand, what you mean with local refs. Are you thinking that there will be a false positive in that example, or are you saying that we should resolve `obj` and analyze all static parameters that `function` is called with?

As we have written it today, it only complains if there is a reference to external data (directly or indirectly starting with `data`) and  and does not resolve functions.
It also does not resolve for loops `some/every` to see if there is a static solution but I also think that would be overkill for the linter.

But unfolding static function calls might be within the scope?

bundle/regal/rules/imports/unknown-reference/unknown_reference.rego
anderseknert: Great!

bundle/regal/rules/imports/unknown-reference/unknown_reference.rego
anderseknert: Excellent!

anderseknert:
@HookFirebolt I guess what I mean would be to deal with something like this:

```rego
import data.foo.obj

function(obj) if {
    obj.foo.bar == 4
}
```

And knowing that `obj.foo.bar` still resolves in its context even if the import doesn't. But yeah, agreed with starting with the basics. If we miss a few cases, that's alright. We just need to avoid false positives.

anderseknert:
@HookFirebolt for sure — there are limits to what static analysis can achieve, and using only Rego even more so. So yeah, aim to make your first version work for the most common cases. Rules can always be improved later, and in fact they are all the time.

bundle/regal/config/provided/data.yaml
```diff
@@ -1,6 +1,4 @@
-features:
-  remote:
-    check-version: true
+capabilities_url: regal:///capabilities/default
```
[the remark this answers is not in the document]
anderseknert: Created https://github.com/StyraInc/regal/issues/1497 for tracking

bundle/regal/config/provided/data.yaml
```diff
@@ -128,6 +128,14 @@ rules:
...
     redundant-data-import:
       level: error
+    unresolved-reference:
+      level: error
+      except-paths:
+        - data.eval.**
+        - data.internal**
+        - data.workspace.**
+        - data.regal.**
+        - data.custom.regal.**
```
anderseknert, line 138: So, is there any node *not* ignored? :)

bundle/regal/rules/imports/unresolved-reference/unresolved_reference.rego
```diff
@@ -0,0 +1,122 @@
+# METADATA
+# description: Reference to unresolved field.
```
anderseknert, line 2: "field" isn't commonly used in Rego terminology. I'd just say "Unresolved reference" I think, but open to suggestions!

bundle/regal/rules/imports/unresolved-reference/unresolved_reference.rego
```diff
@@ -0,0 +1,122 @@
...
+}
+
+# an import is shadowed if it shares name with a variable (or function argument)
+_shadowed_imports contains var_name if {
+	var_name := ast.found.vars[_][_][_].value
+	_import_aliases[var_name]
+}
+
+_refs contains ref if {
+	_ref := ast.found.refs[_][_].value
```
anderseknert, line 39: No need to use underscore prefixes for local vars. You could call this `terms` if you want to use `ref` later.

`_name` could just be called `name`

bundle/regal/rules/imports/unresolved-reference/unresolved_reference.rego
```diff
@@ -0,0 +1,122 @@
...
+	violation := result.fail(rego.metadata.chain(), result.location(ref))
+}
+
+# METADATA
+# description: a reference is valid with respect to a rule if
+# custom:
+#   1: it is the prefix of a rule
+#   2: it indexes into a rule - we do not consider the possible data
+#   3: the reference is ignored in the config
+default _is_resolved_ref(_, _, _) := false
```
anderseknert, line 95: Nit, but I don't think you need this default assignment, as undefined works just as well for negation.

docs/rules/imports/unresolved-reference.md
```diff
@@ -0,0 +1,53 @@
+# unresolved-reference
+
+**Summary**: Unresolved Reference
+
+**Category**: Imports
+
+**Automatically fixable**: No
```
anderseknert, line 7: We normally only include this for fixable rules, so you can delete this line.

docs/rules/imports/unresolved-reference.md
```diff
@@ -0,0 +1,53 @@
...
+```yaml
+rules:
+  imports:
+    unresolved-reference:
+      # one of "error", "warning", "ignore"
+      level: error
+      # list of paths that should be ignored
+      # these may be paths to data, or rules that may
+      # not be present at the time of linting
+      # using glob syntax
```
anderseknert, line 36: Is this consistent with unresolved-import? It would be nice if it was.

anderseknert (commented):
Looks amazing! Outstanding work 👏 Left a few comments for consideration, and I'll have to take this for a test drive too... but this is almost there 👍

bundle/regal/rules/imports/unresolved-reference/unresolved_reference.rego
anderseknert: Also, skip the trailing period.

docs/rules/imports/unresolved-reference.md
HookFirebolt: Yes, unresolved-import also uses glob.match. We just felt a need to add information about the notation to the documentation

bundle/regal/config/provided/data.yaml
HookFirebolt: I'm not quite sure what you mean by "Node". If you are asking if any node in the AST is not ignored, then yes. This is not a glob match on nodes in the AST but on a transformation (absolute path) of the "value" of references in the AST. 
This is a list of run-time data that is mocked in other rules' tests, e.g. https://github.com/StyraInc/regal/blob/main/bundle/regal/rules/bugs/rule-shadows-builtin/rule_shadows_builtin_test.rego#L9.
We could alternatively ignore them in the individual tests if that is preferable.

docs/rules/imports/unresolved-reference.md
anderseknert: Great 👍

bundle/regal/config/provided/data.yaml
HookFirebolt: It was intentional, but reflecting on it, it should probably not be in the default config file. Will remove it and add the in-line ignores where necessary

anderseknert: Fine to put it in the config for the Regal project itself, but then that'd be in `.regal/config.yaml` 
Placing it here would mean these paths are ignored for all users of Regal, as that's the base configuration on which any user configuration is merged on top of

HookFirebolt: Hadn't noticed the .regal/config.yaml file. It's moved now.

anderseknert: Excellent! Last question on this I think, but why exclude `- data.regal.**` etc from this check? Do we not want our own refs checked internally?

Either way, awesome work! I'll play around some with this feature in the weekend, and if all is well, we should be able to have this merged early next week 👏

bundle/regal/rules/imports/unresolved-reference/unresolved_reference.rego
```diff
@@ -0,0 +1,124 @@
...
+
+import data.regal.ast
+import data.regal.config
+import data.regal.result
+import data.regal.util
+
+# METADATA
+# description: collects exported and full of used refs from each module
+aggregate contains content if {
+	# regal ignore:unconditional-assignment
```
anderseknert, line 13: Why? Moving the assignment directly to the head correctly signals that their is no "if" in this rule.

```rego
aggregate contains result.aggregate(rego.metadata.chain(), {
    "exported_rules": object.keys(ast.rule_head_locations),
    "expanded_refs": _all_full_path_refs,
    "prefix_tree": _prefix_tree,
})
```

anderseknert (approved):
Sorry for keeping you waiting! Finally got the time to try this out, and it seems to be working well. You've done an amazing job on this rule, and like I said before, you really dived right into the deep end of the pool with this rule 😄 Impressive. I'd love to chat later about your process developing this, and what you think we can do to improve that!

I've left a few comments, of which most are nits really, but you may want to consider them before this is merged.

This rule does a lot of processing and I think there are some optimizations that could help improve its performance. But it's well within the limits for a new rule of this type, so let's save that for later. 

Awesome work! 👏 

Let me know when you feel this is ready to be merged, and make sure to rebase on top of main at that point.

bundle/regal/rules/imports/unresolved-reference/unresolved_reference.rego
HookFirebolt: I think the code has slightly changed since we added the ignore, and back then we thought the assignment was more readable, but now this does seem better, so it has been removed.

HookFirebolt:
@anderseknert We have implemented your last suggestions, so we are ready to merge. Should we collapse the commits?

anderseknert:
@HookFirebolt awesome. I can do that as part of the merge, but if you do it you get to write the commit message :) So that's often better.

HookFirebolt:
@anderseknert done ;-)

PR #1487, opened 2025-04-16
FindConfigRoots: return error, not panic

srenatus:
Stumbled upon this when playing with the LSP and sending bad init parameters -- that caused `err` to be non-nil, and ran into a nil deref on `info.IsDir()` 💥 

Happy to add a test if you can point out a good spot 😄


anderseknert (approved):
Thanks!

PR #1488, opened 2025-04-16
Rule: narrow-argument

### LINKED ISSUE #1480: Rule: narrow-argument
This is mostly a matter of preference (right?), so I'm thinking this would be an optional rule in the `custom` category. The basic idea is to have a rule tell you when it's possible to "reduce" a value passed as a function argument to a more narrow type. For example:

```rego
internal_user(user) if endswith(user.email, "@acmecorp.com")
internal_user(user) if endswith(user.email, "@acmecorp.org")
```

Could be reduced to:

```rego
internal_user(email) if endswith(email, "@acmecorp.com")
internal_user(email) if endswith(email, "@acmecorp.org")
```

(at which point, renaming the function `internal_email` would likely make sense)

Or it could be an array that is always accessed by the same index within the function:

```rego
expensive_phone_call(list) if list[0].country_code == 61 # Australia
expensive_phone_call(list) if list[0].country_code == 81 # Japan
```

Which could be reduced to:

```rego
expensive_phone_call(country_code) if country_code == 61 # Australia
expensive_phone_call(country_code) if country_code == 81 # Japan
```

Or even [this](https://docs.styra.com/regal/rules/idiomatic/equals-pattern-matching):

```rego
expensive_phone_call(61) # Australia
expensive_phone_call(81) # Japan
```

Functions written this way are arguably easier to read, as they handle only the values of relevance, and it's often easier to predict what the function does from the call site when more narrow types are passed. Additionally, a function taking narrower arguments is more likely to be reusable, as e.g. an email address could be relevant to more than a user, etc.

Feedback much appreciated!

anderseknert:
This was a lot of fun! Check out the docs included in this PR for more information about this rule and potential caveats. But all things considered, I think this rule is great and have enabled it for our own use (with a few exceptions configured).

Even when narrowing isn't the best option, I think it's great to be forced to think about the dependencies of functions as I work on them.

Consistently named args help a lot here, as being able to exclude commonly passed values, like `cfg` in our case, "fixed" almost all the reported violations, and I only had to use a handful of ignore directives for the rest. But most of the issues reported were rewritten, and I think for the better!

(a few things changed here were not directly reported, but things I saw could be improved in the files I opened)

Fixes #1480


docs/rules/custom/narrow-argument.md
```diff
@@ -0,0 +1,187 @@
...
+
+### Reference prefix narrowing
+
+So far we have looked only at functions using an identical reference to one of its arguments. That's not always the
+case, but it doesn't mean the value passed can't be narrowed! Consider the following example:
+
+```rego
+package policy
+
+interal_user(context) if endswith(context.user.email, "@acmecorp.com")
```
charlieegan3, line 73:
```suggestion
internal_user(context) if endswith(context.user.email, "@acmecorp.com")
```

charlieegan3 (approved):
Looks good to me, caveats section is great!

PR #1494, opened 2025-04-21
docs: fix docs that still used v0 syntax

anderseknert:
Also a small few perf fixes I hadn't committed before. Small but
with quite an impact. ~2 million allocs!

```
1088929208 ns/op        3104908016 B/op 59995728 allocs/op
1070530708 ns/op        3041023912 B/op 58063258 allocs/op
```


charlieegan3 (approved):
LGTM!

PR #1496, opened 2025-04-21
Make external-reference rule less annoying

### LINKED ISSUE #1002: Move `external-reference` to `custom` category
This is by far the rule we've had to make the most exceptions for in Regal, and we've heard from a few others that this has been a tad too pedandtic. While users can easily disable the rule, I think it makes sense that those who want this will have to enable it instead. Move this rule to the `custom` category, and make sure all references are updated.

It would be good if we can somehow allow users who have explicitly enabled this in their configuration to have it remain working without changes. Let's see if that's possible.

anderseknert:
And configurable! Instead of flagging all external references in a function, we now allow setting a max number of allowed external references, and have changed the default tolerance from 0 to 2. I hope this will make the rule less distracting, while still help flag excessive undeclared dependencies.

As per usual, a few performance improvements related to code I was working on here also got included, and in particular the improvements to the rule itself made the biggest difference.

```
1145727583 ns/op	3184970896 B/op	61571557 allocs/op
1083952583 ns/op	3105174424 B/op	59992494 allocs/op
```

Fixes #1002

### DIFF
--- a/bundle/regal/ast/ast.rego
+++ b/bundle/regal/ast/ast.rego
@@ -37,10 +37,10 @@ operators := {
 # description: |
 #   returns true if provided term is either a scalar or a collection of ground values
 # scope: document
-is_constant(term) if term.type in scalar_types # regal ignore:external-reference
+is_constant(term) if term.type in scalar_types
 
 is_constant(term) if {
-	term.type in {"array", "object"}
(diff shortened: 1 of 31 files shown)


srenatus (approved):
Much appreciated!

PR #1500, opened 2025-04-23
refactor: continued work to improve config handling and tests

anderseknert:
- Check if files are ignored globally only once, and before proceeding to check if they are ignored by some specific rule's configuration
- Stop using config.for_rule("category", "title") now that we can refer directly to config.rules.category.title
- Stop using data.internal in tests

Obviously, there's gotta be some performance improvements in there somewhere too :P Luckily having the global ignore check done only once per file gave us that!

```
1070530708 ns/op	3041023912 B/op	58063258 allocs/op
1048394166 ns/op	2953960968 B/op	56616841 allocs/op
```
Now getting awfully close to sub-second times :)


charlieegan3 (approved):
Great work! 🤖

PR #1504, opened 2025-04-25
Scan nested expressions

### LINKED ISSUE #82: Rule: constant conditions
```rego
allow {
    1 == 1
    true
    false
}
```
And so on...

Note that we can't do much about a single `true` in an otherwise empty body, as:

```rego
allow := true
```

will be rewritten as:

```rego
allow = true {
    true
}
```
So we can accept that as an exception for now.

anderseknert:
A few of the first Regal rules only ever checked the top level of a body, and would not traverse into e.g. comprehensions or every bodies. This has been annoying for long, and while doing this properly is certainly more expensive (+2 million allocs) I think we can afford that at this point in order to prioritize correctness. May very well be that many will have Regal find "new" issues in the next version due to this.

We now also separate `ast.found.references` and `ast.found.calls` instead of collecting both under `ast.found.references`.  This too comes at a cost, but makes it much more obvious what our code is doing, and we can now afford prioritizing that too.

Fixes #82

### DIFF
--- a/.regal/config.yaml
+++ b/.regal/config.yaml
@@ -17,6 +17,7 @@ rules:
       level: error
       exclude-args:
         - cfg
+        - metadata
         - rule
   style:
     line-length:
--- a/bundle/regal/ast/ast.rego
+++ b/bundle/regal/ast/ast.rego
@@ -203,7 +203,7 @@ rule_index_strings := [sprintf("%d", [i]) | some i, _ in _rules]
 #   keyed by rule index
 function_calls[rule_index] contains call if {
 	some rule_index in rule_index_strings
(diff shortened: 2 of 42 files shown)


srenatus (approved):
The internal/lsp code moves seem unrelated, but OK.

anderseknert:
@srenatus yeah, sorry about that... the commit message is a little more detailed https://github.com/StyraInc/regal/pull/1504/commits/35eafec33c40e85b922ce6edb4e0424dd2b0320f

But even that did not cover everything here as I took the liberty of fixing things here and there that were not directly covered by existing issues.

Issue #359, opened 2023-09-30
Create new Rego setup phase preceeding lint evalutation


anderseknert:
I don't think it's a goal we're touting anywhere, but given the overarching goals of the project it should go without saying that if we can do something either in Go or Rego, we'll do it in Rego. There's currently a few things we either do in Go, or do both in Go and Rego, that could be done only in Rego:

* Parsing and merging of provided / user configuration
* Filtering of files / ignores
* ~~Parsing of input policies~~

Could be that the last step is too expensive to do in Rego (and `regal.parse_module`) but the two other steps would be _cheaper_ to do in a pre-lint step even in Rego, as we currently do at least some parts of config evaluation per input file. This is redundant, of course, as the config doesn't change between input files, and should really only be calculated once.

Another benefit of providing a pre-lint evaluation phase executed in Rego is that policy authors — similarly to when they provide custom rules — could provide custom logic for what to do before linting, and have whatever data collected in that step provided as part of the input when running the lint phase.

anderseknert:
I think we'll have to leave parsing out of this, even though I'd love to do that using `regal.parse_module`. The Go machinery for parsing parallelizes the process across all the input files, and doing this serially in Rego would likely be a too hard hit on initialization time. Anything related to configuration and overrides should still be done here though!

anderseknert:
Adding to the list — querying each rule for notices, and return back the result so that rules can be disabled conditionally based on those. This works today already, but for each linted file, a list like this is returned:

```
  "notices": [
    {
      "category": "bugs",
      "description": "Since OPA 1.0, deprecated-builtin enabled only when provided a v0 policy,\nBUT please note that this may change in the future if new built-in functions\nare deprecated.\n",
      "level": "notice",
      "severity": "none",
      "title": "deprecated-builtin"
    },
    {
      "category": "bugs",
      "description": "Since OPA 1.0, rule-named-if enabled only when provided a v0 policy",
      "level": "notice",
      "severity": "none",
      "title": "rule-named-if"
    },
    {
      "category": "idiomatic",
      "description": "Since OPA 1.0, use-contains enabled only when provided a v0 policy",
      "level": "notice",
      "severity": "none",
      "title": "use-contains"
    },
    {
      "category": "idiomatic",
      "description": "Since OPA 1.0, use-if enabled only when provided a v0 policy",
      "level": "notice",
      "severity": "none",
      "title": "use-if"
    },
    {
      "category": "imports",
[...]

PR #1506, opened 2025-04-28
Add Prepare method for reusing state across linter runs

### LINKED ISSUE #1394: API: allow separating preparation of linting from actual linting
While the various `WithX` methods help set up a configuration for linting, the `Lint` method is currently where all the processing takes place (parsing, preparing the query, linting, preparing the result).

It would be great if the API offered a way to allow preparing for linting and linting in two separate steps, allowing the more flexible/performant use in some scenarios. Another benefit would be benchmarking, where it would be much easier to differentiate costs of linting if the cost of preparing to lint could be accounted for separately from the cost of linting (i.e. the Rego evaluation).

Obviously, we'll still want the current way to keep working. So I'd imagine the `Lint()` call would first check if there's a prepare state to start from, and if not, itself call `Prepare()` (or whatever we want to call it)

```go
linter := NewLinter().
    WithInputPaths([]string{"../../bundle"}).
    WithDisableAll(true).
    WithEnabledRules("opa-fmt").
    Prepare()  // This would be new

rep, err = linter.Lint(context.Background())
if err != nil {
    panic(err)
}
```

anderseknert:
This paves the way for moving initialization logic to Rego later: https://github.com/StyraInc/regal/issues/359

As well as moving the `DetermineEnabledRules` and
`DetermineEnabledAggregateRules` methods to instead use the same query later on.

Fixes #1394

### DIFF
--- a/pkg/linter/linter.go
+++ b/pkg/linter/linter.go
@@ -4,9 +4,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io"
 	"io/fs"
 	"log"
+	"maps"
 	"os"
 	"path/filepath"
 	"slices"
@@ -67,9 +67,27 @@ type Linter struct {
 	profiling            bool
 	instrumentation      bool
 	hasCustomRules       bool
+	isPrepared           bool
+
+	preparedQuery *rego.PreparedEvalQuery
 }
 
-var lintQuery = ast.MustParseBody("lint := data.regal.main.lint")
+var (
+	lintQueryStr         = "lint := data.regal.main.lint"
+	enabledRulesQueryStr = `[rule |
+		data.regal.rules[cat][rule]
+		object.get(data.regal.rules[cat][rule], "notices", set()) == set()
+		not data.regal.config.ignored_rule(cat, rule)
+	]`
+	enabledAggregateRulesQueryStr = `[rule |
+		data.regal.rules[cat][rule].aggregate
+		not data.regal.config.ignored_rule(cat, rule)
+	]`
+
+	lintQuery                  = ast.MustParseBody(lintQueryStr)
+	enabledRulesQuery          = ast.MustParseBody(enabledRulesQueryStr)
+	enabledAggregateRulesQuery = ast.MustParseBody(enabledAggregateRulesQueryStr)
+)
 
 // NewLinter creates a new Regal linter.
 func NewLinter() Linter {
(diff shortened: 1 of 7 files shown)


pkg/linter/linter.go
```diff
@@ -67,9 +67,27 @@ type Linter struct {
...
 	hasCustomRules       bool
+	isPrepared           bool
+
+	preparedQuery *rego.PreparedEvalQuery
 }
 
-var lintQuery = ast.MustParseBody("lint := data.regal.main.lint")
+var (
+	lintQueryStr         = "lint := data.regal.main.lint"
+	enabledRulesQueryStr = `[rule |
```
srenatus, line 77: 💭 I like the momentum here, to do more stuff in Rego. But if it's inline query strings like that, it'll stay under the radar of our beloved linter itself. I suppose we could create some extra files and `go:embed` them...

pkg/linter/linter.go
```diff
@@ -67,9 +67,27 @@ type Linter struct {
 	profiling            bool
 	instrumentation      bool
 	hasCustomRules       bool
+	isPrepared           bool
```
srenatus, line 70: So a `Linter` here is not supposed to be used by multiple callers concurrently? Only a prepared Linter is?

If so, it could make sense to introduce another type for that, so we have
```go
func (l *Linter) Prepare(context.Context) (*PreparedLinter, error)
```

srenatus (approved):
Suggestions and questions only

pkg/linter/linter.go
anderseknert: Hmm, that's a great idea actually.. and I'd much prefer these to be in .rego files somewhere. I don't think we can parse Rego files without a package though? Or maybe we can, but Regal doesn't do that currently.

OTOH, I expect both of the "enabled rules" queries to be dropped shortly, as we consolidate on a single prepared query and use input attributes to determine the route.

pkg/linter/linter.go
anderseknert: You're one step ahead of me, as concurrent use isn't really something I've considered yet. All instances of Linter should be safe for concurrent use I think, as we're not using a pointer receiver, and all "writes" to the linter return a new instance. The difference would be that `Prepare` returns a new `Linter` while `Lint` returns a report.

`isPrepared` is mostly there to have something to track the state of the linter, and to ensure that while the same prepared linter can be reused for multiple inputs, changing something like configuration options should trigger a "reprepare". Does that make sense?

Perhaps making `Linter` and interface and have two different implementations (like the PreparedLinter you suggested) would be more elegant 🤔

Issue #1265, opened 2024-11-18
Remodel rule aggregation format


anderseknert:
Regal evaluates aggregate rules in a 2-step process:

1. Have all aggregate rules collect data they're interested in and send it back to the Go process
2. Have the Go process launch another Rego eval where all `aggregate_report` rules are queried with this data as input.

Each file evaluated by the aggregate rule will result in a data structure like this returned:

```json
{
  "organizational/at-least-one-allow": [
    {
      "aggregate_data": [{"aggregated": "data"}],
      "aggregate_source": {
        "file": "foo/bar.rego",
        "package_path": ["foo", "bar"]
      },
      "rule": {
        "category": "organizational",
        "title": "at-least-one-allow"
      }
    },
    {
       "one of these": "for each aggregated entry!"
    }
  ],
  "other/aggregate_rule" : []  
}
```

Each input file will then append a similar object to the array of any given rule. This data is _verbose_, and while it's easily readable, that's not a huge win here as this isn't meant to be consumed by humans in the first place. Running `regal lint` on its own bundle results in almost one megabyte of this data being sent back and forth, and we don't even have many aggregate rules! As recent work has shown, lots of data equals lots of allocations. Sometimes that's unavoidable, but this is one place where we could be optimizing more.

To make this more efficient, we should look into:

1. Drop the `rule` attribute from the payload. The outer map is already keyed by category/rule and repeating it in the payload is just a waste of resources.
2. Use an object keyed by filenames rather than an array duplicating data.
2. Have the aggregator rules append to the `aggregate_data` list for its file rather than the top level one.

```json
{
  "organizational/at-least-one-allow": {
    "foo/bar.rego": {
      "package_path": ["foo", "bar"],
      "aggregate_data": [{"aggregated": "data"}, {"more": "items"}]
    },
    "baz.rego": {}
  },
  "other/aggregate_rule" : {}  
}
```

As aggregate rules are also available to custom rule authors, this would be a breaking change, and one we'll need to document well how to go from the old format to the new, both on the aggregating side (where this would likely not require a change as long as they are using `result.aggregate` and we update that) and on the reporting side (where policies _would_ need to be updated to handled the new format).

While we are at it, let's make sure we also handle [empty aggregate results](https://github.com/StyraInc/regal/issues/1259) correctly and efficiently for both built-in and custom rules.

anderseknert:
[...]

PR #1530, opened 2025-05-04
perf: improve performance of new unresolved-reference rule

anderseknert:
We've had a few heavy rules in the past, but nothing like this one. That's not really due to the implementation of the rule, but simply the fact that it aggregates every ref found everywhere, which will be a lot no matter what.

Using the standard "regal linting itself" benchmark on main before the rule was merged and after shows a 10%+ increase in memory allocated as well as wall clock time:

```
1111071791 ns/op	3001557280 B/op	58184205 allocs/op
1289814875 ns/op	3363166528 B/op	64788506 allocs/op - unresolved-reference merged
```

There are [some design issues](https://github.com/StyraInc/regal/issues/1265) to address with aggregate rules that likely would mean bigger wins across the board, but the focus of this PR is solely to optimize the code for the rule itself.

While this remains an expensive rule, it's now closer to our other more costly rules. In total, this saves about 2.7 million allocations of the 6.7 million allocs increased with its addition. The steps in which this was done are roughly as follows:

```
1249527041 ns/op	3337275400 B/op	64261924 allocs/op - exclude input refs
1247609750 ns/op	3325728208 B/op	64061112 allocs/op - get rid of _import_aliases
1264200584 ns/op	3324900160 B/op	64037543 allocs/op - simplify _all_full_path_refs
1182358875 ns/op	3322581056 B/op	64013278 allocs/op - inline parts of _is_resolved_ref
1189268416 ns/op	3317109520 B/op	63942894 allocs/op - skip if name in ast.rule_and_function_names
1182895291 ns/op	3305238432 B/op	63739044 allocs/op - only aggregate data used in report
1179427375 ns/op	3273346464 B/op	63188538 allocs/op - custom result.location alternative
1161155834 ns/op	3210160768 B/op	62267845 allocs/op - defer to_location_object to report
1130275167 ns/op	3204268280 B/op	62150928 allocs/op - a few small tweaks
```

I have a few more ideas for improvements, but as they'd touch on more than just this rule, this is probably a good place to stop for now.

This was great for getting to think more about aggregate rules, and what we can do to improve them. Expect more on that soon.

### DIFF
--- a/bundle/regal/util/util.rego
+++ b/bundle/regal/util/util.rego
@@ -67,6 +67,10 @@ to_location_object(loc) := {
 
 to_location_object(loc) := loc if is_object(loc)
 
+# METADATA
+# description: efficiently extracts row number from location string
+to_location_row(loc) := to_number(regex.replace(loc, `^(\d+):.*`, "$1"))
+
 _location_to_text(row, col, end_row, end_col) := substring(
 	input.regal.file.lines[row - 1],
 	col - 1,
--- a/bundle/regal/ast/ast.rego
+++ b/bundle/regal/ast/ast.rego
@@ -145,7 +145,7 @@ function_arg_names(rule) := [arg.value | some arg in rule.head.args]
 
 # METADATA
 # description: all the rule and function names in the input AST
-rule_and_function_names contains ref_to_string(rule.head.ref) if some rule in input.rules
+rule_and_function_names contains ref_static_to_string(rule.head.ref) if some rule in input.rules
 
 # METADATA
 # description: all identifiers in the input AST (rule and function names, plus imported names)
@@ -153,7 +153,7 @@ identifiers := rule_and_function_names | imported_identifiers
 
 # METADATA
 # description: all rule names in the input AST (excluding functions)
-rule_names contains ref_to_string(rule.head.ref) if some rule in rules
+rule_names contains ref_static_to_string(rule.head.ref) if some rule in rules
 
 # METADATA
 # description: |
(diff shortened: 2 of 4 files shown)


bundle/regal/util/util.rego
```diff
@@ -67,6 +67,10 @@ to_location_object(loc) := {
 
 to_location_object(loc) := loc if is_object(loc)
 
+# METADATA
+# description: efficiently extracts row number from location string
+to_location_row(loc) := to_number(regex.replace(loc, `^(\d+):.*`, "$1"))
```
srenatus, line 72: That's efficient? Wow. I'd have expected a split on ":" to be more efficient. 💡

srenatus (approved):
LGTM

bundle/regal/util/util.rego
anderseknert: The speed is about the same (the ns/op values aren't too reliable), but it's 6 less allocations :)

```rego
r1 := to_number(substring(loc, 0, indexof(loc, ":")))
```

```
+-------------------------------------------+------------+
| samples                                   |     131254 |
| ns/op                                     |       8817 |
| B/op                                      |       7440 |
| allocs/op                                 |        129 |
```

```rego
r2 := to_number(split(loc, ":")[0])
```

```
+-------------------------------------------+------------+
| samples                                   |     138829 |
| ns/op                                     |       8030 |
| B/op                                      |       7191 |
| allocs/op                                 |        126 |
```

```rego
r3 := to_number(regex.replace(loc, `^(\d+):.*`, "$1"))
```

```
+-------------------------------------------+------------+
| samples                                   |     143132 |
| ns/op                                     |       8394 |
| B/op                                      |       7117 |
| allocs/op                                 |        120 |
```

PR #1545, opened 2025-05-13
Various small fixes

anderseknert:
- Change benchmarks so that they no longer enable all rules unconditionally, but read our config from the workspace. This makes the benchmark look "better" but that's only because they're no longer spending time reporting violations, and as such this is a better benchmark of Regal really linting itself.
- The benchmarks now additionally fails on any violations, which should ensure we're always measuring the right thing.
- Fix ref_static_to_string so that it correctly deals with values like numbers and booleans.
- Some microoptimizations in ast.found.xxx logic


pkg/config/config.go
```diff
@@ -125,6 +125,21 @@ const (
 	standaloneConfigFileName = ".regal.yaml"
 )
 
+func FromPath(path string) (Config, error) {
```
charlieegan3, line 128: Is there a way we might adjust FindConfig to first check if the path is a valid config file before searching?

anderseknert: But the path provided to FindConfig isn't a config file, but just any directory, no?

PR #1547, opened 2025-05-14
perf: various optimizations

anderseknert:
- My new favorite trick! Using `regex.replace` to extract values out from a text instead of `split`-ing a string on a delimiter. This is surprisingly *much* cheaper
- Use fixed max length of 100 in array.slice instead of `count`, where appropriate (like when slicing terms)
- Simplify and/or inline some custom functions

About 1.2 million allocations saved (see benchmark in PR for details)

### DIFF
--- a/bundle/regal/rules/bugs/top-level-iteration/top_level_iteration.rego
+++ b/bundle/regal/rules/bugs/top-level-iteration/top_level_iteration.rego
@@ -17,7 +17,7 @@ report contains violation if {
 		part.type == "var"
 	]) == 0
 
-	some part in array.slice(rule.head.value.value, 1, 128)
+	some part in array.slice(rule.head.value.value, 1, 100)
 
 	part.type == "var"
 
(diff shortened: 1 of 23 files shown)


charlieegan3 (approved):

bundle/regal/rules/bugs/top-level-iteration/top_level_iteration.rego
```diff
@@ -17,7 +17,7 @@ report contains violation if {
 		part.type == "var"
 	]) == 0
 
-	some part in array.slice(rule.head.value.value, 1, 128)
+	some part in array.slice(rule.head.value.value, 1, 100)
```
charlieegan3, line 20: would a constant somewhere be worthwhile here? so we can explain why this value was chosen?

anderseknert: This is a micro-optimization, and using a constant would mean we'd need to query the virtual cache in each request, which while reasonably fast is still a cost comparable to calling `count` 😄

anderseknert: (I totally agree in principle though!)

charlieegan3: Fair enough! 🙂

PR #1562, opened 2025-05-26
Various small fixes

anderseknert:
- Fix issue with determining Rego version for some paths
- Some Rego improvements


bundle/regal/lsp/completion/providers/locals/locals.rego
```diff
@@ -43,21 +43,15 @@ items contains item if {
...
-	strings.any_prefix_match(expr, {"some", "every"})
+_same_line_loop_vars(line) := vars if {
+	regex.match(`^\s*(some|every)`, line)
 
-	# ---------------------------------------------------> "some k, v in coll"
-	a := substring(expr, 0, indexof(expr, " in")) # -----> "some k, v"
-	b := trim_prefix(trim_prefix(a, "some"), "every") # -> "k, v"
-	c := replace(b, " ", "") # --------------------------> "k,v"
-	d := split(c, ",") # --------------------------------> [k, v] or [v]
+	vars := split(regex.replace(line, `(?:\s*(?:some|every)\s+)(\w+)(?:,?\s*)(\w+)?(?:\s+.*)`, "$1,$2"), ",")
```
johanfylling, line 56: What if there is only one var? E.g. Wouldn't  `some x in foo` output `["x", "in"]`? 
Would this be expected, or we only end up here if we have two output vars?

PR #1564, opened 2025-05-28
docs: Update docs to reflect rules table changes

charlieegan3:
There is a new rules table that can be filtered by category and searched too.

It is out now, https://docs.styra.com/regal/rules

https://github.com/StyraInc/docs/pull/418

This PR updates the contents of the docs so that we're in sync.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -1,8 +1,3 @@
----
-sidebar_custom_props:
-  exclude: true
----
-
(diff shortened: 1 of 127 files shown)


srenatus (approved):

charlieegan3:
Taking another approach here to:

- make the GH markdown look better and,
- make it so we do no markdown templating / bash messing in site generation which is error prone when we add new docs.

PR #1566, opened 2025-05-29
Small fixes before next release

anderseknert:
The non-loop-expression reordering is the major change here, as it accounts for almost 700k allocations less! Which is of course ironic for that rule :P


charlieegan3 (approved):
😆

PR #1583, opened 2025-06-05
e2e: add frankenstein test build

srenatus:
This is meant to use the Regal version from whatever is in the current checkout, together with OPA as defined in Regal's go.mod.

It's an early warning against dependency problems that only become apparent when we're using all of OPA, not the subset of its functionality that Regal already uses.


e2e/testbuild/main.go
```diff
@@ -0,0 +1,20 @@
+package main
+
+import (
+	regal_cmd "github.com/styrainc/regal/cmd"
+
+	"github.com/open-policy-agent/opa/cmd"
```
charlieegan3, line 4: nit: should these be the reverse order?

charlieegan3 (approved):
Thanks for taking the time to add this 😊

e2e/testbuild/main.go
srenatus: I have no preference. 🔀 I'll flip them.

PR #1597, opened 2025-06-17
fix: improve snippets suggestion condition

### LINKED ISSUE #1588: completion: yields snippets when uncalled-for
Then typing `input.`, Regal's LSP will propose snippets, "some" and "every" -- [the only snippets we have.](https://github.com/StyraInc/regal/blob/main/bundle/regal/lsp/completion/providers/snippet/snippet.rego#L74-L91). It really shouldn't 😅 

Image

My guess is that `word.text` is `""`, and the prefix check thus falsely hits:
```
$ opa eval -fpretty 'strings.any_prefix_match("some", "")'
true
```

I'd push a PR to avoid the empty-word case -- but I don't know if that's what we want, should someone invoke the snippet completion on an empty line, that might break...?

anderseknert:
While most editors seemed to be capable of filtering out these suggestions by themselves, not all did. And we should obviously not send suggestions that aren't valid in any case.

Fixes #1588

### DIFF
--- a/bundle/regal/lsp/completion/providers/snippet/snippet.rego
+++ b/bundle/regal/lsp/completion/providers/snippet/snippet.rego
@@ -18,10 +18,13 @@ items contains item if {
 	position := location.to_position(input.regal.context.location)
 	line := input.regal.file.lines[position.line]
 
-	not endswith(trim_space(line), "=")
 	location.in_rule_body(line)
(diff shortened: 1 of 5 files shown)


charlieegan3 (approved):
Very nice, thanks for patching this up!

PR #1603, opened 2025-06-18
lsp/completion/input: don't suggest 'input' following dot

srenatus:
While it's perhaps possible to have something like `data.foo.input.bar`, that `input` wouldn't be the same `input` as suggested here: we're explicitly documenting it as "the input document".

This should fix the situation where you can use suggestions to come up with

    input.input.input.input

### DIFF
--- a/bundle/regal/lsp/completion/providers/input/input.rego
+++ b/bundle/regal/lsp/completion/providers/input/input.rego
@@ -18,6 +18,10 @@ items contains item if {
 
 	startswith("input", word.text)
 
+	# input, as we mean it here, cannot follow dot (like data.input or input.input)
+	before := trim_suffix(line, word.text)
+	not endswith(before, ".")
+
 	item := {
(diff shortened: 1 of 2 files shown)


bundle/regal/lsp/completion/providers/input/input.rego
```diff
@@ -16,6 +16,10 @@ items contains item if {
 
 	word := location.word_at(line, input.regal.context.location.col)
 
+	before := trim_suffix(line, word.text)
+
+	# input, as we mean it here, cannot follow dot (like data.input or input.input)
+	not endswith(before, ".")
```
srenatus, line 23: ❓ Or is there a more efficient way to check this?

anderseknert: No, but you could move the startswith check below up before trimming, as that won’t be needed if that fails.

PR #1604, opened 2025-06-19
lsp: Code Action feature rewritten in Rego

anderseknert:
This is admittedly one of the simpler features, but all the more reason to have it ported :) What's new in this implementation is that we now also honor the `only` attribute sent from the client. This is an array clients can send to only request code actions of a certain kind, like "quickfix" or "source". Good clients are going to filter this anyway, but doing it on the server means less data to transfer. And of course a consistent experience regardless of client used.

This is also the first LSP feature where a schema is used for the input! And that was quite helpful for catching issues in the policy during development. It does pose a problem in the tests though, as optional attributes aren't allowed currently, which requires a few hacks to work around. But this is definitely a model to follow for future work.

While I have deleted the Go code related to this, I decided to keep the test for the handler, as it's good to have something to test the whole chain. I have also tested manually in VS Code, and it works like a charm. Some benchmarks included too, showing pretty good numbers!


bundle/regal/lsp/clients/clients.rego
```diff
@@ -0,0 +1,33 @@
...
+# METADATA
+# title: VS Code
+# description: VS Code, via the OPA extension
+# related_resources:
+#   - https://github.com/open-policy-agent/vscode-opa
+vscode := 1
+
+# METADATA
+# title: Go Test
+# description: ???
```
charlieegan3:
```suggestion
# description: Client identifier used in go tests for the language server. (internal)
```

bundle/regal/lsp/codeaction/codeaction_test.rego
```diff
@@ -0,0 +1,245 @@
...
+}
+
+_diagnostics["opa-fmt"] := {
+	"code": "opa-fmt",
+	"message": "Use opa fmt to format this file",
+	"range": {"start": {"line": 0, "character": 0}, "end": {"line": 0, "character": 1}},
+}
+
+# Silly object.union only to apppease the type checker, who for some reason thinks that
+# this violates the schema — and only in the first test. We'll have to look into that later,
```
charlieegan3, line 234: Add a TODO somewhere in here so we can grep for TODO in future?

charlieegan3: Or open an issue for this, as it does certainly seem to match the JSON schema below.

srenatus: typo: apppease (I had to look three times and squint)

bundle/regal/lsp/clients/clients.rego
anderseknert: Thanks 👍

bundle/regal/lsp/codeaction/codeaction_test.rego
anderseknert: Heh, thanks!

PR #1606, opened 2025-06-24
perf: avoid JSON/map roundtrips in ast.Value transforms

anderseknert:
Leveraging the new capabilities for this in roast, we can avoid a lot of allocations by mapping directly from a source to an ast.Value rather than going via a `map[string]any`, or even JSON. For now, focus has been on how the **input** is mapped in most places of eval. This can be improved further, but the natural next step is of course to eliminate roundtrips the same way when we handle **output**, i.e. the result of evaluation. I imagine the data-heavy aggregate rules in particular could benefit greatly from this.

This PR also cleans up some in LSP handlers related to input transformations, and takes a small step towards harmonizing the input format for those with the goal of eventually having a single schema to describe the common parts of all LSP request payloads.

Perf impact is close to 2.9 million allocs less in the Regal lint benchmark, and greatly reduced ns/op and B/op.

```
944216438 ns/op    2805943268 B/op 51659258 allocs/op - main
894474146 ns/op    2668457832 B/op 48779590 allocs/op - PR
```


pkg/version/version.go
```diff
@@ -1,68 +1,61 @@
...
+	sb.WriteString(vi.GoVersion)
+	sb.WriteString("\nPlatform:   ")
+	sb.WriteString(vi.Platform)
+	sb.WriteString("\nCommit:     ")
+	sb.WriteString(vi.Commit)
+	sb.WriteString("\nTimestamp:  ")
+	sb.WriteString(vi.Timestamp)
+	sb.WriteString("\nHostname:   ")
+	sb.WriteString(vi.Hostname)
+	sb.WriteString("\n")
```
charlieegan3: Might just be a personal preference thing but this is harder to read and check the new lines are correct etc for me than what was there before.

I guess this saves some allocs, but perhaps there's a loop something to keep the perf gain while keeping it a bit more readable?

Minor issue, feel free to ignore.

anderseknert: Nah, you're right — this was just me goofing around :) I'll revert that.

charlieegan3 (approved):
LGTM! Great work!

Issue #1446, opened 2025-03-18
Unexpected code lens error message on invalid policy


anderseknert:
By mistake, I added the number `3` after `0` in a policy I was working on. `03` isn't a valid Rego value, and this should bubble up as an error _somewhere_. That place is however not the code lens handler. This only seems to happen for this specific issue, so it seems like a quirk more than a general issue. 

<img width="1404" alt="Image" src="https://github.com/user-attachments/assets/9cd4d6f8-e893-44aa-8abf-b6a1130befcf" />

charlieegan3:
I think this particular issue is from when we switched to using "github.com/styrainc/roast/pkg/encoding" rather than rio.JSONRoundTrip. These two functions have different behaviour for numbers like `03`, the former fails and the latter does not.

charlieegan3:
Incorrect, this has always been the case while we've had the store and have round tripped parsed modules.

charlieegan3:
I have opened https://github.com/open-policy-agent/opa/issues/7725, in the meantime, what do we want to do here? I think that if the initial loadWorkspaceContents fails for some reason, we should roll over single file errors and log them rather than erroring out completely. This being a great example of one bad file that we don't want to break a project on.

I will do this tomorrow.

anderseknert:
Yeah, no input should be able to crash the server. We log and move on. If something is truly broken, we can send a notice to the client. But panics (which inevitably kills the session) just shouldn’t happen in the server IMO.

charlieegan3:
👍 I can make this function 'partially failable' instead.

charlieegan3:
PR for the underlying issue here: https://github.com/open-policy-agent/opa/pull/7726 will continue with the robustness improvements above as discussed too though.

PR #1611, opened 2025-06-25
lsp: Roll over load file failures

charlieegan3:
Fixes https://github.com/StyraInc/regal/issues/1446


anderseknert (approved):
LGTM! A simple test somewhere would be nice though.

charlieegan3:
Yeah I can get a test sorted for this.

charlieegan3:
I've added a test now that uses an unreadable file to test this new behaviour. (since the 03 json issue is likely to be fixed)

PR #1618, opened 2025-06-30
Add Regal bundle compilation benchmark

anderseknert:
This is really testing OPA rather than Regal, but since it's a good test case for improving OPA for Regal, it belongs here.

Used in https://github.com/open-policy-agent/opa/pull/7740


charlieegan3 (approved):
Sounds good to me!

PR #1619, opened 2025-06-30
Remove quotes from keywords in refs

anderseknert:
Wow, this felt almost like doing something forbidden, after all these years quoting keywords, lol. Awesome work, @johanfylling!

It seems like the formatter rewrites:

```
import data.foo.default as foo
```
to
```
import data.foo["default"] as foo
```
Which I think is a bug, as all other tools handle that. But that was really the only issue I experienced.


johanfylling:
> Which I think is a bug

Yep, looks like a bug 👍. We'll sort that out in the next release 🙂.

charlieegan3:
Filed an issue for the bug here: https://github.com/open-policy-agent/opa/issues/7742

PR #1620, opened 2025-07-01
lsp: Skip loading files on disk when in cache

charlieegan3:
This was an oversight in the previous implementation, if the file is already in the cache then the server is aware of it and we can return earlier.


charlieegan3:
thanks Stephan!

Issue #764, opened 2024-05-29
lsp: support starting server on TCP socket


charlieegan3:
There are projects that support connecting to language servers over TCP, rather than over a stdout. The messages are the same, but the transport different.

One such project of note is https://github.com/TypeFox/monaco-languageclient, this would, in theory allow a browser based monaco editor to connect to Regal LS.

I think we might be able to use https://pkg.go.dev/nhooyr.io/websocket

PR #1623, opened 2025-07-03
add experimental lsp package for websocket communication

srenatus:
For work on a playground, we've been using a regal fork to reach the LSP server through a websocket. This PR is my first attempt to upstream that functionality, so we can ditch the branch.

It includes a small package, marked as experimental, that exposes a subset of the LSP server's functionality via a websocket connection (or some other connection, but it's intended to be used with WS).

Since it's difficult to test this without a proper web-based IDE, I've added a small test page that can be started with a plain `cd dev/ws; go run .` -- intentionally lightweight, so it doesn't require a JS  build step, npm, and all that:

https://github.com/user-attachments/assets/6f2aa5be-f525-475c-9607-14d82ce45a18

That dev/ws code, which is not intended for consumption by anyone but Regal's developers, also shows how the Regal LSP can be hooked up with a websocket. That setup, however, seems intricately linked with the deployment scenario, and I don't think it can be made a part of the Regal code. Whoever wants to use this will use it in their own special way, so there's little we can do for them more than document _one way to do it_.

Related: #764.


internal/lsp/server.go
```diff
@@ -182,6 +182,31 @@ func NewLanguageServer(ctx context.Context, opts *LanguageServerOptions) *Langua
 	return ls
 }
 
+func NewLanguageServerMinimal(ctx context.Context, opts *LanguageServerOptions, cfg *config.Config) *LanguageServer {
```
anderseknert, line 187: Granted this is an internal package, but a doc comment here to explain for future contributors what this is about, would be good, I think.

dev/ws/ws.html
```diff
@@ -0,0 +1,164 @@
...
+    <div id="editor"/>
+
+    <form ws-send="">
+        <h3>initialize</h3>
+        <div>
+            <input name="jsonrpc" type="hidden" style="width:500px;" value="2.0"/>
+            <input name="id" type="hidden" style="width:500px;" value="0"/>
+            <input name="method" type="hidden" style="width:500px;" value="initialize"/>
+            <textarea name="params" style="width:500px;">{
+  "capabilities": {
```
anderseknert: So these are *client* capabilities, right?

dev/ws/main.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/styrainc/regal/pkg/lsp"
+)
+
+//go:embed *.html *.css
+var root embed.FS
+
+func main() {
+	http.Handle("/regal", http.HandlerFunc(handleWS))
+	http.Handle("/", http.FileServerFS(root))
+	panic(http.ListenAndServe(":8787", nil))
```
anderseknert: The Regal language server already starts a web server — used for pprof endpoints and the OPA explorer. Could we not just add an endpoint to that? It currently starts at a random port, but that could be configured / changed.

dev/ws/go.mod
```diff
@@ -0,0 +1,53 @@
+module github.com/styrainc/regal/dev/ws
+
+go 1.24.4
+
+replace github.com/styrainc/regal => ../..
+
+require (
+	github.com/gorilla/websocket v1.5.3
```
anderseknert: We'll want websocket support in the Regal LS anyway at some point, so unless this needs to be a module of its own, let's just add this to the main module.

anderseknert (approved):
All good with me. Just some thoughts about making this a little more generic, if possible. Also, if there's some way we can avoid a new top level directory, let's do it 😅 We could potentially rename "build" to "dev" and have it encompass more functions... but it would be a lot of docs to update..

dev/ws/ws.html
srenatus: Yeah it's an example message. It happens to be the one that the codemirror integration I've used sends, but it really could be anything. It's also editable, for experimentation.

dev/ws/main.go
srenatus: So this bit isn't meant to become part of the standard regal executable. It's an example how to use the Websocket bits as they stand. My assumption is that anyone who's making use of that would have a similar piece of code, close to their integration with the regal LSP used as a package.

This handler is also starting a new LSP session per WS session. If it was part of the standard regal web server, I'd think it should share the current context of the regal LSP...? That's a different approach, which isn't what's currently here.

dev/ws/go.mod
srenatus: (re-iterating the previous point because I've replied out-of-order) My assumption was that to use Regal via WS, you'd integrate the websocket handling in your (Go) application in some way. The new package allows you to start an LSP server per websocket session, which happens (😎) to be what the next playground iteration uses. It's also a bit of multi-tenant that way, since the file system doesn't really have to come into play.

How best to expose a WS from the regal binary is another question. Happy to help but I don't know the best design for this yet. Share an LSP session the current working directory? Start a fresh one...? 🤔

dev/ws/main.go
anderseknert: I watched the video, but I didn't understand this was only for the purpose of demo / testing. If that is so, perhaps we could make room for this inside of the docs directory.

Either way, feel free to merge at your leisure. If possible to avoid contributing to my arch-nemesis, the top level directory, I'd be delighted. But if it's too much of a hassle, fine. Regal was meant to be generally useful more than it was meant to help satisfy my OCD, even if it's done a pretty good job with that so far.

srenatus: I've got 99 problems, but moving this code ain't one

srenatus: No problem whatsoever with moving this. Already did, a push later today will come without a dev/ directory. :D

srenatus:
Updated screencast:

https://github.com/user-attachments/assets/9f476a91-38d2-41f2-91b3-b48a54cf3a63

srenatus:
I'll remove the first tab. Let's make the big init message the example payload for initialize. It'll streamline things a big.

srenatus:
image

srenatus:
@anderseknert thanks for the review. I've made it more generic, and hopefully a useful dev tool for our LSP work. I'm going to _squash and merge_ it when it's green, I think that's more appropriate -- the history of the work on this little PR isn't too interesting.

anderseknert:
Excellent 👍

PR #1624, opened 2025-07-04
internal/io: add FindInputPath when the content isn't needed

srenatus:
This is in preparation of some refactoring for io.FindInput(), which I'd like to make somehow aware of the cached files... but I think it's an OK change as-is. A bit of duplication, but less work when we don't care for the file content.

### DIFF
--- a/internal/io/io.go
+++ b/internal/io/io.go
@@ -84,11 +84,32 @@ func ExcludeTestFilter() filter.LoaderFilter {
 	}
 }
 
+// FindInputPath consults the filesystem and returns the input.json or input.yaml closes to the
+// file provided as arguments.
+func FindInputPath(file string, workspacePath string) string {
+	relative := strings.TrimPrefix(file, workspacePath)
+	components := strings.Split(filepath.Dir(relative), PathSeparator)
+
+	for i := range components {
+		current := components[:len(components)-i]
+		for _, ext := range []string{"json", "yaml"} { // NB(sr): "yml"?
+			inputPath := filepath.Join(workspacePath, filepath.Join(current...), "input."+ext)
+			_, err := os.Stat(inputPath)
+			if err == nil {
+				return inputPath
+			}
+		}
+
+	}
+
+	return ""
+}
+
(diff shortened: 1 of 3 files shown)


internal/io/io.go
```diff
@@ -84,11 +84,32 @@ func ExcludeTestFilter() filter.LoaderFilter {
...
 
+// FindInputPath consults the filesystem and returns the input.json or input.yaml closes to the
+// file provided as arguments.
+func FindInputPath(file string, workspacePath string) string {
+	relative := strings.TrimPrefix(file, workspacePath)
+	components := strings.Split(filepath.Dir(relative), PathSeparator)
+
+	for i := range components {
+		current := components[:len(components)-i]
+		for _, ext := range []string{"json", "yaml"} { // NB(sr): "yml"?
```
anderseknert: Yeah, let's include .yml for good measure. We could also move `filepath.Join(current...)` up a level, as it doesn't depend on the ext. And as long as we don't allow different names, we can just use the full name :)

```go
for i := range components {
	current := append(workspacePath, components[:len(components)-i]...)
	for _, filename := range []string{"input.json", "input.yaml", "input.yml"} {
		inputPath := filepath.Join(append(current, filename))
		if _, err := os.Stat(inputPath); err == nil {
			return inputPath
		}
	}
}
```
(I don't know if it works.. haven't tried it :D)

internal/io/io.go
```diff
@@ -84,11 +84,32 @@ func ExcludeTestFilter() filter.LoaderFilter {
...
+
+	return ""
+}
+
 // FindInput finds input.json or input.yaml file in workspace closest to the file, and returns
 // both the location and the contents of the file (as map), or an empty string and nil if not found.
 // Note that:
 // - This function doesn't do error handling. If the file can't be read, nothing is returned.
-// - While the input data theoritcally could be anything JSON/YAML value, we only support an object.
+// - While the input data theoretically could be anything JSON/YAML value, we only support an object.
```
anderseknert, line 114: 🦅 👁️

Issue #1391, opened 2025-02-04
Completions: suggested package name isn't quoted where necessary


anderseknert:
When parts of the suggested path needs quoting, that should be in the completion suggestion. Example:

<img width="454" alt="Image" src="https://github.com/user-attachments/assets/ff163aa8-dd41-42bf-8b81-222333e69dea" />

Instead of

```rego
bundle.regal.rules.style.opa-fmt
```

the suggestion should be

```rego
bundle.regal.rules.style["opa-fmt"]
```

PR #1625, opened 2025-07-17
lsp/completions: Handle hyphenated package names

charlieegan3:
Fixes https://github.com/StyraInc/regal/issues/1391

### DIFF
--- a/bundle/regal/lsp/completion/providers/packagename/packagename.rego
+++ b/bundle/regal/lsp/completion/providers/packagename/packagename.rego
@@ -42,10 +42,39 @@ _base(path) := substring(path, 0, regal.last(indexof_n(path, "/")))
 _suggestions(dir, text) := [path |
 	parts := split(dir, ".")
 	len_p := count(parts)
+
 	some n in numbers.range(0, len_p)
 
-	path := concat(".", array.slice(parts, n, len_p))
+	formatted_parts := [p |
+		some index, part in array.slice(parts, n, len_p)
+		p := _format_part(part, contains(part, "-"))
+	]
+
+	path := concat("", [p |
+		some index, part in formatted_parts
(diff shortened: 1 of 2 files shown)


bundle/regal/lsp/completion/providers/packagename/packagename.rego
```diff
@@ -42,10 +42,39 @@ _base(path) := substring(path, 0, regal.last(indexof_n(path, "/")))
 _suggestions(dir, text) := [path |
 	parts := split(dir, ".")
 	len_p := count(parts)
+
 	some n in numbers.range(0, len_p)
 
-	path := concat(".", array.slice(parts, n, len_p))
+	formatted_parts := [p |
+		some index, part in array.slice(parts, n, len_p)
+		p := _format_part(part, contains(part, "-"))
```
anderseknert: Rather than checking for `-`, should we not check for any character outside ascii and _? So that we can provide suggestions for any path that has special characters quoted?

bundle/regal/lsp/completion/providers/packagename/packagename.rego
```diff
@@ -42,10 +42,39 @@ _base(path) := substring(path, 0, regal.last(indexof_n(path, "/")))
...
+
+	path := concat("", [p |
+		some index, part in formatted_parts
+		p := _delimit_part(part, array.slice(formatted_parts, index + 1, index + 2))
+	])
+
 	path != ""
 
+	# it's not valid Rego to have a hypenated first part
+	not startswith(path, `[""`)
```
anderseknert: Not sure I follow here... `[""` doesn't have a hyphen in it. But might want to rewrite this anyway for the change suggested above.

charlieegan3: Ahh, that looks like a typo, I think it was meant to be: `["`

Issue #1608, opened 2025-06-24
lsp: new files no longer templated


anderseknert:
Or at least I'm no longer seeing this when working on Regal.

I reached for [the docs](https://docs.styra.com/regal/rules/idiomatic/directory-package-mismatch#regal-fix--editor-support) to try and see what the current expected behavior was, but there's not even a mention of the templating feature, only auto-complete. Is that covered elsewhere?

I know we introduced some restrictions for when we do templating to not make it annoying, like how we don't do it in an empty dir, etc... but I don't think we ever intended to disable it entirely.

anderseknert:
I started looking into this last night, and here's what's going on. Issue is logged as:
```
failed to get rename params: failed to fix violations: failed to fix /dir/p.rego: /dir/p.rego:0: rego_parse_error: empty module
```
What seems to happen here is that: 

1. Creating a new file invokes the `workspace/didCreateFiles` handler.
2. This sends a job on the l.templateFileJobs channel.
3. The template worker successfully writes the template to the server cache for that URI.
4. The template worker now tries to "determine if a rename is needed based on the new file contents"
5. In this step (`fixRenameParams`) the document is fetched from the cache (written to in step 3)
6. But all we get back at this point is an empty document. Why?
7. Because just after step 1, the client also triggered a `textDocument/didOpen` event.
8. Our handler for this [overwrites](https://github.com/StyraInc/regal/blob/main/internal/lsp/server.go#L1785) the cached entry with `""`, as that's the content sent by the client.

It's unclear to me why step 4 happens in the first place, as the "new file contents" was created by our template. If user creates `foo/bar/p.rego` I assume we'll properly template `package foo.bar` and renaming (i.e. moving the file) would never be needed? But I'm sure there's some nuance I'm missing.

If that _is_ needed, we could either try to have some lock ensure that `didOpen` won't overwrite an entry in the process of being templated. Or we could simply avoid it writing empty strings to the cache at all, as we don't have much use for them anyway. Not sure if that'd break other things though.

charlieegan3:
I think step 4 is happening in case the new file had content from the user. e.g. they created a file with foo.bar as the package in the root. At least I seem to remember testing that case when this was added.

I guess we need to be able to write empty strings to the cache if a user clears the contents of a file. but perhaps not on a did open event?

anderseknert:
> I think step 4 is happening in case the new file had content from the user. e.g. they created a file with foo.bar as the package in the root.

[...]

PR #1627, opened 2025-07-22
lsp: fix didOpen templating race

charlieegan3:
Fixes https://github.com/StyraInc/regal/issues/1608

Extract StartTemplateWorker body into processTemplateJob method to make it possible to test.

It might be good to implement more general locking and unlocking of file contents if we make more modifications but for now, I think this is ok as it addresses the main issue in question.

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -131,6 +131,10 @@ type LanguageServer struct {
 	builtinsPositionJobs chan lintFileJob
 	templateFileJobs     chan lintFileJob
 
+	// templatingFiles tracks files currently being templated to ensure
+	// other updates are not processed while the file is being updated.
+	templatingFiles *concurrent.Map[string, bool]
+
 	webServer *web.Server
 
 	workspaceRootURI string
@@ -170,6 +174,7 @@ func NewLanguageServer(ctx context.Context, opts *LanguageServerOptions) *Langua
 		builtinsPositionJobs:        make(chan lintFileJob, 10),
 		commandRequest:              make(chan types.ExecuteCommandParams, 10),
 		templateFileJobs:            make(chan lintFileJob, 10),
+		templatingFiles:             concurrent.MapOf(make(map[string]bool)),
 		completionsManager:          completions.NewDefaultManager(ctx, c, store),
 		webServer:                   web.NewServer(c, opts.LogWriter, opts.LogLevel),
 		loadedBuiltins:              concurrent.MapOf(make(map[string]map[string]*ast.Builtin)),
@@ -199,6 +204,7 @@ func NewLanguageServerMinimal(ctx context.Context, opts *LanguageServerOptions, 
 		builtinsPositionJobs:        make(chan lintFileJob, 10),
 		commandRequest:              make(chan types.ExecuteCommandParams, 10),
 		templateFileJobs:            make(chan lintFileJob, 10),
+		templatingFiles:             concurrent.MapOf(make(map[string]bool)),
 		completionsManager:          completions.NewDefaultManager(ctx, c, store),
 		webServer:                   web.NewServer(c, opts.LogWriter, opts.LogLevel),
 		loadedBuiltins:              concurrent.MapOf(make(map[string]map[string]*ast.Builtin)),
@@ -978,6 +984,11 @@ func (l *LanguageServer) StartWorkspaceStateWorker(ctx context.Context) {
 	}
 }
 
+// StartWebServer starts the web server that serves explorer
+func (l *LanguageServer) StartWebServer(ctx context.Context) {
+	l.webServer.Start(ctx)
+}
+
 // StartTemplateWorker runs the process of the server that templates newly
 // created Rego files.
 func (l *LanguageServer) StartTemplateWorker(ctx context.Context) {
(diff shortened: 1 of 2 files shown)


internal/lsp/server.go
```diff
@@ -1106,6 +1057,78 @@ func (l *LanguageServer) loadEnabledRulesFromConfig(ctx context.Context, cfg con
 	return nil
 }
 
+// processTemplateJob handles the templating of a newly created Rego file.
+func (l *LanguageServer) processTemplateJob(ctx context.Context, job lintFileJob) error {
```
charlieegan3, line 1061: this func body has been moved without significant changes.

internal/lsp/server.go
```diff
@@ -1813,7 +1836,12 @@ func (l *LanguageServer) handleTextDocumentDidOpen(params types.TextDocumentDidO
 		return struct{}{}, nil
 	}
 
-	l.cache.SetFileContents(params.TextDocument.URI, params.TextDocument.Text)
+	// check if file is currently being templated
+	if _, isTemplating := l.templatingFiles.Get(params.TextDocument.URI); isTemplating {
```
charlieegan3, line 1839: this is the fix

internal/lsp/server.go
```diff
@@ -1106,6 +1057,78 @@ func (l *LanguageServer) loadEnabledRulesFromConfig(ctx context.Context, cfg con
 	return nil
 }
 
+// processTemplateJob handles the templating of a newly created Rego file.
+func (l *LanguageServer) processTemplateJob(ctx context.Context, job lintFileJob) error {
+	l.logf(log.LevelMessage, "Template worker received job: %s (reason: %s)", job.URI, job.Reason)
```
anderseknert: This looks like it should be a debug level message to me :)

internal/lsp/server.go
```diff
@@ -1106,6 +1057,78 @@ func (l *LanguageServer) loadEnabledRulesFromConfig(ctx context.Context, cfg con
...
+	var edits []any
+
+	edits = append(edits, types.TextDocumentEdit{
+		TextDocument: types.OptionalVersionedTextDocumentIdentifier{URI: job.URI},
+		Edits:        ComputeEdits("", newContents),
+	})
+
+	// set the cache contents so that the fix can access this content
+	// when renaming the file if required.
+	l.cache.SetFileContents(job.URI, newContents)
```
anderseknert: This is already done above?

internal/lsp/server.go
```diff
@@ -1106,6 +1057,78 @@ func (l *LanguageServer) loadEnabledRulesFromConfig(ctx context.Context, cfg con
...
+	)
+	if err != nil {
+		l.logf(log.LevelMessage, "failed to get rename params: %s", err)
+
+		return nil
+	}
+
+	// send the edit back to the editor so it appears in the open buffer.
+	if err = l.conn.Call(ctx, methodWorkspaceApplyEdit, types.ApplyWorkspaceAnyEditParams{
+		Label: renameParams.Label,
```
anderseknert: Is this set when no rename is needed?

internal/lsp/server_template_test.go
```diff
@@ -344,3 +344,156 @@ func TestNewFileTemplating(t *testing.T) {
 		}
 	}
 }
+
+// TestTemplateWorkerRaceConditionWithDidOpen tests the race condition fix for
+// https://github.com/StyraInc/regal/issues/1608 where didOpen would overwrite
+// templated content in cache (with "") before the template worker could
+// complete.
+func TestTemplateWorkerRaceConditionWithDidOpen(t *testing.T) {
```
anderseknert, line 348: Very nice!

internal/lsp/server_template_test.go
```diff
@@ -344,3 +344,156 @@ func TestNewFileTemplating(t *testing.T) {
...
+	}
+
+	tempDir := testutil.TempDirectoryOf(t, files)
+
+	// set up the server and client connections
+	ctx, cancel := context.WithCancel(t.Context())
+	defer cancel()
+
+	receivedMessages := make(chan []byte, 10)
+	clientHandler := func(_ context.Context, _ *jsonrpc2.Conn, req *jsonrpc2.Request) (result any, err error) {
```
anderseknert: perhaps move this out to a standalone function, as this test is already quite long 🙂

internal/lsp/server.go
charlieegan3: Fixed in 80244bbd5c912903f180b12f8e30c953337216be

internal/lsp/server.go
charlieegan3: ahh, that's in the old func too. removed in 80244bbd5c912903f180b12f8e30c953337216be

internal/lsp/server_template_test.go
charlieegan3: I've extracted as it was also shared with another test, see 9f5838de0f9d892a5f7603341e440efcac8bee5f

internal/lsp/server.go
charlieegan3: Hmm, that's not very clear... I have tried to make it better in: dc0dae291a2737b1b742afae72bdde13ab684737 showing how Document Changes is built up with the optional rename params.

Issue #1609, opened 2025-06-24
Creating empty .rego file and immediately deleting it results in permanent parser error reported


anderseknert:
This has come up before, so let's make sure we improve our test coverage here. Could be that it's been broken for a while, but no one noticed as long as templating worked (#1608) to populate new files?

https://github.com/user-attachments/assets/1ddb42ee-bba8-476b-83fb-3736de6567c0

charlieegan3:
I've made a start on this one just now.

PR #1628, opened 2025-07-22
lsp: Fix stale diagnostics after deletion

charlieegan3:
Fixes https://github.com/StyraInc/regal/issues/1609

`diagnostics` must be set to an empty array, and not nil, for the clients to clear them.

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -2523,7 +2523,7 @@ func (l *LanguageServer) sendFileDiagnostics(ctx context.Context, fileURI string
 	// diagnostics must be a non-nil slice, otherwise diagnostics may not be
 	// cleared by the client.
 	if fileDiags == nil {
-		fileDiags = make([]types.Diagnostic, 0)
+		fileDiags = noDiagnostics
 	}
 
 	resp := types.FileDiagnostics{


internal/lsp/server.go
```diff
@@ -884,9 +883,7 @@ func (l *LanguageServer) StartCommandWorker(ctx context.Context) { //nolint:main
 							value = make(map[string]any)
 						}
 
-						json := encoding.JSON()
-
-						jsonVal, err = json.MarshalIndent(value, "", "  ")
+						jsonVal, err = encoding.JSON().MarshalIndent(value, "", "  ")
```
charlieegan3, line 892: this is unrelated and just a refactor, I think it's correct but not sure why we had `json := encoding.JSON()` before.

internal/lsp/server.go
```diff
@@ -683,7 +682,7 @@ func (l *LanguageServer) StartCommandWorker(ctx context.Context) { //nolint:main
 
 			var args commandArgs
 
-			err = json.Unmarshal([]byte(jsonData), &args)
+			err = encoding.JSON().Unmarshal([]byte(jsonData), &args)
```
charlieegan3, line 691: unrelated, but think we're using roast everywhere now?

internal/lsp/server.go
anderseknert: I think that might be remains of the first implementation, and me trying to make everything look more or less the same as before. That should of course have been an import alias if so thugh. Now I consider it more of a feature that it looks like different from stdlib `json`. Well, at least until the `v2` package arrives 🙂

internal/lsp/server.go
anderseknert: Yeah I _think_ there are a few places (the JSON reporters?) where we couldn't switch, because indentation / whitespace looked differently — or if it was map key ordering or something like this. This doesn't look like that place though, so this should be good 👍

internal/lsp/sendfilediagnostics_test.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/sourcegraph/jsonrpc2"
+
+	"github.com/styrainc/regal/internal/lsp/types"
+
+	"github.com/styrainc/roast/pkg/encoding"
+)
+
+// TestSendFileDiagnosticsEmptyArrays replicates the scenario from
+// https://github.com/StyraInc/regal/issues/1609 where a file that's been
+// deleted from the cache has null rather than empty arrays as diagnostics.
```
anderseknert, line 17: 👍

internal/lsp/server.go
```diff
@@ -2496,6 +2493,12 @@ func (l *LanguageServer) sendFileDiagnostics(ctx context.Context, fileURI string
 		fileDiags, _ = l.cache.GetFileDiagnostics(fileURI)
 	}
 
+	// diagnostics must be a non-nil slice, otherwise diagnostics may not be
+	// cleared by the client.
+	if fileDiags == nil {
+		fileDiags = make([]types.Diagnostic, 0)
```
anderseknert: We have already allocated this slice in the package 🙂 so unless this is mutated somewhere later, you should be able to do:

```Go
if fileDiags == nil {
    fileDiags = noDiagnostics
}
```

anderseknert (approved):
Great!

internal/lsp/server.go
charlieegan3: Ahh yeah, I'd forgotten about this!

PR #1629, opened 2025-07-23
build/lsp: add web client demo

srenatus:
This really is what I wanted to do back when contributing build/ws. It's a web-based editor using the lovely CodeMirror 6 and its new @codemirror/lsp-client package! ✨ 

https://github.com/user-attachments/assets/6768a6b4-d171-4874-8df6-ed200103cb49

What doesn't work yet without further frills is diagnostics. To get that done requires _a little more effort_, as can be seen in [this demo repo](https://github.com/andy0130tw/codemirror-lsp-client-demo/blob/master/src/lib/index.ts). However, here, I wanted to keep it as simple as possible 🤞 

## How to use

No JS build step:

```sh
cd build/lsp
go run .
````


srenatus:
I'll add a README, sure thing. I'm hesitant to link to the video because I don't know how long github's hosting of such a thing is guaranteed... 🤔

anderseknert:
@srenatus you could put it under docs/assets .. but it's not toi important

srenatus:
OK went ahead without the asset... 😅

Issue #1571, opened 2025-05-31
Add the ability for the language server to use custom linting rules.


frek818:
I don't know if I'm just lacking the knowledge but it seems that the language server is only able use bundled rules (built-in rules) and not custom rules. Is that correct? 

It's OK if that's the design decision, but I expect the language server linter to be able to "unknown" rules and lint using the remaining known rules.

I added a custom rule named `scoped-naming-convention` to my `.regal/config.yml` as shown here:

```
rules:
  custom
    scoped-naming-convention:
       level: error
```

The language server logs this message when I open a `.rego` file in neovim. 

```
failed to update all diagnostics: failed to lint: failed to prepare linter: validation failed: unknown rules: [scoped-naming-convention]
```

I've verified that the rule works outside of the language server by running the following command.

```
regal lint policy
```

Here are the full logs from neovim LSP.

```
[START][2025-05-31 09:02:42] LSP logging initiated
[ERROR][2025-05-31 09:02:42] ...p/_transport.lua:36	"rpc"	"regal"	"stderr"	"Regal Language Server (path: /Users/frek818/.local/share/nvim/mason/bin/regal, version: 0.34.0)\n"
[ERROR][2025-05-31 09:02:42] ...p/_transport.lua:36	"rpc"	"regal"	"stderr"	'using supplied workspace root directory: "/Users/frek818/code/acmeco/terraform-policies/policy", config may be inherited from parent directory\nstarting web server for docs on http://127.0.0.1:5052\n'
[ERROR][2025-05-31 09:02:42] ...p/_transport.lua:36	"rpc"	"regal"	"stderr"	"using config file: /Users/frek818/code/acmeco/terraform-policies/.regal/config.yaml\n"
[ERROR][2025-05-31 09:02:42] ...p/_transport.lua:36	"rpc"	"regal"	"stderr"	"failed to update all diagnostics: failed to lint: failed to prepare linter: validation failed: unknown rules: [scoped-naming-convention]\n"
[ERROR][2025-05-31 09:02:42] ...p/_transport.lua:36	"rpc"	"regal"	"stderr"	"failed to update file diagnostics: failed to lint: failed to prepare linter: validation failed: unknown rules: [scoped-naming-convention]\n"
[...]

PR #1631, opened 2025-07-23
lsp: Add support for custom rules

charlieegan3:
Screenshot 2025-07-23 at 15 54 00

Fixes https://github.com/StyraInc/regal/issues/1571

### DIFF
--- a/docs/custom-rules.md
+++ b/docs/custom-rules.md
@@ -367,6 +367,14 @@ occur between versions. If you need stable versions of rules and functions found
 library of your own, and use in your custom rules. Or engage with the Regal community and tell us what you need and
 depend on, and we'll try to take it into account, or at least help you find ways to make it work!
 
+## Language server
+
+When using custom rules with the Regal language server, rules will be loaded
+from the `.regal/rules` relative to the selected workspace root.
+
+Custom rules will be loaded each time a lint is run for a file or the workspace.
+If your custom rules are broken, fixes will be used on the next linting run.
+
 ## Community
 
 If you'd like to discuss custom rules development or just talk about Regal in general, please join us in the `#regal`
--- a/docs/custom-rules.md
+++ b/docs/custom-rules.md
@@ -370,7 +370,7 @@ depend on, and we'll try to take it into account, or at least help you find ways
 ## Language server
 
 When using custom rules with the Regal language server, rules will be loaded
-from the `.regal/rules` relative to the selected workspace root.
+from the `.regal/rules` directory relative to the selected workspace root.
 
 Custom rules will be loaded each time a lint is run for a file or the workspace.
 If your custom rules are broken, fixes will be used on the next linting run.
--- a/internal/lsp/lint.go
+++ b/internal/lsp/lint.go
@@ -28,7 +28,7 @@ type diagnosticsRunOpts struct {
 	RegalConfig      *config.Config
 	WorkspaceRootURI string
 	UpdateForRules   []string
-	CustomRulesPaths []string
+	CustomRulesPath  string
 
 	// File-specific
 	FileURI string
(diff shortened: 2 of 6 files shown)


internal/lsp/server.go
```diff
@@ -1022,12 +1024,31 @@ func (l *LanguageServer) getEnabledAggregateRules() []string {
 	return l.loadedConfigEnabledAggregateRules
 }
 
+func (l *LanguageServer) getCustomRulesPaths() []string {
```
anderseknert: Why is this returning a slice? The returned value is either nil or a single-item string slice, so might as well be `getCustomRulePath` and return just a string, right?

internal/lsp/lint.go
```diff
@@ -228,6 +233,7 @@ func updateAllDiagnostics(
 	overwriteAggregates bool,
 	aggregatesReportOnly bool,
 	updateDiagnosticsForRules []string,
+	customRulesPaths []string,
```
anderseknert: Maybe not for this PR, but the number of args here suggests we should use a struct instead 🙂

internal/lsp/server_custom_rules_test.go
```diff
@@ -0,0 +1,89 @@
...
+func TestLanguageServerCustomRule(t *testing.T) {
+	t.Parallel()
+
+	files := map[string]string{
+		".regal/config.yaml": "",
+		".regal/rules/custom.rego": `# METADATA
+# description: No var named "custom"
+# schemas:
+# - input: schema.regal.ast
+package custom.regal.rules.naming["custom"]
```
anderseknert: Does `custom` really need quoting? 🙂

internal/lsp/server.go
charlieegan3: The rationale was that it's only used with WithCustomRules, which takes a slice.

internal/lsp/server_custom_rules_test.go
charlieegan3: Nope! I'll get that fixed.

internal/lsp/server.go
anderseknert: I'd rather we have functions unaware of where they'll be used 🙂

charlieegan3: Fair enough, I was also wondering if there might be a case where a user might have multiple custom rule paths. I vaguely imagined nested configs etc. but didn't think too deeply.

If you think it's still not worth it, I'll make it return just a string.

charlieegan3:
> Does this handle updates to custom rules as well?

`customRulesPaths` are loaded from the filesystem each time a linter runs. So changes will be used on the next linting run. However, a change to a custom rule, won't automatically run a full workspace lint.

> If there's an issue in a custom rule, how do we handle that?

The linting will fail until the issue is fixed. There is no fallback to not using them.

internal/lsp/lint.go
charlieegan3: yeah I think I can do that now.

internal/lsp/server.go
charlieegan3: I also thought the nil, check was nicer than the "" check...

anderseknert: `linter.WithCustomRules` taking a slice seems reasonable (although `paths ...string` would have been preferable), as it provides more flexibility for API users to do whatever they may want. But having both the CLI/LSP locked to a single well-known location is also ideal, IMHO.

This is obviously a nit to begin with, and maybe nil checks are nicer than `""`, but also forever a fan of [narrowing](https://docs.styra.com/regal/rules/custom/narrow-argument) your inputs and outputs :)

internal/lsp/lint.go
```diff
@@ -22,6 +22,22 @@ import (
 	"github.com/styrainc/regal/pkg/rules"
 )
 
+// diagnosticsRunOpts contains options for file and workspace linting.
+type diagnosticsRunOpts struct {
```
anderseknert, line 26: Very nice! 👏

internal/lsp/server.go
charlieegan3: ok fair enough, I'll get it updated.

charlieegan3: Ok, I've updated this now

docs/custom-rules.md
```diff
@@ -367,6 +367,14 @@ occur between versions. If you need stable versions of rules and functions found
 library of your own, and use in your custom rules. Or engage with the Regal community and tell us what you need and
 depend on, and we'll try to take it into account, or at least help you find ways to make it work!
 
+## Language server
+
+When using custom rules with the Regal language server, rules will be loaded
+from the `.regal/rules` relative to the selected workspace root.
```
anderseknert: > from the `.regal/rules` relative

-> from the `.regal/rules` directory relative

charlieegan3: Thanks!

Issue #1630, opened 2025-07-23
Navigating by ctrl+click on external rule fails in vscode on windows


geirs73:
When ctrl + clicking on a rule or other reference that is imported from another file in our rego folder, it cannot find those files. I think it is because the path is not given correctly. We get \c%3a\ instead of C:\ in the start of the path.

I'm not quite sure whether this is due to the regal langauge server for vs code, or the opa vs code extension itself, but my guess is the former.

<img width="457" height="116" alt="Image" src="https://github.com/user-attachments/assets/b3099749-b88a-4777-97c7-3da2b7c3461a" />

#### Regal
Version:    0.35.1
Go Version: go1.24.3
Platform:   windows/amd64
Commit:     92d1ba9bffa4eae59ae848714097e4125c82ff93
Timestamp:  2025-06-30T16:05:10Z
Hostname:   github.actions.local

#### VS Code
Version: 1.102.1 (user setup)
Commit: 7adae6a56e34cb64d08899664b814cf620465925
Date: 2025-07-15T16:41:17.147Z
Electron: 35.6.0
ElectronBuildId: 11847422
Chromium: 134.0.6998.205
Node.js: 22.15.1
V8: 13.4.114.21-electron.0
OS: Windows_NT x64 10.0.22631

#### Open Policy Agent extension
Version: 0.19.0

anderseknert:
Thank you! For that feature we actually delegate most of the logic to OPA, but as a dependency of Regal, so your guess is probably correct. We’ll have it fixed before the next release.

charlieegan3:
Hey @geirs73, this should work but feel free to test if you are keen! https://github.com/StyraInc/regal/pull/1633

geirs73:
Was keen. Seems to work. 👍

charlieegan3:
Thanks for being keen! Glad this fixes the issue.

PR #1633, opened 2025-07-24
lsp: Address windows goto definition URI issue

charlieegan3:
Fixes https://github.com/StyraInc/regal/issues/1630

We were double encoding the URI in this function, this was fine on Unix like but led to an incorrect result for Windows paths.

🔈 

https://github.com/user-attachments/assets/eb3767f8-418b-4088-9104-a3651afb806b


anderseknert (approved):
Thanks!

charlieegan3:
Thanks!

PR #1634, opened 2025-07-24
Vendor roast dependency

anderseknert:
The idea of having roast be a separate project was that other projects could benefits when building tools using an optimized AST format. I still believe that's a good idea, but until we actually have another project ask for that, keeping roast here means less work for us managing these projects separately.

Most time spent here was with a golangci-lint bug, which was resolved by updating our dependency.


docs/custom-rules.md
```diff
@@ -66,7 +66,7 @@ in a policy, like the package declaration, imports and rules.
 Since OPA expects JSON as input, the module and all its child nodes are serialized, and then made available as `input`
 in Regal linter rule policies. The `opa parse --format json` command can be used to get an idea of what the structure
 of the serialized AST looks like. However, recent versions of Regal leverage an optimized AST JSON representation called
-[roast](https://github.com/styrainc/roast), which is both more compact and performant to traverse as part of
+[roast](roast.md), which is both more compact and performant to traverse as part of
```
charlieegan3, line 69: I think this will need to be (./roast)

anderseknert: Thanks, I’ll update before merge 👍

charlieegan3 (approved):
When you see the whole project like that it's a lot! Great work past and present!

PR #1636, opened 2025-07-29
Fix `regal new rule` issues + refactor io code

### LINKED ISSUE #1497: Stop `regal new rule` from deleting the first section of the data.yaml configuration
Not sure what's causing this, but I think that part of the config was added after we built `regal new`, and probably isn't taken into account (for whatever reason that is needed).

anderseknert:
- Fix issue where `regal new rule` would delete the `features` node from the config file
- Fix issue where `regal new rule` would add an internal `capabilities_url` attribute to the config file
- Fix test template for new built-in rules, as the location in the generated code was wrong
- The `--output` flag is now honored also when generating docs
- Add launch configuration for easier debugging of `regal new rule` in the future
- A **bunch** of refactoring done, mostly in our code dealing with io. I'm particularly happy about the new FIleWalker feature and how it compares to our previous code. But there's something here for everyone :)

Fixes #1497

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -35,6 +35,23 @@
         "test",
         "bundle"
       ]
+    },
+    {
+      "name": "regal new rule --type builtin --category custom --name bar",
+      "type": "go",
+      "request": "launch",
+      "mode": "auto",
+      "program": "${workspaceFolder}",
+      "args": [
+        "new",
+        "rule",
+        "--type",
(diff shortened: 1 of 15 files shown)


anderseknert:
@charlieegan3 Those are all astute observations! Or I'd like to think so, as they are also in my notes :) I just felt like I had to cut a PR at some point, so I'll follow up with those changes next.

Issue #1635, opened 2025-07-25
Error popups with file path encoding errors when running evaluate with regal in win-vscode


geirs73:
When running evaluation on a test that references rego policies that contain illegal code, I get error messages with what seems to be windows encoding issues for the file path. Ref issue #1630 (same setup). Same caveat as in that issue, I suspect it is in regal, but could be in opa or opa extension. If you are supposed to be able to navigate directly to the file and line number given in the error message, it certainly doesn't work. If not, then the issue is not that important.

I also notice that I get two of them every time, but I'm not sure whether that is a bug or not. Maybe worth checking for when replicating. 

<img width="573" height="582" alt="Image" src="https://github.com/user-attachments/assets/a9ee0976-490c-49b9-87d7-e079a050489a" />

charlieegan3:
Hey, this looks like another similar issue. I'll take a look into it.

charlieegan3:
Hey @geirs73, I've not been able to replicate this. While the `c%3A` doesn't look right in the errors, the actual issue seems to be related to the Rego code. Can you share a minimal example so I can test locally as the evaluate lens seems to be working ok for me.

geirs73:
Ahoy @charlieegan3 , I'll try to describe it better when back from vacation in the end of this week. 🌞

charlieegan3:
Hey, thanks @geirs73! No hurry, enjoy your time off!

geirs73:
The rego code was just a random problem that I've already resolved. I've just been hunting any signs of windows encoding issues lately. As I said, this one is very minor unless it hinders direct navigation to the problem.

Here is what I get when running Evaluate on the test method on the single rego (test) file in this repo: https://github.com/geirs73/reproduce-regal-eval-error-issue -- I've just commented out a line to trigger the error by introducing an unsafe variable.

<img width="1486" height="993" alt="Image" src="https://github.com/user-attachments/assets/100ad054-813a-4762-b836-b757529bfaf9" />

charlieegan3:
Yeah there's also an issue that the message is shown 2x! I have a fix in https://github.com/StyraInc/regal/pull/1638 that shows relative filenames instead.

geirs73:
I'm sorry to report that the 'go to definition' (F12) functionality that was fixed in #1633 stopped working in windows VS Code after #1638 was merged to main. I'm not sure how vscode handles file paths, but it cannot resolve the new relative ones when using 'go to definition'. 

[...]

PR #1638, opened 2025-07-31
lsp: Present filename relative to root in errors

charlieegan3:
This makes a change so that when a file is shown in an error message, we show the filename relative to the root of the project rather than the full URI.

This is done by changing the name of the file used in parsing.

I have also adjusted updateParse to take Opts struct instead to make it less verbose.

Fixes https://github.com/StyraInc/regal/issues/1635

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -315,7 +315,14 @@ func (l *LanguageServer) StartDiagnosticsWorker(ctx context.Context) {
 
 				// updateParse will not return an error when the parsing failed,
 				// but only when it was impossible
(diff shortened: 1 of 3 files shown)


internal/lsp/server.go
```diff
@@ -837,14 +858,6 @@ func (l *LanguageServer) StartCommandWorker(ctx context.Context) { //nolint:main
 				if err != nil {
 					fmt.Fprintf(os.Stderr, "failed to evaluate workspace path: %v\n", err)
 
-					cleanedMessage := strings.Replace(err.Error(), l.workspaceRootURI+"/", "", 1)
```
charlieegan3, line 840: This is also removed since ANY error is reported below and this just makes it a duplicated message.

anderseknert (approved):
Nice!

PR #1640, opened 2025-08-02
test: refactor e2e test, add new regal command runner

anderseknert:
While really important, the e2e tests have been a bit painful to work with, much because so much of the tests have been procedural scaffolding rather than test logic. With this refactor, I'm hoping to improve that by introducing a more declararative approach in the form of a tiny DSL for a command runner.

The result is less than half the lines of test code, and tests that IMHO make it quite easy to quickly understand what's being tested. I hope others feel the same :)


charlieegan3 (approved):
This is a great improvement, it'll hopefully make future e2e test development less error-prone too. 👏

PR #1642, opened 2025-08-05
refactor: Improve code quality throughout the codebase

anderseknert:
This started off as a follow up PR to #1636, but grew somewhat from there :)

- Move the new file walker to a separate package + improvements
- Merge logic shared by the lint and fix command. Still more to do here but it's a start
- Genral improvements in primarily the linter packages


internal/lsp/uri/uri.go
```diff
@@ -70,6 +68,7 @@ func ToPath(client clients.Identifier, uri string) string {
 		path = strings.TrimPrefix(path, "/")
 		// handling case for windows when the drive letter is set
 
+		// TODO; never set?
```
charlieegan3, line 71: to be removed?

charlieegan3 (approved):
Image

😅

internal/lsp/uri/uri.go
anderseknert: Sure! This was meant as a note to you as I haven't worked much on this. It seemed like your intention was to use this, but I take it this isn't needed then?

```go
// handling case for windows when the drive letter is set

// TODO; never set?
var driveLetter string

if matches := drivePatternMaybeEncoded.FindStringSubmatch(path); len(matches) > 1 {
    path = strings.TrimPrefix(path, matches[0])
    path = matches[1] + ":" + strings.TrimPrefix(path, driveLetter)
} else {
    path = "/" + path
}
```

Should the whole `drivePatternMaybeEncoded` block be removed?

charlieegan3: yeah looks like it's not used... unsure why they didn't get caught in the linter?

PR #1643, opened 2025-08-05
Fix missing source.explore code action

anderseknert:
This went missing without us noticing :( Due to the input not being what the policy expected (and even verified at compile time). We should look into how to leverage those schemas for runtime checks, as that could save a lot of time troubleshooting. But for now, fixing the issue and adding a regression test for the handler.

### DIFF
--- a/internal/lsp/handle_text_document_code_action_test.go
+++ b/internal/lsp/handle_text_document_code_action_test.go
@@ -58,6 +58,123 @@ func TestHandleTextDocumentCodeAction(t *testing.T) {
 		},
 	}
 
+	actualAction := invokeHandleTextDocumentCodeAction(t, l, params)
+
+	assertExpectedCodeAction(t, expectedAction, actualAction)
+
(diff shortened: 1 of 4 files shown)


internal/lsp/handle_text_document_code_action_test.go
```diff
@@ -58,66 +58,140 @@ func TestHandleTextDocumentCodeAction(t *testing.T) {
...
+	assertExpectedCodeAction(t, expectedAction, actualAction)
+
 	expArgs, actualArgs := *expectedAction.Command.Arguments, *actualAction.Command.Arguments
 	if exp, got := len(expArgs), len(actualArgs); exp != got {
 		t.Fatalf("expected %d arguments, got %d", exp, got)
 	}
+}
 
-	var expDecoded, actualDecoded map[string]any
+func assertExpectedCodeAction(t *testing.T, expected, actual types.CodeAction) {
```
charlieegan3, line 130: very nice!

PR #1644, opened 2025-08-05
lsp/uri: Refactor URI handling

charlieegan3:
This had been botched some time back and can be made simpler.

### DIFF
--- a/internal/lsp/uri/uri.go
+++ b/internal/lsp/uri/uri.go
@@ -9,10 +9,7 @@ import (
 	"github.com/styrainc/regal/internal/lsp/clients"
 )
 
-var (
-	drivePattern             = regexp.MustCompile(`^([A-Za-z]):`)
-	drivePatternMaybeEncoded = regexp.MustCompile(`^([A-Za-z])(%3[aA]|:)`)
-)
+var drivePattern = regexp.MustCompile(`^\/?([A-Za-z]):`)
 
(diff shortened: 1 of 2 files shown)


internal/lsp/uri/uri.go
```diff
@@ -56,26 +53,20 @@ func FromPath(client clients.Identifier, path string) string {
 // client URIs into a standard file paths.
 func ToPath(client clients.Identifier, uri string) string {
 	// if the uri appears to be a URI with a file prefix, then remove the prefix
-	path, hadPrefix := strings.CutPrefix(uri, "file://")
-	if hadPrefix {
-		// if it looks like a URI, then try and decode it
-		if decodedPath, err := url.QueryUnescape(path); err == nil {
```
anderseknert, line 62: Why change this? 🙂 Seems like CutPrefix answers the same question in one function call instead of two, no?

log.txt
anderseknert, line 1: This should probably not be here

charlieegan3: yeah I thought so too, so I've removed it?

internal/lsp/uri/uri.go
charlieegan3: yeah that makes sense. I'll keep that in there!

log.txt
anderseknert: lol, my mistake!

PR #1645, opened 2025-08-05
Tidying up more code

anderseknert:
No functional changes here, or at lest not intentionally so 😅

Just making code a bit more short and to the point where possible. Some things to disagree with maybe, and let me know if so as I'm open to suggestions.


charlieegan3 (approved):
LGTM!

PR #1646, opened 2025-08-06
Provide development option to load bundle from disk

anderseknert:
This allows live editing of Rego in the language server, and see the result propagated to the UI right away. A very nice development experience!

This is however very much experimental at this point, and only by setting an `REGAL_BUNDLE_PATH` to point at the bundle path will this be enabled. Most likely a few issues left to resolve here, but let's take it from here!

If we start using this often, we'll also want to optimize bundle loading too, as that happens *very* frequently. Perhaps we could have a file watcher notify us of changes or something like that, and only refresh the bundle when that happens. @charlieegan3 let's talk about that some time :)

https://github.com/user-attachments/assets/40b96ef7-d180-42d8-9c4a-d35a4306f8de

### DIFF
--- a/bundle/bundle.go
+++ b/bundle/bundle.go
@@ -2,15 +2,68 @@ package bundle
 
 import (
 	"embed"
+	"fmt"
+	"os"
+	"sync"
+	"sync/atomic"
+
+	"github.com/open-policy-agent/opa/v1/bundle"
 
 	rio "github.com/styrainc/regal/internal/io"
 )
 
 // Bundle FS will include the tests as well, but since that has negligible impact on the size of the binary,
-// it's preferable to filter them out from the bundle than to e.g. create a separate directory for tests
-//
-//go:embed *
-var Bundle embed.FS
+// it's preferable to filter them out from the bundle than to e.g. create a separate directory for tests.
+var (
+	//go:embed *
+	regalBundle    embed.FS
+	devPath        = os.Getenv("REGAL_BUNDLE_PATH")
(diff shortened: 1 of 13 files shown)


bundle/bundle.go
```diff
@@ -2,15 +2,57 @@ package bundle
...
+			if last := lastErrMsg.Load(); last != nil {
+				lastErrMsg.Store(nil)
+
+				fmt.Fprintln(os.Stderr, "development bundle back to a good state, no longer using embedded bundle")
+			}
+
+			return b
+		}
+
+		// Avoid flooding the console/logs with the same error message
```
charlieegan3, line 56: I guess you see this when there are issues parsing a WIP bundle file? Could add a note about that in the comment here.

bundle/bundle.go
```diff
@@ -2,15 +2,57 @@ package bundle
...
+	})
+)
+
+// LoadedBundle contains the Regal bundle.
+func LoadedBundle() *bundle.Bundle {
+	// For development, allow bundle to be loaded dynamically from path instead
+	// of the normal one embedded in the compiled binary. This allows editing e.g.
+	// LSP policies while the language server is running. This should be considered
+	// *very* experimental at this point.
+	if devPath != "" {
```
charlieegan3, line 42: On the topic of logs, we should log (perhaps only once) that we are using dev mode.

anderseknert: Good idea! I'll amend the PR to do that before merge.

PR #1647, opened 2025-08-07
lsp/eval: Ensure filenames use URI in print output

charlieegan3:
This is a regression fix from https://github.com/StyraInc/regal/pull/1638 which caused the print output to not be rendered in supported clients.

Fixes https://github.com/open-policy-agent/vscode-opa/issues/352


internal/lsp/eval.go
```diff
@@ -178,11 +185,16 @@ func prepareRegoArgs(
 }
 
 func (h PrintHook) Print(ctx print.Context, msg string) error {
-	if _, ok := h.Output[ctx.Location.File]; !ok {
-		h.Output[ctx.Location.File] = make(map[int][]string)
+	filename := ctx.Location.File
+	if h.FileNameBase != "" {
+		filename = strings.TrimSuffix(h.FileNameBase, "/") + "/" + ctx.Location.File
```
anderseknert: Can use `util.EnsureSuffix` here. Added recently :)

internal/lsp/eval_test.go
```diff
@@ -57,6 +78,16 @@ func TestEvalWorkspacePath(t *testing.T) {
 	if val, ok := res.Value.(bool); !ok || val != true {
 		t.Fatalf("expected true, got false")
 	}
+
+	expectedPrintOutput := map[string]map[int][]string{
+		policy2URI: {
+			4: {"1"},
+		},
+	}
```
anderseknert, line 84: Does this one item map really warrant 5 lines? lol

internal/lsp/eval_test.go
```diff
@@ -57,6 +78,16 @@ func TestEvalWorkspacePath(t *testing.T) {
...
+
+	if diff := cmp.Diff(expectedPrintOutput, res.PrintOutput); diff != "" {
```
anderseknert, line 86: Nice I didn't know about cmp.Diff

anderseknert (approved):

internal/lsp/eval_test.go
charlieegan3: it's pretty good, used it in OPA earlier too.

internal/lsp/eval_test.go
charlieegan3: Fixed now!

internal/lsp/eval.go
charlieegan3: Ahh yeah, _nearly_ remembered that! next time!

PR #1652, opened 2025-08-11
Rewrite more completion providers in Rego. Resolve schemas at runtime.

anderseknert:
- Add capability to resolve shemas from annotations at runtime
- Rewrite `ruleheadkeyword` completions provider in Rego
- Rewrite `rulename` (previously `rulehead`) completions provider in Rego
- Refactor Go code for readability, fix minor issues

The main "feature" here is how it's now possible to resolve schemas from annotations at runtime:

```rego
package regal.lsp.codeaction

input_validated if json.match_schema(input, data.schema.regal.lsp.codeaction)
```

This allows us to quickly check that the input we get from the handler is what our policy expects. For now, we don't actually use this, but I'm planning to add a "development mode" toggle to do this in my next PR.


pkg/config/bundle.go
```diff
@@ -13,21 +11,19 @@ import (
 )
 
 func LoadConfigWithDefaultsFromBundle(regalBundle *bundle.Bundle, userConfig *Config) (Config, error) {
-	path := []string{"regal", "config", "provided"}
-
-	bundled, err := util.SearchMap(regalBundle.Data, path)
+	bundled, err := util.SearchMap(regalBundle.Data, "regal", "config", "provided")
 	if err != nil {
-		return Config{}, fmt.Errorf("config path not found %s: %w", strings.Join(path, "."), err)
+		panic(err)
```
charlieegan3, line 16: Unsure if these were meant to be left as panic?

charlieegan3 (approved):
This is a great improvement Anders! 👏

bundle/bundle.go
```diff
@@ -17,20 +17,32 @@ import (
...
-func init() {
-	if devPath != "" {
-		fmt.Fprintln(os.Stderr, "REGAL_BUNDLE_PATH set. Will attempt using development bundle from:", devPath)
-	}
+type DevSettings struct {
+	bundlePath string
+	mux        sync.RWMutex
+}
+
+func (ds *DevSettings) SetBundlePath(path string) {
```
charlieegan3, line 35: Unused?

pkg/config/bundle.go
anderseknert: Yes! If we for some unlikely reason we fail to parse the *default* configuration, that’d be a condition from which we have no way to recover. User configuration issues should be reported, but are not critical as they can be fixed. My ideal is that no code should need to do nil checks for config, as it’s *never* nil, but at worst only the provided default one.

bundle/bundle.go
anderseknert: Almost, but not quite :) 

Screenshot 2025-08-11 at 19 10 27

PR #1654, opened 2025-08-13
lsp: Add signature help

charlieegan3:

### DIFF
--- a/bundle/regal/lsp/signature-help/main.rego
+++ b/bundle/regal/lsp/signature-help/main.rego
@@ -1,0 +1,79 @@
+# METADATA
+# description: package for signature help provider policies.
+# schemas:
+# - input: {type: object}
+package regal.lsp["signature-help"]
+
+import rego.v1
+
+# METADATA
+# description: main entry point for builtin signature help.
+# entrypoint: true
+default result := null
+
+result := {
+	"signatures": [{
+		"label": _build_function_label(builtin_info.decl, func_info.name),
+		# some builtins had a space at the start
+		"documentation": trim_space(builtin_info.description),
+		"parameters": _build_parameters(builtin_info.decl.args),
+		"activeParameter": func_info.active_param - 1,
+	}],
+	"activeSignature": 0,
+	"activeParameter": func_info.active_param - 1,
+} if {
+	func_info := _function_at_position(input.content, input.position)
+	builtin_info := data.workspace.builtins[func_info.name]
+}
+
+default _function_at_position(_, _) := {}
+
+_function_at_position(content, position) := {
+	"name": last_match[1],
+	"active_param": strings.count(last_match[2], ",") + 1,
+} if {
+	result := regex.find_all_string_submatch_n(
+		`([a-zA-Z_][a-zA-Z0-9_.]*)\(([^)]*)$`,
+		_text_up_to_position(split(content, "\n"), content, position),
+		-1, # we want the last one specifically, so need to get all
+	)
+
+	last_match := result[count(result) - 1]
+}
+
+# when position is after the last line
+_text_up_to_position(lines, content, position) := content if position.line >= count(lines)
+
+# when char is off the last line
+_text_up_to_position(lines, content, position) := content if {
+	position.line < count(lines)
+	current_line := lines[position.line]
+	position.character >= count(current_line)
+}
+
+_text_up_to_position(lines, _, position) := concat("\n", all_lines) if {
+	position.line < count(lines)
+
+	current_line := lines[position.line]
+	position.character < count(current_line)
+
+	complete_lines := [line | some i, line in lines; i < position.line]
+	partial_line := substring(current_line, 0, position.character)
+
+	all_lines := array.concat(complete_lines, [partial_line])
+}
+
+_build_function_label(decl, func_name) := label if {
+	param_labels := concat(", ", [_param_label(param) | some param in decl.args])
+	label := sprintf("%s(%s) -> %s", [func_name, param_labels, decl.result.type])
+}
+
+_build_parameters(args) := [{
+	"label": label,
+	"documentation": sprintf("(%s): %s", [label, param.description]),
+} |
+	some param in args
+	label := _param_label(param)
+]
+
+_param_label(param) := concat(": ", [param.name, param.type])
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -259,6 +259,8 @@ func (l *LanguageServer) Handle(ctx context.Context, _ *jsonrpc2.Conn, req *json
 		return handler.WithContextAndParams(ctx, req, l.handleTextDocumentCodeLens)
 	case "textDocument/completion":
 		return handler.WithContextAndParams(ctx, req, l.handleTextDocumentCompletion)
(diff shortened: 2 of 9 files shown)


bundle/regal/lsp/signature-help/main.rego
```diff
@@ -0,0 +1,79 @@
...
+		"activeParameter": func_info.active_param - 1,
+	}],
+	"activeSignature": 0,
+	"activeParameter": func_info.active_param - 1,
+} if {
+	func_info := _function_at_position(input.content, input.position)
+	builtin_info := data.workspace.builtins[func_info.name]
+}
+
+function_pattern := `([a-zA-Z_][a-zA-Z0-9_.]*)\(([^)]*)$`
```
anderseknert: You can inline this in `_function_at_position` as it's only referenced there, and OPA will cache the compiled query anyway 🙂 (and a local reference will be faster)

bundle/regal/lsp/signature-help/main.rego
```diff
@@ -0,0 +1,79 @@
+package regal.lsp["signature-help"]
+
+import rego.v1
+
+default result := null
+
+result := {
+	"signatures": [{
+		"label": _build_function_label(builtin_info, func_info.name),
+		# some builtins had a space at the start
```
anderseknert, line 17: Interesting. Should we fix that?

bundle/regal/lsp/signature-help/main.rego
```diff
@@ -0,0 +1,79 @@
...
+	"activeSignature": 0,
+	"activeParameter": func_info.active_param - 1,
+} if {
+	func_info := _function_at_position(input.content, input.position)
+	builtin_info := data.workspace.builtins[func_info.name]
+}
+
+function_pattern := `([a-zA-Z_][a-zA-Z0-9_.]*)\(([^)]*)$`
+
+default _function_at_position(_, _) := {}
```
anderseknert, line 29: Is the default value needed?

bundle/regal/lsp/signature-help/main.rego
```diff
@@ -0,0 +1,79 @@
...
+		-1, # we want the last one specifically, so need to get all
+	)
+
+	last_match := result[count(result) - 1]
+}
+
+# when position is after the last line
+_text_up_to_position(content, position) := content if {
+	lines := split(content, "\n")
+	position.line >= count(lines)
```
anderseknert: You could use strings.count of "\n" + 1 here too to avoid the split. Or perhaps pass `lines` as an arg since all functions use that

bundle/regal/lsp/signature-help/main.rego
```diff
@@ -0,0 +1,79 @@
...
+
+_build_parameters(builtin_info) := [{
+	"label": label,
+	"documentation": sprintf("(%s): %s", [label, param.description]),
+} |
+	some param in builtin_info.decl.args
+	label := _param_label(param)
+]
+
+_param_label(param) := sprintf("%s: %s", [param.name, param.type])
```
anderseknert: `concat(": ", [param.name, param.type])` should perform better here if it might matter.

bundle/regal/lsp/signature-help/main.rego
```diff
@@ -0,0 +1,79 @@
...
+
+# when char is off the last line
+_text_up_to_position(content, position) := content if {
+	lines := split(content, "\n")
+	position.line < count(lines)
+	current_line := lines[position.line]
+	position.character >= count(current_line)
+}
+
+_text_up_to_position(content, position) := concat("\n", all_lines) if {
```
anderseknert: Multi-line functions? Awesome!

internal/lsp/server.go
```diff
@@ -2412,6 +2455,20 @@ func (l *LanguageServer) builtinsForCurrentCapabilities() map[string]*ast.Builti
 	return rego.BuiltinsForCapabilities(ast.CapabilitiesForThisVersion())
 }
 
+// storeBuiltinsAndCacheSignatureHelp stores builtins in the rego store and caches
+// the signature help query. This should be called whenever builtins are loaded/updated.
+func (l *LanguageServer) storeBuiltinsAndCacheSignatureHelp(ctx context.Context, builtins map[string]*ast.Builtin) error {
```
anderseknert: Nice 👍

internal/lsp/server_signature_help_test.go
```diff
@@ -0,0 +1,142 @@
...
+		"regex.match": ast.RegexMatch,
+	}
+
+	// this is needed to prepare the server for signature help queries
+	err := ls.storeBuiltinsAndCacheSignatureHelp(ctx, builtins)
+	if err != nil {
+		t.Fatalf("failed to store builtins and cache signature help: %s", err)
+	}
+
+	testCases := map[string]struct {
```
anderseknert, line 50: Great tests!

internal/lsp/types/types.go
```diff
@@ -111,6 +111,7 @@ type ServerCapabilities struct {
 	CompletionProvider         CompletionOptions       `json:"completionProvider"`
 	InlayHintProvider          InlayHintOptions        `json:"inlayHintProvider"`
 	HoverProvider              bool                    `json:"hoverProvider"`
+	SignatureHelpProvider      *SignatureHelpOptions   `json:"signatureHelpProvider,omitempty"`
```
anderseknert: Even if spec says it's optional, we will always have this set. So we can just make this a value instead of a pointer.

internal/lsp/types/types.go
```diff
@@ -348,6 +358,36 @@ type FoldingRange struct {
...
+
+type SignatureHelp struct {
+	Signatures      []SignatureInformation `json:"signatures"`
+	ActiveSignature *uint                  `json:"activeSignature,omitempty"`
+	ActiveParameter *uint                  `json:"activeParameter,omitempty"`
+}
+
+type SignatureInformation struct {
+	Label           string                 `json:"label"`
+	Documentation   any                    `json:"documentation,omitempty"` // string | MarkupContent
```
anderseknert: Unless we send markdown (doesn't look like it?), let's just use a string value here.

bundle/regal/lsp/signature-help/main.rego
charlieegan3: It was just to avoid undefined in the tests when there were no matches.

bundle/regal/lsp/signature-help/main.rego
charlieegan3: https://github.com/open-policy-agent/opa/pull/7836

charlieegan3: Just one as far as I can tell.

bundle/regal/lsp/signature-help/main.rego
anderseknert: Fair enough!

srenatus:
💭 does it show in the example lsp setup?

charlieegan3:
It doesn't show in the editor, but I can add a new custom message type?

charlieegan3:
Hmm, looking into this more I think I'll leave for another PR. We need to have a way to get the builtins set in the store when not using the config worker.

PR #1657, opened 2025-08-14
New LSP features and internal improvements

anderseknert:
- New LSP features implemented
  - `textDocument/documentHighlight` which currently only helps highlight attribute names in metadata annotations, but can do more in the future
  - `textDocument/documentLink` which allows adding links to specific ranges of text in a document. First implementation uses this to link rule names in inline ignore directives to their documentation, but this too can be used for a lot more later.
  - `$/setTrace` implemented, allowing clients to set the log level of the server at runtime.
  - Add experimental support for incremental updates in `textDocument/didChange`. I had some minor issues running with this so this is not yet enabled by default, but a good start for when we want to proceed further with this.
- General cleanup of logging code in the language server
- Continued improving tests for readability


internal/lsp/types/types.go
```diff
@@ -1,588 +1,575 @@
...
-
-type iuint interface{ ~int | ~uint }
-
-func RangeBetween[T1, T2, T3, T4 iuint](startLine T1, startCharacter T2, endLine T3, endCharacter T4) Range {
-	return Range{
-		Start: Position{Line: uint(startLine), Character: uint(startCharacter)},
-		End:   Position{Line: uint(endLine), Character: uint(endCharacter)},
-	}
-}
+type (
```
charlieegan3, line 8: this is done to make sure we have no methods in here?

internal/lsp/server.go
```diff
@@ -120,7 +117,7 @@ type LanguageServer struct {
 	loadedConfigAllRegoVersions          *concurrent.Map[string, ast.RegoVersion]
 	loadedBuiltins                       *concurrent.Map[string, map[string]*ast.Builtin]
 
-	clientInitializationOptions types.InitializationOptions
+	client types.Client
```
charlieegan3, line 120: I can't see types.Client in types 🔭 am I missing something?

internal/lsp/server_load_workspace_test.go
```diff
@@ -129,7 +129,7 @@ func TestLoadWorkspaceContents(t *testing.T) {
 
 			server := NewLanguageServer(t.Context(), &LanguageServerOptions{})
 			server.workspaceRootURI = "file://" + tempDir
-			server.clientIdentifier = clients.IdentifierGeneric
+			server.client.Identifier = clients.IdentifierGeneric
```
charlieegan3, line 132: To me, `client` is more clientConfig since it's not actually the connected client. But no strong feelings here.

internal/lsp/types/types.go
anderseknert: Yeah, but mostly to try it out as it was a new discovery to me :P

charlieegan3: ok nice, might be good to leave a comment just in case it's unclear.

internal/lsp/server_load_workspace_test.go
anderseknert: > To me, client is more clientConfig since it's not actually the connected client.

...yet 🙂

internal/lsp/server.go
anderseknert: there's an internal.go file for types that aren't part of the LSP

PR #1665, opened 2025-08-26
Add SECURITY.md

tsandall:
Note I've explicitly not done signed off in the commit message because I want to exercise the DCO check.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,5 @@
+# Security Policy
+
+Please refer to the [OPA Security Policy](https://openpolicyagent.org/security)
+for details on how to report security issues, our disclosure policy, and how to
+receive notifications about security issues.


anderseknert:
Seems to be working :)

anderseknert:
Now just to figure out how we turned off that annoying Codecov bot before. 

We can close this, right, @tsandall ? That doc exists here already https://github.com/open-policy-agent/regal/blob/main/docs/SECURITY.md

srenatus:
@anderseknert Would it kill you to have one more line in the repo file view? 😉

tsandall:
Thanks! The other one is fine and the DCO check is working. Closing this.

anderseknert:
@srenatus probably not, but those standard docs are about as likely to be read as the license, so whatever can be placed under `docs/` should IMHO be placed there. And well, they _are_ docs, aren't they, and placed under `docs/` makes for less to scroll on the main page before users hit the README, etc :)

Not sure why this isn't an official GitHub resource: https://github.com/joelparkerhenderson/github-special-files-and-paths

srenatus:
Oh that's a great resource. Also: 

> Default SECURITY files that apply across all repositories can be placed in an organization's .github repository, see https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/creating-a-default-community-health-file.

anderseknert:
Wow, yeah that would do even more to satisfy my puritan OCD… but it also adds a GitHub dependency on basic information for a repo. So someone deciding to run a fork on Codeberg or whatever, would not have these files served to their visitors. Sounds like a great idea for a private org though!

PR #1667, opened 2025-08-27
Enable pprof endpoint only with env var set

anderseknert:
🔐🎭

### DIFF
--- a/internal/web/server.go
+++ b/internal/web/server.go
@@ -7,6 +7,8 @@ import (
 	"html/template"
 	"net"
 	"net/http"
+	"net/http/pprof"
+	"os"
 	"strings"
 
 	"github.com/arl/statsviz"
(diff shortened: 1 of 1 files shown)


internal/web/server.go
```diff
@@ -182,8 +196,7 @@ func (s *Server) Start(_ context.Context) {
 	s.log.Message("starting web server for docs on %s", s.baseURL)
 
 	//nolint:gosec // this is a local server, no timeouts needed
-	err = http.Serve(listener, mux)
-	if err != nil {
+	if err = http.Serve(listener, mux); err != nil {
```
srenatus, line 199: [nit]
```suggestion
	if err := http.Serve(listener, mux); err != nil {
```

srenatus (approved):
LGTM

internal/web/server.go
anderseknert: We don't do nit comments here, we do linter rules :) Not sure there's one with a toggle for this though..

PR #1675, opened 2025-09-02
Rego-based routing of Rego handlers + refactor

anderseknert:
This is a lot, and I'm sorry about that 😅 Should have pushed something earlier, but was caught up in other things, and this was mostly work done during travel and brief moments here and there.

The main thing here is the new Rego-router, which handles the requirements of individual routes, and uses a single prepared query + dynamic policy composition for routing in Rego.

- Add new main entrypoint for Rego-based routes
- Rearrange files in LSP directory and use subpackages to avoid having everything in just one big dir/package
- Merge util package of roast with internal/util

There are now only 2 remaining completion providers written in Go! These are a little more complex to port, but is of course the obvious next step. This would allow us to get rid of the Go-based completions manager entirely.


bundle/regal/lsp/codeaction/codeaction.rego
```diff
@@ -11,8 +11,11 @@ package regal.lsp.codeaction
 import data.regal.lsp.clients
 
 # METADATA
-# description: A set of all code actions applicable in the current document
 # entrypoint: true
+result["response"] := actions
```
srenatus, line 15: It's the pretty-printer that changes this from
```
result.response := actions
```
I guess. Too bad... 🤔

srenatus (approved):
Reviewed the first half, skimmed the rest. It looks OK to me, it would have been easier on the eyes (and the browser) if it was done in separate commits. (Psst, can I interest you in another git frontend called jj? ㊙️)

bundle/regal/lsp/codeaction/codeaction.rego
anderseknert: Indeed :(

bundle/regal/lsp/main/main.rego
```diff
@@ -0,0 +1,28 @@
...
+		# The payload as defined in the language server specification
+		"response": result.response,
+		# Optional metadata that handlers may want to send back as part
+		# of their response. This can be processed either here or in the
+		# Go handler later.
+		"regal": object.get(result, ["regal"], {}),
+	}
+}
+
+_handler_for(method) := lower(name) if ["textDocument", name] = split(method, "/")
```
charlieegan3, line 28: I guess we are only really expecting to use these for textDocuemnt methods for starters?

charlieegan3: I can't reasonably see how the other methods would work unless the entire state was in in the rego store etc.

anderseknert: Yes, limited to textDocument/* methods for the time being. A text document URI is the lowest common denominator here, and all policies assume one in the params. Can add something to the description in the next PR.

PR #1680, opened 2025-09-08
jjlsp: GTD use URI for location

charlieegan3:
Parsed modules use relative paths now to keep error messages correct for users. This means we have to contextualise filenames if we need to use them for URIs in LSP messages. This is done using the current root URI from the server.


charlieegan3:
Thanks Stephan!

charlieegan3:
Caused by https://github.com/open-policy-agent/regal/pull/1638

PR #1682, opened 2025-09-08
build: Add windows runners to build workflow

charlieegan3:
We used to run tests on windows runners, hopefully it'll help catch bugs earlier.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -16,7 +16,7 @@ permissions: read-all
 jobs:
   build-matrix:
     env:
-      RQ_VERSION: v0.0.9
+      RQ_VERSION: v0.0.15
     name: Matrix
     strategy:
       matrix:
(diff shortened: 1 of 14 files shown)


.github/workflows/build.yaml
```diff
@@ -122,3 +122,21 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
+        with:
+          go-version-file: go.mod
+      - uses: open-policy-agent/setup-opa@34a30e8a924d1b03ce2cf7abe97250bbb1f332b5 # v2.2.0
+        with:
+          version: latest
+          static: ${{ matrix.os.static }}
+      # TODO: windows tests are failing, but we should still run the built
```
anderseknert: build :)

internal/lsp/uri/uri.go
```diff
@@ -11,12 +11,15 @@ import (
 
 var drivePattern = regexp.MustCompile(`^\/?([A-Za-z]):`)
 
+// uris always use / as the uriSeparator, regard.
```
anderseknert: Sounds like a const then.

PR #1684, opened 2025-09-08
Weekend refactor

anderseknert:
More changes I've been playing around with. Some things have been late night additions, so let me know if something is too much :)

### DIFF
--- a/cmd/languageserver.go
+++ b/cmd/languageserver.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
-	"path/filepath"
 	"syscall"
 
 	"github.com/spf13/cobra"
@@ -18,7 +17,7 @@ import (
 )
 
 func init() {
-	verboseLogging := false
+	verbose := false
 
 	languageServerCommand := &cobra.Command{
 		Use:   "language-server",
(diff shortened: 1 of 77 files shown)


cmd/languageserver.go
```diff
@@ -29,34 +28,23 @@ func init() {
 			ctx, cancel := context.WithCancel(context.Background())
 			defer cancel()
 
-			exe, err := os.Executable()
-			if err != nil {
+			if exe, err := os.Executable(); err != nil {
```
charlieegan3, line 31: Until recently, we used to have noinlineerr in regal

https://github.com/open-policy-agent/regal/blame/27d19e6f72a07c9cc94ce7e50564268543fed2b4/.golangci.yaml#L24

I personally preferred that but just my 2c

charlieegan3 (approved):

cmd/languageserver.go
anderseknert: I'm curious why that is? 

The Go authors didn't give us much in terms of handling errors in a non-verbose way, but the inline form at least saves us one line of scrolling per error, so I'll take it!

(that `err` also doesn't escape the scope of the `if` where it doesn't need to is another benefit of this)

charlieegan3: Mostly I was just checking in as we did have it for a long time in the project and I grew used to it I suppose.

I guess I can see the scope benefit too.

But personally find the more verbose err := foo()\nif err != nil to be easier to read, particularly when foo() is long with more params etc.

No strong feelings.

anderseknert: Yeah I think it’s mostly getting used to a new habit. Agreed on the long lines! I have commited a few of those sins I think. Better break those up into two lines where possible, or if not use the longer err check form. 

Sadly there doesn’t seem to be a linter to check for the opposite! Ie that the check is unlined where possible 😄 I don’t think we’d want to run it by default, but could be useful to do occasionally

PR #1686, opened 2025-09-09
Fix documentLink issue when no user config provided

anderseknert:
Pretty sure the documentLink issue Charlie found was only the tip of the iceberg, as we did a lot of initialization logic seemingly only when a custom user config was found..

Good to have that fixed!

Also, update references to docs.styra.com to now point at their new OPA docs home.

### DIFF
--- a/internal/lsp/rego/query/query.go
+++ b/internal/lsp/rego/query/query.go
@@ -1,11 +1,229 @@
 package query
 
+import (
+	"context"
+	"fmt"
+	"os"
+	"regexp"
+	"sync"
+
+	"github.com/open-policy-agent/opa/v1/ast"
+	"github.com/open-policy-agent/opa/v1/bundle"
+	"github.com/open-policy-agent/opa/v1/rego"
+	"github.com/open-policy-agent/opa/v1/resolver"
+	"github.com/open-policy-agent/opa/v1/storage"
+	"github.com/open-policy-agent/opa/v1/topdown"
+
+	rbundle "github.com/open-policy-agent/regal/bundle"
+	"github.com/open-policy-agent/regal/internal/compile"
+	"github.com/open-policy-agent/regal/internal/util"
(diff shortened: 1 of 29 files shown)


internal/lsp/server_signature_help_test.go
```diff
@@ -61,10 +58,9 @@ allow if concat(",", "a", "b") == "b,a"`
 			t.Parallel()
 
 			ls := NewLanguageServer(ctx, &LanguageServerOptions{Logger: log.NewLogger(log.LevelDebug, t.Output())})
-			ls.loadedConfig = &config.Config{}
 			ls.cache.SetFileContents(mainRegoURI, content)
 
-			testutil.NoErr(PutBuiltins(ctx, ls.regoStore, builtins))(t)
+			// testutil.NoErr(PutBuiltins(ctx, ls.regoStore, builtins))(t)
```
srenatus: hm?

srenatus (approved):
LGTM

internal/lsp/server_signature_help_test.go
anderseknert: Sorry, that should just be deleted. The built-ins are now loaded in the language server constructor already. I'll do that.

internal/lsp/rego/query/query.go
```diff
@@ -1,11 +1,229 @@
...
+				added.Add(s.Schema.String())
+			}
+		}
+	}
+
+	return resolvers
+})
+
+// Eval implements the resolver.Resolver interface to resolve schemas from annotations at runtime.
+func (sr schemaResolver) Eval(_ context.Context, _ resolver.Input) (resolver.Result, error) {
```
srenatus: [nit]
```suggestion
func (sr schemaResolver) Eval(context.Context, resolver.Input) (resolver.Result, error) {
```

anderseknert: Fixed 👍

PR #1692, opened 2025-09-18
lsp: Fix missing initial root

charlieegan3:
Fixes https://github.com/open-policy-agent/regal/issues/1671

when there is a user level config
Screenshot 2025-09-18 at 11 20 10

when there is no config at all
Screenshot 2025-09-18 at 11 19 06

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -218,7 +218,7 @@ func (l *LanguageServer) Handle(ctx context.Context, _ *jsonrpc2.Conn, req *json
 	case "textDocument/diagnostic":
 		return l.handleTextDocumentDiagnostic()
 	case "textDocument/didOpen":
-		return handler.WithParams(req, l.handleTextDocumentDidOpen)
+		return handler.WithContextAndParams(ctx, req, l.handleTextDocumentDidOpen)
 	case "textDocument/didClose":
 		return handler.WithParams(req, l.handleTextDocumentDidClose)
 	case "textDocument/didSave":
@@ -875,19 +875,21 @@ func (l *LanguageServer) loadConfig(ctx context.Context, conf config.Config) {
 	}
 
 	// Rego versions may have changed, so reload them.
-	allRegoVersions, err := config.AllRegoVersions(l.workspacePath(), &conf)
-	if err != nil {
-		l.log.Debug("failed to reload rego versions: %s", err)
-	} else {
-		l.loadedConfigAllRegoVersions.Clear()
+	if l.workspacePath() != "" {
+		allRegoVersions, err := config.AllRegoVersions(l.workspacePath(), &conf)
+		if err != nil {
+			l.log.Debug("failed to reload rego versions: %s", err)
+		} else {
+			l.loadedConfigAllRegoVersions.Clear()
 
-		for k, v := range allRegoVersions {
-			l.loadedConfigAllRegoVersions.Set(k, v)
+			for k, v := range allRegoVersions {
+				l.loadedConfigAllRegoVersions.Set(k, v)
+			}
 		}
 	}
 
 	// Enabled rules might have changed with the new config, so reload.
-	if err = l.loadEnabledRulesFromConfig(ctx, conf); err != nil {
+	if err := l.loadEnabledRulesFromConfig(ctx, conf); err != nil {
 		l.log.Message("failed to cache enabled rules: %s", err)
 	}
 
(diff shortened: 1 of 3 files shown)


internal/lsp/server.go
```diff
@@ -1528,7 +1530,21 @@ func (l *LanguageServer) handleTextDocumentDefinition(params types.DefinitionPar
...
-func (l *LanguageServer) handleTextDocumentDidOpen(params types.DidOpenTextDocumentParams) (any, error) {
+func (l *LanguageServer) handleTextDocumentDidOpen(
+	ctx context.Context,
+	params types.DidOpenTextDocumentParams,
+) (any, error) {
+	// then we have started the server, and not yet received a suitable root to use.
+	if l.workspaceRootURI == "" {
+		err := l.updateRootURI(ctx,
+			// get the URI of the file's immediate parent
+			uri.FromPath(l.client.Identifier, filepath.Dir(uri.ToPath(l.client.Identifier, params.TextDocument.URI))),
```
anderseknert: You can use `l.fromPath` and `l.toPath` here

internal/lsp/server.go
```diff
@@ -2029,45 +2011,84 @@ func (l *LanguageServer) handleInitialize(ctx context.Context, params types.Init
...
+	// consistency
+	normalizedRootURI := strings.TrimSuffix(rootURI, string(os.PathSeparator))
 
-		if configFilePath != "" {
-			l.log.Message("using config file: %s", configFilePath)
-			l.configWatcher.Watch(configFilePath)
-		} else {
-			l.log.Message("no config file found for workspace")
-		}
+	configRoots, err := lsconfig.FindConfigRoots(uri.ToPath(l.client.Identifier, normalizedRootURI))
```
anderseknert: Same here and later in this function.

internal/lsp/server.go
```diff
@@ -2029,45 +2011,84 @@ func (l *LanguageServer) handleInitialize(ctx context.Context, params types.Init
 		l.log.Message("failed to cache enabled rules: %s", err)
 	}
 
-	if l.workspaceRootURI != "" {
-		workspaceRootPath := l.workspacePath()
+	if params.RootURI != "" {
+		err := l.updateRootURI(ctx, params.RootURI)
```
anderseknert, line 2015: Good!

internal/lsp/server.go
```diff
@@ -2029,45 +2011,84 @@ func (l *LanguageServer) handleInitialize(ctx context.Context, params types.Init
 		l.log.Message("failed to cache enabled rules: %s", err)
 	}
 
-	if l.workspaceRootURI != "" {
-		workspaceRootPath := l.workspacePath()
+	if params.RootURI != "" {
```
anderseknert, line 2014: Should we also look for this in the workspaceFolders param?
The spec suggests that's where it should be, and we've mostly been lucky that clients don't seem to honor that https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initializeParams

(and it's easy to see why they don't given how poorly that is documented)

Not suggesting we handle multiple workspaces as part of this, but we could perhaps just check if there's _something_ there, and if there is we go with the first one (if nothing else was given).

internal/lsp/server_no_workspace_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+	mainRegoContents := `package main
+
+import  rego.v1
+allow =  true
+`
+
+	files := map[string]string{
+		"foo/main.rego": mainRegoContents,
+		// here we are ignoring two issues with the file, so we expect no to
```
anderseknert: no -> not :)

internal/lsp/server_no_workspace_test.go
```diff
@@ -0,0 +1,79 @@
...
+      level: ignore
+`,
+	}
+
+	// set up the workspace content with some example rego and regal config
+	tempDir := testutil.TempDirectoryOf(t, files)
+	mainRegoURI := uri.FromPath(clients.IdentifierGeneric, filepath.Join(tempDir, filepath.FromSlash(mainRegoFileName)))
+
+	receivedMessages := make(chan types.FileDiagnostics, defaultBufferedChannelSize)
+	clientHandler := test.HandlerFor(methodTdPublishDiagnostics, test.SendsToChannel(receivedMessages))
```
anderseknert, line 45: Humble brag, but these helpers are so nice!

charlieegan3: yeah, big improvement

charlieegan3:
Thanks for the review, I have made a number of edits in a second commit: 60e9e0124d1b89f76090c0c9aa9b3f6d888e2d87

internal/lsp/server.go
charlieegan3: yeah that's fair, I have something in there now

charlieegan3: noted in a comment that it's untested but based on the spec

PR #1699, opened 2025-09-25
Move completion handler into Rego router

anderseknert:
Although many completion providers have been Rego based since a long time back, these providers used their own input format, and a separate prepared query — not at all integrated with other Rego handlers. No more! All completion providers are now 100% Rego-based, and use the same Rego router entrypoint as all other Rego handlers. Since the input format now mirrors what the LSP client sends, a whole lot of tests had to be updated.

Also the usual style fixes here and there.. although I tried to show some restraint :)


internal/lsp/server.go
```diff
@@ -1443,31 +1442,6 @@ func (l *LanguageServer) handleTextDocumentInlayHint(params types.InlayHintParam
 	return inlayhint.FromModule(module, bis), nil
 }
 
-func (l *LanguageServer) handleTextDocumentCompletion(ctx context.Context, params types.CompletionParams) (any, error) {
```
charlieegan3, line 1446: 🎉

anderseknert: Up next is the rest of the text document handlers. But first, we feast! 🎉

PR #1718, opened 2025-10-18
lsp: document highlighting supporting function args

anderseknert:
A function arg variable at requested position will now have the document highlight handler return all references of that arg from the function head/body for highlighting, making it easy to quickly identify the locations where the arg is used.

Less interesting to users is the *much* improved `REGAL_BUNDLE_PATH` development mode added with this PR. This is the result of a substantial side quest when working on the highlighting feature, and finding the current implementation to perform too poorly. Which was no surprise given how previously we'd re-prepare the query on every call to `LoadedBundle`, which sometimes happens several times per second!  The new implementation instead makes use of a watcher process, and re-prepares the relevant LSP query only when a file in the bundle is changed and that change is saved.

Also properly documented the dev mode feature as it now makes LSP development so much more enjoyable.

Finally, renamed a few methods to avoid duplicating names from the package (like `config.FindConfig` -> `config.Find`)

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -41,7 +41,7 @@ Regal.
 - A rule that misses a few cases is better than no rule at all, but it's good to document any known edge cases
 - False positives should however always be avoided
 - Add tests for as many cases as you can think of
-- Any new rule should have an example violation added in `e2e/testada/violations/most_violations.rego`
+- Any new rule should have an example violation added in `e2e/testdata/violations/most_violations.rego`
 - All the steps for building, testing and linting in this document should pass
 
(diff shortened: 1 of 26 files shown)


docs/CONTRIBUTING.md
```diff
@@ -65,6 +65,36 @@ the policy you're working on, providing an extremely fast feedback loop for deve
 
 ![Use AST of file as input](./assets/lsp/eval_use_as_input.png)
 
+## Contributing to the Language Server
+
+Since the linter is also a core component of the language server, it's likely a good strategy to first understand how
```
charlieegan3, line 70: thanks for writing this up!

anderseknert: It's such a great workflow now, and I'm excited to have you try it out!
And it would be a shame if only we knew about it 🙂

PR #1722, opened 2025-10-24
lsp: Implement selection ranges and linked editing range

anderseknert:
Selection ranges provide smart selection of code that can expand and shrink based on knowledge of the AST rather than simple text properties like hyphens or whitespace. I didn't think this would be particularly important, but having it used it for some time while working on this feature, this is definitely a feature that I now wouldn't want to be without. While not likely to be used much outside of rules, ranges also expand in both package declarations and imports.

https://github.com/user-attachments/assets/9fd98c3c-eaa9-4922-a16a-aed7778ad0e5

Linked editing range OTOH was mostly implemented to try out the feature, and at this point has not proven to be particularly useful, and the feature is therefore kept behind a new `REGAL_EXPERIMENTAL` flag for now. More info about that in the docs included in this PR.

Also:
- Remove the "log only once" logic in the development bundle loader, as it's no longer flooding the logs
- Use shared schema for all `TextDocumentPosition` params handlers
- Document all environment variables available for development and debugging purposes


bundle/regal/lsp/util/location/location.rego
```diff
@@ -14,3 +14,42 @@ to_range(location) := {
 		"character": location.end.col - 1,
 	},
 }
+
+# METADATA
+# description: |
+#   parse location string sl to LSP Range object
+#   example:
+#   "1:5:1:10" -> {"start": {"line":0,"character":4},"end":{"line":0,"character":9}}
+parse_range(sl) := {
```
charlieegan3, line 23: `sl` - string location? short location?

bundle/regal/lsp/util/location/location.rego
```diff
@@ -14,3 +14,42 @@ to_range(location) := {
 		"character": location.end.col - 1,
 	},
 }
+
+# METADATA
+# description: |
+#   parse location string sl to LSP Range object
```
charlieegan3, line 20:
```
#   parse location string sl to LSP Range object
                           ^

```

I'd make this the unpacked term in the comment here. but I am unsure what it is.

docs/CONTRIBUTING.md
```diff
@@ -197,3 +197,18 @@ wasmtime-cli 13.0.0
...
 ```
+
+## Environment variables for development and debugging
+
+The following environment variables can be used to modify the behavior of Regal, commonly for development purposes:
+
+- `REGAL_BUNDLE_PATH`: When set to a valid path, Regal will load its bundle from disk instead of using the embedded
+  bundle. See [Contributing to the Language Server](#contributing-to-the-language-server) for details.
+- `REGAL_EXPERIMENTAL`: When set to `true`, enables experimental features in Regal, such as
+  [linked editing ranges](./language-server.md#linked-editing-ranges) in the language server.
```
charlieegan3, line 208: Since this file is not on the website, should this link to https://www.openpolicyagent.org/projects/regal/language-server?

Since that's the canonical location for that file to be viewed.

bundle/regal/lsp/util/location/location.rego
anderseknert: Yes, string location. Might as well just rename it `str` as it should be clear what is being parsed. I'll fix before merge.

anderseknert:
Thanks Charlie! Yeah, it's amazing how we've gotten to this point, where entirely new LSP features can be implemented more or less without touching Go code. I am planning to refactor that location.rego file tonight, and so I'll make the naming edits you suggested as part of that.

PR #1724, opened 2025-10-26
Fix false positive in narrow-argument

### LINKED ISSUE #1701: False positive in ` narrow-argument` and arg array indices
```rego
infix_expr_location(terms) := location(sprintf("%s:%s:%s:%s", array.concat(
	array.slice(split(terms[1].location, ":"), 0, 2),
	array.slice(split(terms[0].location, ":"), 2, 4),
)))
```

Flagged as:
```
Argument terms always referenced by a common prefix, value passed can be narrowed to terms[0].location
```
Which is of course not correct.

anderseknert:
This rule did not properly handle function arg refs of arrays, and would incorrectly report that `f(x) if x[0] + x[1]` could be narrowed to `x[1]`.

Fixes #1701

### DIFF
--- a/bundle/regal/util/util.rego
+++ b/bundle/regal/util/util.rego
@@ -188,14 +188,20 @@ longest_prefix(coll) := [] if {
 	arr := to_array(coll)
 	end := min([count(seq) | some seq in arr]) - 1
 	rng := numbers.range(0, end)
-	eqn := max([n |
+
+	# collect indices where items differ
+	# we only care about the first diff, but no way to exit early with that value
+	dif := [n |
 		some n in rng
 
 		first := arr[0][n]
-		every sub in arr {
-			sub[n] == first
-		}
-	])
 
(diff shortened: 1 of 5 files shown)


bundle/regal/util/util.rego
```diff
@@ -188,14 +188,20 @@ longest_prefix(coll) := [] if {
 	arr := to_array(coll)
 	end := min([count(seq) | some seq in arr]) - 1
 	rng := numbers.range(0, end)
-	eqn := max([n |
+
+	# collect indices where items differ
+	# we only care about the first diff, but no way to exit early with that value
+	eqn := [n |
```
charlieegan3: `eqn` as a name could be clearer for me as a reader here.

anderseknert: Indeed, I'll see what I can come up with while keeping the 3 character uniformity :P

PR #1731, opened 2025-10-28
Fix false positive in use-some-for-output-vars

### LINKED ISSUE #1489: False positive in use-some-for-output-vars when var in ref head ref
```rego
rule[arr[i]] contains value if {
    arr := [1, 2, 3]

    some i
    value := input.foo[i]
}
```

This currently reports `i` in the ref head as not being declared using `some`, while clearly it was. This seems to have something to do with the var being used in a nested ref, as `rule[i]` works as expected. Interestingly, making that change will instead have Regal _correctly_ report `input.foo[i]` as a `use-some-in-iteration` violation... and that's somehow masked by `i` being in a nested ref. So two bugs really, but possibly related, so we'll see if there's a need for a separate issue for that later.

anderseknert:
Which was reported for nested ref head vars

Fixes #1489

### DIFF
--- a/bundle/regal/rules/idiomatic/use-some-for-output-vars/use_some_for_output_vars_test.rego
+++ b/bundle/regal/rules/idiomatic/use-some-for-output-vars/use_some_for_output_vars_test.rego
@@ -160,3 +160,10 @@ test_success_not_an_output_var if {
 
 	r == set()
 }
+
+# verify fix for https://github.com/open-policy-agent/regal/issues/1489
+test_success_nested_ref_head_var_declared if {
+	r := rule.report with input as ast.policy(`rule[input[i]] contains i if some i`)
+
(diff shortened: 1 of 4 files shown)


bundle/regal/rules/idiomatic/use-some-for-output-vars/use_some_for_output_vars_test.rego
```diff
@@ -160,3 +160,10 @@ test_success_not_an_output_var if {
 
 	r == set()
 }
+
+# verify fix for https://github.com/open-policy-agent/regal/issues/1489
+test_succes_nested_ref_head_var_declared if {
```
charlieegan3:
```suggestion
test_success_nested_ref_head_var_declared if {
```

PR #1732, opened 2025-10-28
lsp: Use filename path for current buffer for defns

charlieegan3:
There was a bug here where the defn would only work when the result was another file. Now we use the same format of reference for the current buffer as we use for other modules (a path rather than a URI)

Fixes https://github.com/open-policy-agent/regal/issues/1700

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"path"
 	"path/filepath"
 	"regexp"
 	"slices"
@@ -1511,7 +1510,7 @@ func (l *LanguageServer) handleTextDocumentDefinition(params types.DefinitionPar
 
 	query := oracle.DefinitionQuery{
 		// The value of Filename is used if the defn in the current buffer.
-		Filename: strings.TrimPrefix(params.TextDocument.URI, util.EnsureSuffix(l.workspaceRootURI, "/")),
+		Filename: l.toRelativePath(params.TextDocument.URI),
 		Pos:      positionToOffset(contents, params.Position),
 		Modules:  modules,
 		Buffer:   outil.StringToByteSlice(contents),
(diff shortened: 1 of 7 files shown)


anderseknert (approved):
Thank you! This not working has been a real pain in the neck, so this is a much welcome fix <3 Perhaps worth breaking out the `Filename` and `URI` value creation just so that we can test them? And make sure we can't change them accidentally without also having to change tests (which likely will alert us about breaking something)?

(even if it means some pretty silly tests)

charlieegan3:
The tests are a pretty similar to the existing URI tests, but I've added some new helpers for the relative operations. We're doing these in a few other places too, if mostly in tests. Lmk if that look ok.

charlieegan3:
One thing I dislike is the client identifier functionality. I think I perhaps added that prematurely when after noticing some minor differences in VSCode. I think my preference is to keep it for now, and to do a review of client specific behaviour separately.

internal/lsp/server.go
```diff
@@ -1510,7 +1510,7 @@ func (l *LanguageServer) handleTextDocumentDefinition(params types.DefinitionPar
 
 	query := oracle.DefinitionQuery{
 		// The value of Filename is used if the defn in the current buffer.
-		Filename: params.TextDocument.URI,
+		Filename: uri.ToRelativePath(l.client.Identifier, params.TextDocument.URI, l.workspaceRootURI),
```
anderseknert: There's also
```go
func (l *LanguageServer) toRelativePath(fileURI string) string {
	return strings.TrimPrefix(l.toPath(fileURI), l.workspacePath()+string(os.PathSeparator))
}
```
Which could be updated to use your new function, and then we can use that here instead

charlieegan3: Ahh! I missed that one...

charlieegan3: OK, updated that function now.

PR #1733, opened 2025-10-28
Fix incorrect behavior in pointless-import

### LINKED ISSUE #1728: pointless-import rule incorrectly flags imports of subpackages
### Description
The `pointless-import` rule incorrectly flags imports of subpackages as pointless when they are actually necessary. Specifically, importing a subpackage that has its own separate package declaration should be valid and not flagged as a pointless import.

### Example
In my case, I have:
- Main package: `package validation`  
- Subpackage: `package validation.helpers` (defined in a separate file)

When I import the helpers subpackage:
```rego
package validation

import validation.helpers
```

Regal incorrectly flags this with:
```
Rule:         	pointless-import                                                             	
Description:  	Importing own package is pointless                                           	
Category:     	imports                                                                      	
Location:         validation.rego:3:8                            	
Text:         	import data.validation.helpers                            	
Documentation:	https://www.openpolicyagent.org/projects/regal/rules/imports/pointless-import	
```

### Expected Behavior
The import should be considered valid because:
1. `helpers` is a separate subpackage with its own `package validation.helpers` declaration
2. The functions in the helpers package need to be imported to be accessible
3. This is a legitimate use case for organizing code into subpackages

### Actual Behavior
The rule flags this as a pointless import, suggesting it's importing "own package" when it's actually importing a subpackage.

### Workaround
Currently working around this by using the full path `data.validation.helpers.function_name` instead of importing, but this makes the code less readable.

### Environment
- Regal version: [latest as of Oct 27, 2025]
- File structure:
```
.
└── validation
    ├── helpers
    │   └── helpers.rego
    └── validation.rego
```

```rego
# ./validation/validation.rego
package validation

import data.validation.helpers

errors contains helpers.some_check
```

```rego
# ./validation/helpers/helpers.rego
package validation.helpers

some_check if {
	input != null
}

```

The rule should distinguish between importing the exact same package vs importing a legitimate subpackage.

anderseknert:
The intention was to flag imports of own package, and imports of paths defined in the same **module**, not the same package (i.e. across files), which is now what this rule does.

Credits to @sascha-cariad for reporting this, and in a very detailed way!

Fixes #1728

### DIFF
--- a/bundle/regal/ast/ast.rego
+++ b/bundle/regal/ast/ast.rego
@@ -183,12 +183,6 @@ is_output_var(rule, var) if {
 }
 
 # METADATA
-# description: as the name implies, answers whether provided value is a ref
-# scope: document
-is_ref(value) if value.type == "ref"
-is_ref(value) if value[0].type == "ref"
-
-# METADATA
 # description: |
 #   returns an array of all rule indices, as strings. this will be needed until
 #   https://github.com/open-policy-agent/opa/issues/6736 is fixed
@@ -234,6 +228,18 @@ ref_value_equal(v1, v2) if {
 		part.value == v2[i].value
 	}
 }
+
+# METADATA
+# description: |
+#   returns a new ref value made by extending terms1 with terms2,
+#   where the first term of terms2 transformed to string
+extend_ref_terms(terms1, terms2) := array.concat(
+	terms1,
+	array.concat(
+		[object.union(terms2[0], {"type": "string"})],
+		array.slice(terms2, 1, 100),
+	),
+)
 
 # METADATA
 # description: |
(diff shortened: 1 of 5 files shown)


bundle/regal/ast/ast.rego
```diff
@@ -182,12 +182,6 @@ is_output_var(rule, var) if {
 	num_above - num_some == 0
 }
 
-# METADATA
-# description: as the name implies, answers whether provided value is a ref
-# scope: document
-is_ref(value) if value.type == "ref"
-is_ref(value) if value[0].type == "ref"
```
anderseknert, line 189: This is unrelated — just some functions being old and unused.

anderseknert:
@sascha-cariad the binaries built from this PR can be found at the bottom of [this page](https://github.com/open-policy-agent/regal/actions/runs/18885949230). If you would like to verify everything works as you'd expect before this goes out in the next release (later this week or next).

sascha-cariad:
Works! ❤️ 

```bash
.
├── regal
└── validation
    ├── helpers
    │   └── helpers.rego
    └── validation.rego

3 directories, 3 files

❯ ./regal lint .
2 files linted. No violations found.
```

PR #1736, opened 2025-10-29
Various small fixes

anderseknert:
- Add debugging configuration to enable inlineValues in project
- Don't use hardcoded "/" in eval test paths (as that won't work on Windows)
- More efficient input creation in aggregate lint step

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -17,5 +17,13 @@
     // This is better handled by Regal's selection ranges implementation
     "editor.smartSelect.selectLeadingAndTrailingWhitespace": false,
     "editor.smartSelect.selectSubwords": false
-  }
+  },
+  // Enable to have values displayed inline during debugging
+  // The default value is "auto", which consults the language extension on whethher
+  // to have this enabled or not. And since the extension currently doesn't do anything
+  // on its own to support this, we need to explicitly enable it here if we want this.
+  // Long term, the extension + Regal should provide support for this, as it'll also
+  // result in a better experience when we can be more selective about what to show,
+  // and how.
+  "debug.inlineValues": "on"
 }
--- a/internal/lsp/eval_test.go
+++ b/internal/lsp/eval_test.go
@@ -106,16 +106,19 @@ func TestFindInputPath(t *testing.T) {
 				t.Fatalf("did not expect to find input.%s", tc.fileExt)
 			}
 
-			createWithContent(t, tmpDir+"/workspace/foo/bar/input."+tc.fileExt, tc.fileContent)
+			inputPath := filepath.Join(workspacePath, "foo", "bar", "input."+tc.fileExt)
+			createWithContent(t, inputPath, tc.fileContent)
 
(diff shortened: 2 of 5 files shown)


.vscode/settings.json
```diff
@@ -17,5 +17,13 @@
     // This is better handled by Regal's selection ranges implementation
     "editor.smartSelect.selectLeadingAndTrailingWhitespace": false,
     "editor.smartSelect.selectSubwords": false
-  }
+  },
+  // Enable to have values displayed inline during debugging
+  // The default value is "auto", which consults the language extension on whethher
```
srenatus, line 22:
```suggestion
  // The default value is "auto", which consults the language extension on whether
```

internal/lsp/eval_test.go
```diff
@@ -106,16 +106,19 @@ func TestFindInputPath(t *testing.T) {
 				t.Fatalf("did not expect to find input.%s", tc.fileExt)
 			}
 
-			createWithContent(t, tmpDir+"/workspace/foo/bar/input."+tc.fileExt, tc.fileContent)
+			inputPath := filepath.Join(workspacePath, "foo", "bar", "input."+tc.fileExt)
```
srenatus, line 109: `filepath.Join` is quite clever, I think you should be able to do this, too:
```suggestion
			inputPath := filepath.Join(workspacePath, "foo/bar/input."+tc.fileExt)
```
`Join` also calls `Clean`, so it'll adjust your `/` and `\` as needed, IIRC.

pkg/linter/linter.go
```diff
@@ -869,27 +869,22 @@ func (l Linter) lintWithAggregateRules(
...
-			},
-		},
-	}
+	regal := ast.ObjectTerm(
+		ast.Item(ast.InternedTerm("operations"), ast.ArrayTerm(ast.InternedTerm("aggregate"))),
+		ast.Item(ast.InternedTerm("file"), ast.ObjectTerm(
+			ast.Item(ast.InternedTerm("name"), ast.InternedTerm("__aggregate_report__")),
+			ast.Item(ast.InternedTerm("lines"), ast.InternedEmptyArray),
+		)),
+	)
```
srenatus, line 878: 🤔 I wonder if the compiler will raise this out of the function body, since it's constant... alternatively, you could do that, if this is called more than once, perhaps in the LSP scenario.

pkg/linter/linter_test.go
srenatus, line 449:
```suggestion
	policies := map[string]string{
		"foo.rego": `package foo
		import data.bar

		default allow := false
	`,
		"bar.rego": `package bar
		import data.foo.allow
	`,
	}
```

charlieegan3 (approved):

pkg/linter/linter.go
anderseknert: Yes, definitely for the LSP scenario! Will update 👍

anderseknert:
@srenatus I'll come back to address your (all valid!) points later. Highest priority now is to just get as many of the windows tests to work before we prepare the next release, so I merged this so @charlieegan3 (who's also working on this) and I have a shared state.

PR #1737, opened 2025-10-29
windows: Correct use of slash paths

charlieegan3:
Address number of test bugs on windows due to use of / in paths.

### DIFF
--- a/internal/lsp/config/find_test.go
+++ b/internal/lsp/config/find_test.go
@@ -77,10 +77,7 @@ func TestFindConfigRoots(t *testing.T) {
 				gotTrimmed := make([]string, len(got))
 
 				for i, path := range got {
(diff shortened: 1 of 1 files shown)


pkg/fixer/fileprovider/cache.go
```diff
@@ -28,7 +28,7 @@ func NewCacheFileProvider(c *cache.Cache, ci clients.Identifier) *CacheFileProvi
 		ClientIdentifier: ci,
 		modifiedFiles:    util.NewSet[string](),
 		deletedFiles:     util.NewSet[string](),
-		toPath:           util.Partial2(uri.ToPath, ci),
+		toPath:           uri.ToPath,
```
srenatus, line 31: Looks like we could just use `uri.ToPath` instead of the field `toPath` then?

PR #1738, opened 2025-10-30
More path handling improvements

anderseknert:
Hopefully taking us a little closer to green tests on Windows.

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 
 	rio "github.com/open-policy-agent/regal/internal/io"
+	"github.com/open-policy-agent/regal/internal/util"
 	"github.com/open-policy-agent/regal/pkg/config"
 )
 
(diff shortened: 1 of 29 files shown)


internal/compile/compile.go
```diff
@@ -33,7 +33,7 @@ func NewCompilerWithRegalBuiltins() *ast.Compiler {
 // Currently only used by the test command. Should we want to expand the use of this later,
 // we'll probably want to only read the schemas relevant to the context.
 var RegalSchemaSet = sync.OnceValue(func() *ast.SchemaSet {
-	schemaSet, _ := files.DefaultWalkReducer("schemas", ast.NewSchemaSet()).
+	return util.FirstValue(files.DefaultWalkReducer("schemas", ast.NewSchemaSet()).
```
srenatus, line 36: Hm. I guess beauty is in the eye of the beholder, but I think `util.FirstValue` is taking it a bit too far. `x, _ :=  ....` seems clear enough, and less typing, too.

cmd/utils.go
```diff
@@ -39,18 +40,12 @@ func (f *repeatedStringFlag) Set(s string) error {
...
-		if !rio.Exists(searchPath) {
-			searchPath = "" // This is handled elsewhere — we don't need to fail here
+		if abs, _ := filepath.Abs(args[0]); rio.Exists(abs) {
+			searchPath = abs
 		}
 	}
 
-	searchPath = cmp.Or(searchPath, rio.Getwd())
-	if searchPath == "" {
+	if searchPath = cmp.Or(searchPath, rio.Getwd()); searchPath == "" {
```
srenatus, line 48: I don't see the benefit of putting that into the `if`. It's not reducing the scope or anything.

charlieegan3:
Hmm, there seem to be 32 failures here. by 23 in https://github.com/open-policy-agent/regal/actions/runs/18918176175/job/54007011415 (now on main...)

internal/compile/compile.go
anderseknert: You don't have to love it, but it's hardly more typing? I made it specifically for this:

Screenshot 2025-10-30 at 12 36 11

Where the second return value completely breaks composability. The other locations were mostly just a bonus once I had the function to play with.

anderseknert:
Closing this to push a new PR only with the non-windows fixes.

internal/compile/compile.go
srenatus: Yeah that one is OK. The other one is where I wasn't convinced: 

Image

PR #1741, opened 2025-10-30
Use generic client identifier in tests

charlieegan3:


.github/workflows/build.yaml
```diff
@@ -135,8 +135,6 @@ jobs:
           static: ${{ matrix.os.static }}
       # TODO: windows tests are failing, but we should still run the build & tests for inspection
       - run: go build
-        continue-on-error: true
       - run: go test ./...
-        continue-on-error: true
```
anderseknert, line 140: 🎉 !!!

PR #1744, opened 2025-11-03
deps: stick to github.com/vektah/gqlparser/v2@v2.5.30

srenatus:
Because the latest version would require a change to OPA's YAML test cases. For projects that use both Regal and OPA, like EOPA, it's perhaps better to have OPA control the version of third party libs.

The extra field that v2.5.31 returns is `ExpectedTypeHasDefault`, introduced in https://github.com/vektah/gqlparser/commit/5b4126931203c088dfcbb0b00a7af0813ef21694.

Tl;dr:


anderseknert (approved):
Yeah, what's messed up here is either the project trying to keep its dependencies up to date, or it's the one introducing breaking changes in _patch_ releases. Hard to say.

PR #1747, opened 2025-11-03
docs: Minor fixes to docs typos and links

charlieegan3:
Also made the debugging page title more accessible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 <!-- markdownlint-disable MD041 -->
 
 [![Build Status](https://github.com/open-policy-agent/regal/workflows/Build/badge.svg)](https://github.com/open-policy-agent/regal/actions)
-![OPA v1.10.0](https://openpolicyagent.org/badge/v1.10.0)
+![OPA v1.10.0](https://www.openpolicyagent.org/badge/v1.10.0)
 [![codecov](https://codecov.io/github/StyraInc/regal/graph/badge.svg?token=EQK01YF3X3)](https://codecov.io/github/StyraInc/regal)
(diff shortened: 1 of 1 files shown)


docs/custom-rules/index.md
```diff
@@ -76,7 +76,7 @@ In order to view the JSON AST representation of a policy, use the `regal parse`
 If we were to write the simplest policy possible, and parse it using `regal parse`, it would contain nothing but a
 package declaration:
 
-**policy.rego**
+`policy.rego`
```
anderseknert: Any particular reason for these changes? I don't mind much, but I think it makes the name easier to miss, as it camouflaged next to the markdown block below.

Screenshot 2025-11-04 at 09 48 46

docs/debug-adapter.md
```diff
@@ -1,11 +1,11 @@
-# Debug Adapter
+# Debugging Rego
```
anderseknert, line 1: Excellent

anderseknert (approved):
That's a lot of missing `www`s 😄 And some very good corrections 👏 There's a markdown linter failure in CI though

docs/custom-rules/index.md
charlieegan3: I just wanted to them to be consistent with other filenames where we use `.

We could also use `title="foo.rego"` which is what looks best in docusaurus IMO, perhaps I'll do that.

charlieegan3: https://www.openpolicyagent.org/docs/ocp#2-define-a-bundle is an example of how it looks.

charlieegan3: I have updated these two places to do this instead.

anderseknert: Nice! Thanks

PR #1760, opened 2025-11-10
explorer: allow hiding stages without an effect (+misc)

### LINKED ISSUE #1756: Add compiler explorer option to hide collapsed stages
We often use screenshots from the compiler explorer to demonstrate how something is rewritten by the OPA compiler. This is sometimes made more difficult by the fact the stages of interest won't fit on a single screen when the explorer lists all the stages in between — including those that made no difference for the result. As an example:

```rego
package p

allow if not "a" in input.foo.bar
```

Displays as:

Image

With the relevant output found just further down:

Image

This isn't a huge issue, but I think it would be great if there was a toggle to completely hide all stages without any output, to better visualize the stages where the action is.

srenatus:
Also changes the web page to not feature an editable policy -- you have that in your IDE when you get to that page already.

Far from perfect, but better than before:

https://github.com/user-attachments/assets/f00bb08d-f646-4cc4-bc74-a7e0e7dc8d00

Fixes #1756

### DIFF
--- a/internal/web/assets/main.tpl
+++ b/internal/web/assets/main.tpl
@@ -19,7 +19,7 @@
         type="checkbox"
         name="hide_identical"
         id="hide_identical">
-      <label for="hide_identical">hide stages without effect on code</label>
+      <label for="hide_identical">Hide stages without effect on code</label>
     </div>
   </form>
   <section id="output">


srenatus:
So many things one could do here... like an actual "diff" display for the stages -- because right now, who can really tell what happens between two stages if the policy is sufficiently large? 🔍

internal/web/assets/main.tpl
```diff
@@ -7,21 +7,25 @@
...
-                    hx-trigger="keyup changed delay:200ms"
-                    class="flex-grow:1 monospace"
-          >{{ .Code }}</textarea>
+      <input hx-get="?tmpl=output"
+        hx-target="#output"
+        hx-trigger="change"
+        type="checkbox"
+        name="hide_identical"
+        id="hide_identical">
+      <label for="hide_identical">hide stages without effect on code</label>
```
srenatus:
```suggestion
      <label for="hide_identical">Hide stages without effect on code</label>
```

PR #1764, opened 2025-11-11
Rename: encoding/util -> encoding/write

anderseknert:
It was annoying having more than one `util` package in Regal, as completion providers would randomly suggest on or the other. The new name better describes the purpose anyway. Also fix some new issues reported by golangci-lint linters.


anderseknert:
@srenatus certainly. I'll just wait for the issue I found to be fixed first. Not sure if I'm allowed to report it 🥴

PR #1776, opened 2025-11-17
Enable more golangci-lint linters

anderseknert:
And disable a few that aren't relevant to us (SQL, etc)

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -24,10 +24,10 @@ jobs:
           - runner: ubuntu-latest
             name: linux
             static: true
-          - runner: macos-13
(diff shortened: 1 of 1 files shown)


charlieegan3:
I think this is failing because macOS-13 is deprecated now. I have a fix for that in #1777, but perhaps we can just add it here too.

charlieegan3:
I am unsure how to get rid of these phantom jobs without opening a new PR.

Screenshot 2025-11-18 at 15 52 56

anderseknert:
@charlieegan3 I'll go ahead and will merge without those ever getting green. Thanks for the fixes!

PR #1777, opened 2025-11-18
lsp: Add code action and command to ignore rule

charlieegan3:
Fixes https://github.com/open-policy-agent/regal/issues/622

https://github.com/user-attachments/assets/a9819c80-c87a-4f73-b166-9f229a3a423b

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -551,7 +551,7 @@ func (l *LanguageServer) StartConfigWorker(ctx context.Context) {
 	}
 }
 
-func (l *LanguageServer) StartCommandWorker(ctx context.Context) { //nolint:maintidx
+func (l *LanguageServer) StartCommandWorker(ctx context.Context) {
 	// note, in this function conn.Call is used as the workspace/applyEdit message is a request, not a notification
 	// as per the spec. In order to be 'routed' to the correct handler on the client it must have an ID
 	// receive responses too.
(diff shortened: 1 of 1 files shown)


internal/lsp/server.go
```diff
@@ -1252,6 +1261,63 @@ func (l *LanguageServer) fixRenameParams(label, fileURI string) (types.ApplyWork
...
+		if err := os.MkdirAll(regalDir, 0o755); err != nil {
+			return fmt.Errorf("failed to create .regal directory: %w", err)
+		}
+
+		configPath = filepath.Join(regalDir, "config.yaml")
+	}
+
+	var currentContent string
+
+	if _, err := os.Stat(configPath); err == nil {
```
anderseknert: If the stat fails here, we'll just proceed as normal? That doesn't seem right. even if it is unlikely to fail. I'd just call os.ReadFile directly.

pkg/config/modify/modify.go
```diff
@@ -0,0 +1,143 @@
...
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+)
+
+// SetKey sets a key at the given path to the specified value.
+// Comments are preserved, but indentation is always 2.
```
anderseknert, line 13: As God intended for YAML.

Did you try this on a more "complex" configuration file, like our own?

pkg/config/modify/modify_test.go
```diff
@@ -0,0 +1,144 @@
...
+        files:
+          - "*_test.rego"`,
+			path:  []string{"rules", "style", "line-length", "level"},
+			value: "ignore",
+			expected: `# Configuration file
+rules:
+  style:
+    prefer-snake-case:
+      level: error # Important rule
+      # Additional configuration
```
anderseknert, line 60: Good!

internal/lsp/server.go
charlieegan3: Yeah that makes sense, fixed in db0ae2d3634846aee6159d8ffaf7cc0c24e8e621

pkg/config/modify/modify.go
charlieegan3: https://github.com/user-attachments/assets/d16b33db-865c-4be5-84e3-d7d429408d9a

it seems to work ok from what I can tell.

PR #1788, opened 2025-12-02
Optimize and improve code for built-in functions

anderseknert:
* Register Regal's built-in functions globally only once and in a single location. The various `rego` APIs for working with custom functions haven't been a great experience for us to work with, and consolidating our code for this simplifies things quite a bit. Note that we've tried to do this in the past but never got it quite working. Seems like it does now.
* Set `CanSkipBctx` to `true` on all the OPA built-in functions that we use internally, as we don't need the cancellation or inter-query caches, and building the context for each call isn't cheap.
* Pass a no-op external cancel function to eval to avoid having OPA start one per each file linted.

While working on this I also found an optimization opportunity in OPA, which I will submit a PR for too.


internal/lsp/eval.go
```diff
@@ -12,12 +12,14 @@ import (
...
 	rbundle "github.com/open-policy-agent/regal/bundle"
+	"github.com/open-policy-agent/regal/internal/exp"
 	rquery "github.com/open-policy-agent/regal/internal/lsp/rego/query"
 	"github.com/open-policy-agent/regal/internal/lsp/uri"
 	"github.com/open-policy-agent/regal/internal/util"
-	"github.com/open-policy-agent/regal/pkg/builtins"
 	"github.com/open-policy-agent/regal/pkg/config"
 	"github.com/open-policy-agent/regal/pkg/roast/transform"
+
+	_ "github.com/open-policy-agent/regal/pkg/builtins"
```
charlieegan3, line 22: We have this required now in many places, what guards are there (or could there be) for in future if we need this _ import but forget to do it?

anderseknert: Since they are functions, missing this import would typically result in OPA not compiling a policy that makes use of them. Certainly a risk, and you're right this is hard to future-proof, but at least it should fail rather loudly.

I believe the best remedy is to have a single entry point for all evaluation throughout the Regal codebase. As you know, I've been rather conscious about this already, and we're now down to... 3 or so? But ideally only a single Regal module wraps `rego`[^1] directly, and any code doing evals route through that. At that point, it should be simple to assert programmatically that no other code imports `rego` directly.

[^1]: Tbh, I'm increasingly considering the feasibility of using the even more "low level" `query` [API](https://github.com/open-policy-agent/opa/blob/main/v1/topdown/query.go) (which `rego` is largely a wrapper around) directly in Regal, as I think we're well past the point of convenience being more important than control. And as I think this PR demonstrates, the convenience is often quite inconvenient (for our use case).

internal/exp/exp.go
```diff
@@ -0,0 +1,29 @@
...
+
+// topDownCancelNoOp is a no-op implementation of the rego.TopDownCancel interface.
+// When no external cancel function is provided to evaluation, OPA instantiates one
+// itself (per evaluation, or file linted). This cost is negligible. More interesting is
+// the fact that OPA also starts a goroutine per cancel function it creates, launching
+// a "waitForDone" function that blocks (inside its goroutine) until evaluation completes.
+// While this comes with some overhead, it too is rather insignificant. It does however mean
+// that the type of multi-threaded evaluations we typically in the linter will have pprof
+// report 99% of all "blocked" time here, making it much harder to identify blocking we do
+// ourselves / care about.
```
srenatus, line 21: Still sounds like something we could fix upstream. "unreadable pprof output" sounds like reason enough

anderseknert: Indeed. This was just an easier fix for the time being :)

Issue #1734, opened 2025-10-28
Add fixer for prefer-equals-comparison rule


anderseknert:
This is a [brand new linter rule](https://github.com/open-policy-agent/regal/pull/1726), but the PR did not include a matching fixer. Let's have it added! Working with legacy Rego sometimes mean hundreds or thousands of `=` usages will be flagged, so not having to fix this manually is a huge improvement for modernization efforts.

This should hopefully be an easy one to do, as the fix is simply to add another `=` at the reported location. Which is more or less identical to what the use-assigment-operator fix does already.

Fix should work both using `regal fix` from the command line, as well as via a Code Action in the language server.

charlieegan3:
fixed in  #1790

PR #1790, opened 2025-12-02
Add fixer for prefer-equals-comparison rule

SeanLedford:
PR to add fixer for the prefer-equals-comparison rule, as well as adequate coverage and docs updates. Done to address issue #1734.

### DIFF
--- a/pkg/fixer/fixes/preferequalscomparison_test.go
+++ b/pkg/fixer/fixes/preferequalscomparison_test.go
@@ -3,6 +3,8 @@ package fixes
 import (
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
+
 	"github.com/open-policy-agent/regal/pkg/report"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/fixer/fixes/preferequalscomparison_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+			if !tc.fixExpected && len(fixResults) != 0 {
+				t.Fatalf("unexpected fix applied")
+			}
+
+			if !tc.fixExpected {
+				return
+			}
+
+			if tc.fixExpected && fixResults[0].Contents != tc.contentAfterFix {
```
charlieegan3: You might want to look into using cmp.Diff, that can be helpful for comparing strings. we use it in some other tests.

pkg/fixer/fixes/preferequalscomparison_test.go
```diff
@@ -0,0 +1,89 @@
...
+			contentAfterFix: "package test\n\nallow := true\n\nallow == true\nallow == false\n",
+			fixExpected:     true,
+			runtimeOptions:  &RuntimeOptions{Locations: []report.Location{{Row: 5, Column: 7}, {Row: 6, Column: 7}}},
+		},
+		"different columns": {
+			fc: &FixCandidate{
+				Filename: "test.rego",
+				Contents: "package test\n\nlong := true\n\nlonger = true\nlongest = false\n",
+			},
+			contentAfterFix: "package test\n\nlong := true\n\nlonger == true\nlongest == false\n",
```
charlieegan3: I think for some of these tests we might use a raw string instead.
```suggestion
			contentAfterFix: `package test
long := true

longer == true

longest == false
`,
```

pkg/fixer/fixes/fixes.go
```diff
@@ -22,6 +22,7 @@ func NewDefaultFixes() []Fix {
 		&NoWhitespaceComment{},
 		&DirectoryPackageMismatch{},
 		&NonRawRegexPattern{},
+		&PreferEqualsComparison{},
```
charlieegan3, line 25: Have you run 

```
go run main.go lint foo/wow.rego
go run main.go fix --force foo/wow.rego
```

?

There seems to be a slight difference in how the locations are matched which is causing the fix not to fire.

SeanLedford: Gotcha, addressed! Wondering how we feel about updating the rule itself to provide the exact column location of the violation :thinking:, the [use-assignment-operator](https://github.com/open-policy-agent/regal/blob/main/bundle/regal/rules/style/use-assignment-operator/use_assignment_operator.rego) rule does something like that it looks like.

PR #1794, opened 2025-12-09
Fixers for redundant-existence-check and constant-condition rules

SeanLedford:
Draft PR for redundant-existence-check and constant-condition rules! Will mark ready for review when both fixers are implemented

### DIFF
--- a/e2e/cli_test.go
+++ b/e2e/cli_test.go
@@ -455,13 +455,13 @@ allow { input == 1 }
 		"v1/main.rego": `package v1
 
 #comment
-allow if { input == 1 }
+allow if { true }
 `,
 		"unrelated.txt": `foobar`,
 	}
 
 	td := testutil.TempDirectoryOf(t, initialState)
-	exp := fmt.Sprintf(`15 fixes applied:
+	exp := fmt.Sprintf(`16 fixes applied:
 In project root: %[1]s
 bar/main.rego -> wow/foo-bar/baz/main.rego:
 - directory-package-mismatch
@@ -490,6 +490,7 @@ main.rego:
 
 In project root: %[3]s
 main.rego:
+- constant-condition
 - no-whitespace-comment
 - opa-fmt
 `, td, filepath.Join(td, "v0"), filepath.Join(td, "v1"))
@@ -529,7 +530,7 @@ allow if input == 1
 		"v1/main.rego": `package v1
 
 # comment
-allow if input == 1
+allow if {}
 `,
 		"unrelated.txt": `foobar`,
 	}
--- a/pkg/fixer/fixes/constantcondition_test.go
+++ b/pkg/fixer/fixes/constantcondition_test.go
@@ -19,9 +19,9 @@ func TestConstantCondition(t *testing.T) {
 	}{
 		"no change": {
 			fc:              &FixCandidate{Filename: "test.rego", Contents: "package test\n\nallow := true\n"},
-			contentAfterFix: "package test\n\nallow := true\n",
-			fixExpected:     false,
 			runtimeOptions:  &RuntimeOptions{},
+			fixExpected:     false,
+			contentAfterFix: "package test\n\nallow := true\n",
 		},
 		"no change because no location": {
 			fc: &FixCandidate{
@@ -33,14 +33,14 @@ allow if {
     endswith(input.user.email, "@acmecorp.com")
 }`,
 			},
+			runtimeOptions: &RuntimeOptions{},
+			fixExpected:    false,
 			contentAfterFix: `package test
 
 allow if {
     true
     endswith(input.user.email, "@acmecorp.com")
 }`,
-			fixExpected:    false,
-			runtimeOptions: &RuntimeOptions{},
 		},
 		"single change": {
 			fc: &FixCandidate{
@@ -52,13 +52,6 @@ allow if {
     endswith(input.user.email, "@acmecorp.com")
 }`,
 			},
-			contentAfterFix: `package test
-
-allow if {
-    
-    endswith(input.user.email, "@acmecorp.com")
-}`,
-			fixExpected: true,
 			runtimeOptions: &RuntimeOptions{
 				Locations: []report.Location{
 					{
@@ -68,6 +61,13 @@ allow if {
 					},
 				},
 			},
+			fixExpected: true,
+			contentAfterFix: `package test
+
+allow if {
+    
+    endswith(input.user.email, "@acmecorp.com")
+}`,
 		},
 		"bad change": {
 			fc: &FixCandidate{
@@ -79,13 +79,6 @@ allow if {
     endswith(input.user.email, "@acmecorp.com")
 }`,
 			},
-			contentAfterFix: `package test
-
-allow if {
-    true
-    endswith(input.user.email, "@acmecorp.com")
-}`,
-			fixExpected: false,
 			runtimeOptions: &RuntimeOptions{
 				Locations: []report.Location{
 					{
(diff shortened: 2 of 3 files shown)


go.mod
```diff
@@ -23,6 +23,7 @@ require (
 	github.com/sourcegraph/jsonrpc2 v0.2.1
 	github.com/spf13/cobra v1.10.1
 	github.com/spf13/pflag v1.0.10
+	golang.org/x/sync v0.18.0
```
charlieegan3: I am unsure where this change came from, if you checkout go.mod and run go mod tidy, do you still get this? what's your go version? (ideally it matches 1.25.0)

pkg/fixer/fixes/redundantexistencecheck_test.go
```diff
@@ -0,0 +1,138 @@
...
+    endswith(input.user.email, "@acmecorp.com")
+}`,
+			},
+			contentAfterFix: `package test
+
+employee if {
+    input.user.email
+    endswith(input.user.email, "@acmecorp.com")
+}`,
+			fixExpected:    false,
```
charlieegan3: This is the example from the docs right? I am unsure why it's not expected.

go.mod
SeanLedford: Yeah I was unsure about this as well since I didn't touch this dependency, but yeah checking out go.mod, and even main entirely and then running go mod tidy causes this update. My go version is 1.25.4, so I guess a few minor versions ahead?

pkg/fixer/fixes/redundantexistencecheck_test.go
SeanLedford: Yeah this is the bad change test, which for previous testing files looks like is intended to test for an incorrect location being provided for the fixer. So that's why I have this test not expecting a fix.

go.mod
charlieegan3: What happens if you run `go mod why golang.org/x/sync`

```
$ go mod why golang.org/x/sync
# golang.org/x/sync
(main module does not need package golang.org/x/sync)
```

pkg/fixer/fixes/redundantexistencecheck_test.go
charlieegan3: Ahh gotcha, thanks!

go.mod
SeanLedford: Hmm I get the same output that you did. Yesterday I tried discarding my go.mod changes and was getting some PR check failures, doing so again so we can see the results

e2e/cli_test.go
```diff
@@ -504,29 +507,29 @@ allow if {
 `,
 		filepath.FromSlash("wow/foo-bar/baz/main_test.rego"): `package wow["foo-bar"].baz_test
 
-test_allow := true
+test_allow if {}
```
charlieegan3, line 511: Are these changes needed? I can see why we need the `allow if { input == 1 }` edits above, but these don't look quite correct to me. since `test_allow := true` is not a constant condition violation. 🤔

pkg/fixer/fixes/constantcondition_test.go
```diff
@@ -0,0 +1,183 @@
...
+
+			if !tc.fixExpected && len(fixResults) != 0 {
+				t.Fatalf("unexpected fix applied")
+			}
+
+			if !tc.fixExpected {
+				return
+			}
+
+			if diff := cmp.Diff(fixResults[0].Contents, tc.contentAfterFix); tc.fixExpected && diff != "" {
```
charlieegan3, line 174: `diff` here, if not empty, contains a nice string to print out in the error case. Let's use that to save manually crafting the error.

pkg/fixer/fixer_test.go
```diff
@@ -51,20 +53,25 @@ deny = true
...
+			"no-whitespace-comment",
+			"opa-fmt",
+			"constant-condition",
+		},
 	}
 	expectedFileContents := map[string]string{
 		filepath.Join(rootPath, "test", "main.rego"): `package test
 
-allow := true
+allow if {}
```
charlieegan3, line 66: Similar to the comment above, I am unsure why this is needed.

pkg/fixer/fixes/redundantexistencecheck_test.go
```diff
@@ -0,0 +1,170 @@
...
+				Locations: []report.Location{
+					{
+						Row: 4, Column: 5, End: &report.Position{
+							Row: 4, Column: 21,
+						},
+					},
+				},
+			},
+		},
+		"bad change": {
```
charlieegan3, line 72: It's not immediately clear to me what is bad about this one

pkg/fixer/fixes/redundantexistencecheck_test.go
```diff
@@ -0,0 +1,170 @@
...
+    input.user.email
+    endswith(input.user.email, "@acmecorp.com")
+}
+
+is_admin(user) if {
+    user
+    "admin" in user.roles
+}`,
+			},
+			contentAfterFix: `package test
```
charlieegan3: I'd reorder some fields here, fc first, then runtime options, then fix expected, then context after fix. I think that might help the reading flow a bit.

pkg/fixer/fixes/redundantexistencecheck_test.go
```diff
@@ -0,0 +1,170 @@
...
+				t.Fatalf("unexpected fix applied")
+			}
+
+			if !tc.fixExpected {
+				return
+			}
+
+			if diff := cmp.Diff(fixResults[0].Contents, tc.contentAfterFix); tc.fixExpected && diff != "" {
+				t.Fatalf(
+					"unexpected content, got:\n%s---\nexpected:\n%s---",
```
charlieegan3: Same comment as other case of `diff` having a value.

pkg/fixer/fixes/redundantexistencecheck_test.go
SeanLedford: Good call I agree

pkg/fixer/fixes/redundantexistencecheck_test.go
SeanLedford: Yeah it gets passed a bad location here (out of range), this is really just testing to make sure the test isn't panicking or making an errant update

e2e/cli_test.go
SeanLedford: Yeah this is needed because this is the new expected result in this case. Originally for a rule like:

test_allow if {
    true
}

The formatter would shorten this to one line: `test_allow := true`. But now since this fix removes constant conditions, the formatter is just linting the newlines in the empty object

pkg/fixer/fixer_test.go
SeanLedford: Yep replied to the one above

e2e/cli_test.go
charlieegan3: Hmm, but `test_allow := true` is not a constant violation right? I am unsure where `test_allow := true` is causing issues since I hoped it'd not be relevant for to the changes here./

pkg/fixer/fixes/redundantexistencecheck_test.go
charlieegan3: Ahh thanks, I might have the name of the test say that the location is the bad part just to make this easier to read, but having the fc, and runtime options listed first helps too.

SeanLedford: Oh yeah I agree that's pretty unclear, I'll update this

charlieegan3 (approved):
Great work here Sean! thanks for all the back and forth on the tests.

PR #1797, opened 2025-12-17
docs: Update messy rule docs

charlieegan3:
We had the following feedback

```
Feedback Type
  Negative
Page URL
  https://www.openpolicyagent.org/projects/regal/rules/style/messy-rule
Comment
  I couldn't understand this quickly.
  Please explain in more detail and with more examples.
```

I hope this makes this page a little more accessible.


anderseknert (approved):
Thanks!

PR #1798, opened 2025-12-17
version: Add OPA version to version output

charlieegan3:
Fixes https://github.com/open-policy-agent/regal/issues/1785

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -4,6 +4,8 @@ import (
 	"cmp"
 	"runtime"
 	"strings"
+
+	opaversion "github.com/open-policy-agent/opa/v1/version"
 )
 
 const platform = runtime.GOOS + "/" + runtime.GOARCH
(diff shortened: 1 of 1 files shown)


pkg/version/version.go
```diff
@@ -21,35 +23,38 @@ var (
...
+			"Version:      " + vi.Version,
+			"Go Version:   " + vi.GoVersion,
+			"OPA Version:  " + vi.OPAVersion,
+			"Platform:     " + vi.Platform,
+			"Commit:       " + vi.Commit,
+			"Timestamp:    " + vi.Timestamp,
+			"Installed via: " + vi.Hostname,
 		},
 		"\n",
 	) + "\n"
```
srenatus, line 47: Heh I guess we could also add an empty string to the `[]string{}` above. Doesn't matter.

pkg/version/version.go
```diff
@@ -21,35 +23,38 @@ var (
...
-	Commit    string `json:"commit"`
-	Timestamp string `json:"timestamp"`
-	Hostname  string `json:"hostname"`
+	Version    string `json:"version"`
+	GoVersion  string `json:"go_version"`
+	OPAVersion string `json:"opa_version"`
+	Platform   string `json:"platform"`
+	Commit     string `json:"commit"`
+	Timestamp  string `json:"timestamp"`
+	Hostname   string `json:"hostname"`
```
srenatus, line 32: Please can we remove hostname? 😵‍💫

charlieegan3: Ahh, I took it to be a user/human change only, leaving Hostname as the build flag

https://github.com/Homebrew/homebrew-core/blob/main/Formula/r/regal.rb#L26 would also need to be updated if we changed that.

srenatus: Well we could just ignore the value if set. But I also believe if it's a ref into nowhere, go build won't fail. But yeah this can wait.

pkg/version/version.go
charlieegan3: Hmm, not sure what `[]string{} above` refers to... 🤔 (specifically what needs an empty string?)

srenatus: Sorry it's just nitpicking. I thought we'd get the same thing as

```go
return strings.Join([]string{"Version:  "+vi.Verison, /* ... */}, "\n")+"\n"
```
via this trick
```go
return strings.Join([]string{"Version:  "+vi.Verison, /* ... *}, "", "\n")
```

charlieegan3: Ohh! 🤦  got it at last!

Fixed!

PR #1810, opened 2026-01-06
Code Action for prefer-equals-comparison fixer

SeanLedford:
Implemented first code action fix to supplement the recently added fixes. Marked as draft until coverage requirements and AC have been confirmed.

### DIFF
--- a/bundle/regal/lsp/codeaction/codeaction.rego
+++ b/bundle/regal/lsp/codeaction/codeaction.rego
@@ -129,6 +129,7 @@ rules := {
 		"Move file so that directory structure mirrors package path",
 		["target", "diagnostic"],
 	],
+	"prefer-equals-comparison": ["Replace = with == in comparison", ["target", "diagnostic"]],
 }
 
 # METADATA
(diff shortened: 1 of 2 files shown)


bundle/regal/lsp/codeaction/codeaction.rego
```diff
@@ -129,6 +129,7 @@ rules := {
 		"Move file so that directory structure mirrors package path",
 		["target", "diagnostic"],
 	],
+	"prefer-equals-comparison": ["Replace `=` with `==` in comparison", ["target", "diagnostic"]],
```
charlieegan3: Not 100% sure, but I don't think these are markdown-able?

```suggestion
	"prefer-equals-comparison": ["Replace = with == in comparison", ["target", "diagnostic"]],
```

SeanLedford: Yeah addressed, I believe you're right and we don't use backticks anywhere else in these descriptions it looks like.

PR #1811, opened 2026-01-07
OPA v1.12.2 and full string interpolation support

anderseknert:
- Bump our capabilities.json to include new feature
- And the deprecated attribute on builtins
- Remove related workaround that was previously required
- Update docs on RoAST to cover the `interpolated` attribute added to template string expressions
- Update constant-condition rule to include expressions that only contain a template string
- Make `ast.ref_static_to_string` and `ast.static_ref` aware of template strings, and that they are not to be counted as static.
- Make `ast.has_named_var` take template strings into account
- Add various tests to ensure template strings are considered by helper rules and functions in the `ast` package
- Replace `sprintf` and `concat` calls all over the code base. Extremely satisfying! Although this also uncovered a few issues in OPA which required a v1.12.2 release
- RoAST serialization of template strings

### DIFF
--- a/bundle/regal/lsp/completion/providers/snippet/snippet_test.rego
+++ b/bundle/regal/lsp/completion/providers/snippet/snippet_test.rego
@@ -185,4 +185,8 @@ test_metadata_snippet_completion if {
 	}
 }
 
-_with_header(policy) := concat("\n\n", ["package policy", "import rego.v1", policy])
+_with_header(policy) := $`package policy
+
+import rego.v1
+
+{policy}`
--- a/bundle/regal/rules/imports/circular-import/circular_import.rego
+++ b/bundle/regal/rules/imports/circular-import/circular_import.rego
@@ -50,13 +50,10 @@ aggregate_report contains violation if {
 		some loc in _package_locations[m1][m2]
 	][0]
 
-	violation := result.fail(
-		rego.metadata.chain(),
-		{
-			"description": sprintf("Circular import detected in: %s", [concat(", ", sort(g))]),
-			"location": location,
-		},
-	)
+	violation := result.fail(rego.metadata.chain(), {
+		"description": $`Circular import detected in: {concat(", ", sorted_group)}`,
+		"location": location,
+	})
 }
 
 # METADATA
@@ -70,13 +67,10 @@ aggregate_report contains violation if {
 
 	some pkg in g # this will the only package
 
-	violation := result.fail(
-		rego.metadata.chain(),
-		{
-			"description": sprintf("Circular self-dependency in: %s", [pkg]),
-			"location": [e | some e in _package_locations[pkg][pkg]][0],
-		},
-	)
+	violation := result.fail(rego.metadata.chain(), {
+		"description": $`Circular self-dependency in: {pkg}`,
+		"location": [e | some e in _package_locations[pkg][pkg]][0],
+	})
 }
 
 # METADATA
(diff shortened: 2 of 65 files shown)


bundle/regal/lsp/completion/providers/snippet/snippet_test.rego
```diff
@@ -185,4 +185,4 @@ test_metadata_snippet_completion if {
 	}
 }
 
-_with_header(policy) := concat("\n\n", ["package policy", "import rego.v1", policy])
+_with_header(policy) := $"package policy\n\nimport rego.v1\n\n{policy}"
```
srenatus: We could use multi-line here? 🤔 
```suggestion
_with_header(policy) := $`package policy

import rego.v1

{policy}`
```

Hmm not sure it's worth it.

bundle/regal/rules/imports/circular-import/circular_import.rego
```diff
@@ -50,13 +50,10 @@ aggregate_report contains violation if {
...
 
-	violation := result.fail(
-		rego.metadata.chain(),
-		{
-			"description": sprintf("Circular import detected in: %s", [concat(", ", sort(g))]),
-			"location": location,
-		},
-	)
+	violation := result.fail(rego.metadata.chain(), {
+		"description": $`Circular import detected in: {concat(", ", sorted_group)}`,
```
srenatus, line 54: 💭 Do we _need_ the `? Would this not work, too?
```suggestion
		"description": $"Circular import detected in: {concat(", ", sorted_group)}",
```

internal/explorer/stages.go
```diff
@@ -34,6 +34,7 @@ var stages = []stage{
 	{"SetModuleTree", "compile_stage_set_module_tree"},
 	{"SetRuleTree", "compile_stage_set_rule_tree"},
 	{"RewriteLocalVars", "compile_stage_rewrite_local_vars"},
+	{"RewriteTemplateStrings", "compile_stage_rewrite_template_strings"},
```
srenatus, line 37: :+1:

bundle/regal/lsp/completion/providers/snippet/snippet_test.rego
anderseknert: Fixed

bundle/regal/rules/imports/circular-import/circular_import.rego
anderseknert: It does. I think I'm just not used to having strings-in-strings just yet.

anderseknert:
@srenatus there's an impact for sure, but given recent perf improvements in both OPA and Regal, we have some headroom to afford this. I have a bunch of performance improvements related to interpolation lined up for  the next release, and if they work out (and it seems like they do) we'll have paid off the cost and more with quite some margin even. Stay tuned.

internal/roast/encoding/templatestring_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"testing"
+
+	jsoniter "github.com/json-iterator/go"
+
+	"github.com/open-policy-agent/opa/v1/ast"
+)
+
+func TestTemplateStringMarshalling(t *testing.T) {
```
charlieegan3, line 11: It'd be nice to have a nested case here e.g.

```
foo := $`foo {$`bar {1+1}`}`
```

charlieegan3 (commented):
LGTM

internal/roast/encoding/templatestring_test.go
anderseknert: Yeah, good call. I'll add one in the next PR.

srenatus:
> there's an impact for sure, but given recent perf improvements in both OPA and Regal, we have some headroom to afford this.

I wasn't certain that it would be a negative impact, tbh. You think it'll be worse, then, compared to the `sprintf` approach, before applying further optimizations? I haven't measured anything, and I don't have an intuition re: multiple set comprehensions vs handling stuff in `fmt.Sprinf`.

PR #1812, opened 2026-01-07
Move `chained-rule-body` to `custom` category

anderseknert:
As this is effectively covered by the `opa-fmt` rule already, there's little point in having this enabled by default. For that reason moving this to the `custom` category.


anderseknert:
@charlieegan3 I don't think this rule was ever that important, but let me know if there's anything we should do with regards to the doc site. I'm thinking perhaps a redirect from style to custom. I don't think we really need it though.

PR #1815, opened 2026-01-08
bundle: Improve performance of any_set_item

charlieegan3:
Turns out that sort is fastest for this function so we are using that  now!


anderseknert (approved):
Nice with the table!

PR #1817, opened 2026-01-09
Add `prefer-value-in-head` option to count interpolated string as scalar

anderseknert:
I noticed in some rule I worked on that Regal didn't flag this, which was right given that interpolated strings aren't exactly scalar values. But I _do_ want to have Regal recommend moving them to the head assignment when possible, while keeping the `only-scalars` option set as `true`. This PR adds a new `include-interpolated` option to allow this precisely, and makes it enabled in our own settings.

### DIFF
--- a/docs/rules/custom/prefer-value-in-head.md
+++ b/docs/rules/custom/prefer-value-in-head.md
@@ -5,6 +5,7 @@
 **Category**: Custom
 
 **Avoid**
+
 ```rego
 package policy
 
@@ -20,6 +21,7 @@ deny contains message if {
 ```
 
 **Prefer**
+
 ```rego
 package policy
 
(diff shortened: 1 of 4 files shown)


docs/rules/custom/prefer-value-in-head.md
```diff
@@ -74,6 +77,9 @@ rules:
       # whether to only suggest moving scalar values (strings, numbers, booleans, null)
       # to the head, and not expressions or functions
       only-scalars: false
+      # when set to true, counts interpolated strings as a scalar value, and will suggest
+        # moving them to the head even when `only-scalars` is true
```
charlieegan3:
```suggestion
      # moving them to the head even when `only-scalars` is true 
```

PR #1827, opened 2026-01-13
bundle: Surface configured rule notices

charlieegan3:
If a rule is configred, and set to non-ignore level, then we should show the notices from that rule if it is not run to aid users.

Fixes https://github.com/open-policy-agent/regal/issues/1795


anderseknert (approved):
LGTM! Did you check to see if this had any noticeable impact on the benchmarks?

charlieegan3:
The numbers appear to be pretty comparable on consecutive runs 👍

PR #1831, opened 2026-01-15
Implement completionItem/resolve handler

anderseknert:
I was looking to fix some robustness issues in our code completion feature, but I still have more to do there. However, while logging completion requests and responses I noticed that a non-negligible amount of the data transfered was from the built-in provider and the documentation for each built-in it sends along with every built-in suggestion. These markdown documents are only visible when the user clicks to see them, which is realistically not very common (but nice when you need it). So I figured I'd make a little detour to implement our first [resolve provider](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion). These allow resolving parts of some responses on demand rather than upfront, thus saving procesing time and bytes in transfer. Not a huge win, mind you, but it's something.. and paves the way for more resolve providers in the future, as they are now rather simple to implement using only Rego.

Also:
- Add HasFileContents for easier debugging as GetFileContents
- Use VersionedTextIdentifier as per language server spec, although we don't currently use the version number for anything (but possibly logging)

### DIFF
--- a/bundle/regal/lsp/completion/main.rego
+++ b/bundle/regal/lsp/completion/main.rego
@@ -18,6 +18,18 @@ import data.regal.util
 result["response"] := {
 	"items": items,
 	"isIncomplete": true,
+} if {
+	input.method == "textDocument/completion"
+}
+
+# METADATA
+# schemas:
+#   - input: {}
+result["response"] := data.regal.lsp.completion.resolvers[input.params.data.resolver].resolve if {
+	input.method == "completionItem/resolve"
+} else := input.params if {
+	# if there was nothing to resolve, return the input as-is
+	input.method == "completionItem/resolve"
 }
 
 # METADATA
--- a/internal/lsp/rego/router.go
+++ b/internal/lsp/rego/router.go
@@ -34,7 +34,7 @@ var (
 
 type (
 	Providers struct {
-		ContextProvider              func(uri string, reqs *Requirements) RegalContext
+		ContextProvider              func(uri string, reqs *Requirements) *RegalContext
 		ContentProvider              func(uri string) (string, bool)
 		IgnoredProvider              func(uri string) bool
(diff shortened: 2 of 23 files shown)


bundle/regal/lsp/completion/main.rego
```diff
@@ -18,6 +18,18 @@ import data.regal.util
...
+}
+
+# METADATA
+# schemas:
+#   - input: {}
+result["response"] := data.regal.lsp.completion.resolvers[input.params.data.resolver].resolve if {
+	input.method == "completionItem/resolve"
+} else := input.params if {
+	# if there was nothing to resolve, return the input as-is
+	input.method == "completionItem/resolve"
```
charlieegan3, line 32: It might be nice to have some default if there is one called that does not exist so we at least get an error message in such cases, even if the client behaviour is the same.

internal/lsp/rego/router.go
```diff
@@ -114,6 +118,12 @@ func (m *RegoRouter) Handle(ctx context.Context, _ *jsonrpc2.Conn, req *jsonrpc2
 			return nil, fmt.Errorf("no prepared query for %s", query.MainEval)
 		}
 
+		if strings.HasSuffix(req.Method, "/resolve") && route.resolver != nil {
```
charlieegan3, line 121: For now, does it make sense to log a if we get something other than completion resolve so we know that something unexpected is being served?

charlieegan3 (approved):
Looks like a nice addition to me.

internal/lsp/rego/router.go
anderseknert: We'd get an error logged about that, as there would be no matching handler. While it wouldn't be detailed I think that's enough for the time being, as the server configuration specifically says only completion items can be resolved. And if the client does not respect that, most of our bets are probably off anyway..

charlieegan3: ok sounds good!

PR #1838, opened 2026-01-19
25% ns/op reduction by aggregate remodel + Rego prepare stage

### LINKED ISSUE #359: Create new Rego setup phase preceeding lint evalutation

### LINKED ISSUE #1265: Remodel rule aggregation format

anderseknert:
This concludes work I started before the holidays, and which I have worked on when given the time. It's a lot! Sorry about that, but it'll be worth it, I'm sure :)

- Remodel aggregate collection to keep AST shape without roundtrips
- Remove all metadata on each aggregate item for performance
- Introduce `ogre` package for lightweight `rego` alternative
- Introduce new prepare stage done in Rego, where we set up things once and persist to store instead of in every file linted

Fixes #359
Fixes #1265

And probably a few more, lol


internal/ogre/evaluator.go
```diff
@@ -0,0 +1,118 @@
...
+	e.txn = txn
+
+	return e
+}
+
+func (e *Evaluator) Profiler() *profiler.Profiler {
+	return e.profiler
+}
+
+func (e *Evaluator) WithResultHandler(f func(result ast.Value) error) *Evaluator {
```
srenatus: [nit]
```suggestion
func (e *Evaluator) WithResultHandler(f func(st.Value) error) *Evaluator {
```

pkg/linter/linter_test.go
```diff
@@ -563,6 +563,15 @@ func TestEnabledAggregateRules(t *testing.T) {
 	}
 }
 
+func ref(s ...string) ast.Ref {
+	terms := make([]*ast.Term, len(s))
+	for i, part := range s {
+		terms[i] = ast.StringTerm(part)
```
srenatus: No InternedSomething? 💭

internal/ogre/evaluator.go
anderseknert: There's a typo in your nit 😄 But I'll update the code! You'd think there would be linters to cover this completely in 2026.

pkg/linter/linter_test.go
anderseknert: Funny, I don't even remember having written this code. Must be before the time of the many julbord. Yes, I'll update.

bundle/regal/prepared/prepared.rego
```diff
@@ -0,0 +1,54 @@
...
+
+notices := data.internal.prepared.notices if not file_notices
+
+notices := ext_notices.promoted_notices if file_notices
+
+# METADATA
+# description: determine if per-file notices should be processed or fetched from prepared state
+# scope: document
+file_notices if input.regal.file.rego_version != "v1"
+file_notices if count(config.capabilities.special.no_filename) > 0
```
charlieegan3, line 54: A little more of a comment about these two cases might help in future. Even just a reminder that no_filename is stdin, and therefore we can't lookup.

bundle/regal/rules/bugs/impossible-not/impossible_not.rego
```diff
@@ -56,20 +56,20 @@ _negated_refs contains negated_ref if {
 
 # METADATA
 # description: collects imported symbols, multi-value rules and negated refs
-aggregate contains result.aggregate(rego.metadata.chain(), {
+aggregate contains {
```
charlieegan3, line 59: Does this change need to be noted for custom rule authors in the custom rule docs too?

charlieegan3: (the removal of rego.metadata.chain etc)

charlieegan3:
Nice work here, ogre is a nice abstraction for the project!

bundle/regal/rules/bugs/impossible-not/impossible_not.rego
anderseknert: See the comment above _aggregate_report_inputs in main.rego. We reconstruct the old format for custom aggregate rules, so they should be unaffected at this point. They also won't benefit from this though! So eventually we'll want to allow them to switch over and deprecate/remove the legacy structure.

anderseknert:
Running the tests repeatedly on my laptop and I'm not able to reproduce any of the issues that fail the CI build. I'll look some more into it. They're all related to the language server, and I have mostly focused on the linter here. Could be that there's some more work needed to have them play nicely together. But it's hard when you can't reproduce the issue locally :(

anderseknert:
Oh, I had forgotten to revert the change of the default timeout for the LSP tests from 20 seconds to 2. That could explain a thing or two... :) 

I have found it quite useful to use a much lower value for local development, as issues tend to surface much quicker that way. So I'll see about having something that allows keeping 20 seconds in CI while setting it lower locally.

PR #1845, opened 2026-01-27
Semantic Token Request Framework

SeanLedford:
Implementation for semantic token requests using some starter use cases, namely package names and function variable declarations/references. Needs feedback on the parsing of the AST,  and a test written once that's resolved.

### DIFF
--- a/internal/lsp/semantictokens/semantictokens.go
+++ b/internal/lsp/semantictokens/semantictokens.go
@@ -3,14 +3,15 @@ package semantictokens
 import (
 	"context"
 	"fmt"
+	"slices"
 
-	"sort"
 	"strconv"
 	"strings"
 
 	"github.com/open-policy-agent/opa/v1/ast"
 	"github.com/open-policy-agent/regal/internal/lsp/types"
 	"github.com/open-policy-agent/regal/internal/ogre"
+	"github.com/open-policy-agent/regal/pkg/roast/rast"
 	"github.com/open-policy-agent/regal/pkg/roast/transform"
 )
 
@@ -32,7 +33,7 @@ type Token struct {
 	Modifiers uint
 }
 
-func Full(module *ast.Module) (*types.SemanticTokens, error) {
+func Full(ctx context.Context, module *ast.Module) (*types.SemanticTokens, error) {
 	tokens := make([]Token, 0)
 
 	packageTokens, err := extractPackageTokens(module)
@@ -41,7 +42,7 @@ func Full(module *ast.Module) (*types.SemanticTokens, error) {
 	}
 	tokens = append(tokens, packageTokens...)
 
-	variableTokens, err := extractVariableTokens(module)
+	variableTokens, err := extractVariableTokens(ctx, module)
 	if err != nil {
 		return nil, fmt.Errorf("failed to extract variable tokens: %w", err)
 	}
@@ -54,11 +55,8 @@ func extractPackageTokens(module *ast.Module) ([]Token, error) {
 	var tokens []Token
 
 	if module.Package != nil && module.Package.Path != nil {
-		for _, term := range module.Package.Path {
+		for _, term := range module.Package.Path[1:] {
 			packageString := term.Value.String()
-			if packageString == "data" {
-				continue
-			}
 
 			trimmedValue := strings.Trim(packageString, `"`)
 			length := uint(len(trimmedValue))
@@ -76,16 +74,17 @@ func extractPackageTokens(module *ast.Module) ([]Token, error) {
 	return tokens, nil
 }
 
-func extractVariableTokens(module *ast.Module) ([]Token, error) {
-	ctx := context.Background()
+var (
+	semanticTokensQuery = ast.MustParseBody(`arg_tokens = data.regal.lsp.semantictokens.arg_tokens`)
+)
+
+func extractVariableTokens(ctx context.Context, module *ast.Module) ([]Token, error) {
 	var tokens []Token
 
 	roastInput, err := transform.ToAST("policy.rego", "", module, true)
 	if err != nil {
 		return nil, fmt.Errorf("failed to transform to roast format: %w", err)
 	}
-
-	query := ast.MustParseBody(`arg_tokens = data.regal.lsp.semantictokens.arg_tokens`)
 
 	resultHandler := func(result ast.Value) error {
 		resultObj := result.(ast.Object)
@@ -94,14 +93,16 @@ func extractVariableTokens(module *ast.Module) ([]Token, error) {
 			varObj := varTerm.Value.(ast.Object)
 			typeStr := string(typeTerm.Value.(ast.String))
 
-			locationStr := string(varObj.Get(ast.StringTerm("location")).Value.(ast.String))
-			varName := string(varObj.Get(ast.StringTerm("value")).Value.(ast.String))
+			locationStr := rast.GetString(varObj, "location")
+			varName := rast.GetString(varObj, "value")
 
 			trimmedVarName := strings.Trim(varName, `"' `)
 
-			parts := strings.Split(locationStr, ":")
-			row, _ := strconv.Atoi(parts[0])
-			col, _ := strconv.Atoi(parts[1])
+			rowStr, rest, _ := strings.Cut(locationStr, ":")
+			colStr, _, _ := strings.Cut(rest, ":")
+
+			row, _ := strconv.Atoi(rowStr)
+			col, _ := strconv.Atoi(colStr)
 
 			modifier := ModifierReference
 			if typeStr == "declaration" {
@@ -121,7 +122,7 @@ func extractVariableTokens(module *ast.Module) ([]Token, error) {
 		return nil
 	}
 
-	q, err := ogre.New(query).Prepare(ctx)
+	q, err := ogre.New(semanticTokensQuery).Prepare(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("failed to run ogre query: %w", err)
 	}
@@ -144,11 +145,11 @@ func encodeTokens(tokens []Token) *types.SemanticTokens {
 	}
 
 	// Sort tokens by position (line first, then column)
-	sort.Slice(tokens, func(i, j int) bool {
-		if tokens[i].Line != tokens[j].Line {
-			return tokens[i].Line < tokens[j].Line
+	slices.SortFunc(tokens, func(a, b Token) int {
+		if a.Line != b.Line {
+			return int(a.Line) - int(b.Line)
 		}
-		return tokens[i].Col < tokens[j].Col
+		return int(a.Col) - int(b.Col)
 	})
 
 	data := make([]uint, 0, len(tokens)*5)
@@ -177,5 +178,4 @@ func encodeTokens(tokens []Token) *types.SemanticTokens {
 	}
 
 	return &types.SemanticTokens{Data: data}
-
 }
--- a/internal/lsp/semantictokens/semantictokens_test.go
+++ b/internal/lsp/semantictokens/semantictokens_test.go
@@ -79,7 +79,7 @@ test_function(param1) := result if {
 
 			module := ast.MustParseModule(tc.policy)
 
-			result, err := Full(module)
+			result, err := Full(t.Context(), module)
 			if err != nil {
 				t.Fatalf("unexpected error: %v", err)
 			}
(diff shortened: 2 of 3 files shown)


internal/lsp/semantictokens/semantictokens.go
```diff
@@ -0,0 +1,150 @@
...
+		}
+	}
+
+	// My thought process here was:
+	// 1. mark all variables found in rule head as declared via map
+	// 2. variables found in body are then references
+	//
+	// This is a faulty assumption though, as variables could be redeclared
+	// and not necessarily be references. Could use some feedback on how to account
+	// for this, I wasn't certain after looking at what the AST provides
```
SeanLedford: Relevant comment to take a look at.

charlieegan3: Interesting, I do wonder if there is a need here to reuse some of the rego code we have for finding things in the AST already. E.g. https://github.com/open-policy-agent/regal/blob/main/bundle/regal/rules/bugs/var-shadows-builtin/var_shadows_builtin.rego#L9C9-L9C32 is one way to get a list of variables.

anderseknert: This is indeed tricky to do! If you want prior art for how to do this in Go, your best bet is to study the [compiler](https://github.com/open-policy-agent/opa/blob/main/v1/ast/compile.go), as that does this kind of work in many of its stages. Just like in Regal, it involves a lot of `walk`s across the AST. It's not a 100% perfect comparison though, as he compiler stages also _rewrite_ the AST. This all happens at a later stage than the AST processing we do in Regal. If you are curious to see the results of what the compiler does in each stage, we have a great tool for that @srenatus wrote, predictably called the compiler explorer. You can access that from language server clients like VS code by right clicking inside of a policy and choose "Source Action..." then "Explore compiler stages for this policy". Step by step, you'll see the declarative language lie unfold before your eyes, as your Rego is rewritten to pure unification :P 

Anyway! Back to the topic here, I previously suggested starting with function arguments since they are at least somewhat simpler to reason about in this context. Why? Because the compiler does not allow redeclaring variables set in arguments in neither the head or the body. The below will therefore not compile:

```rego
fun(a, b) := [a, b] if {
	a := 1
    b := 2
}
```

A simple implementation could thus simply treat variables defined in args as _variable declaration_ tokens, and their use anywhere in the head or body as _variable references_. Now of course, nothing in life is really that simple, and even if arguably bad form, OPA _does_ allow reusing the bindings in child scopes, like comprehensions, `every` statements and so on. So this would be valid:

```rego
fun(a, b) := [a, b] if {
    comp := {a: b | 
    	a := 1
    	b := 2
    }
}
```

But I think a first implementation can ignore that fact, and 

> treat variables defined in args as _variable declaration_ tokens, and their use anywhere in the head or body as _variable references_

even if there are some cases where this isn't true. This is is just for highlighting after all, and we can improve things like that later.

But also totally possible to do this in Rego, and I think we have plenty of helpers to assist with that in the `ast` package today. Try things out and see which you prefer.

internal/lsp/semantictokens/semantictokens.go
```diff
@@ -0,0 +1,204 @@
...
+
+	tokens := make([]Token, 0)
+
+	if module.Package != nil && module.Package.Path != nil {
+		for _, term := range module.Package.Path {
+			if term.Value.String() == "data" {
+				continue
+			}
+
+			trimmedValue := strings.Trim(term.Value.String(), `"`)
```
anderseknert: `term.Value.String()` was called above already, so store that and use here

internal/lsp/semantictokens/semantictokens.go
```diff
@@ -0,0 +1,204 @@
...
+			some expr in expressions
+			some term in expr.terms
+			term.type == "var"
+			
+			arg_names := {v.value | some v in ast.found.vars[rule_index].args}
+			term.value in arg_names
+			
+			var := term
+		}
+	`)
```
anderseknert: The policy above looks OK to me, but it'd be better placed somewhere under `bundle/regal/lsp` as that'll make it tested and linted along with all other policies. And you won't need to parse it here per request.

internal/lsp/semantictokens/semantictokens.go
```diff
@@ -0,0 +1,204 @@
...
+		return nil, fmt.Errorf("Failed to parse module here: %w", err)
+	}
+
+	query := ast.MustParseBody(`arg_tokens = data.semantictokens.arg_tokens`)
+
+	resultHandler := func(result ast.Value) error {
+		// Process variable tokens (args) from rego
+		resultObj := result.(ast.Object)
+
+		resultObj.Iter(func(varTerm, typeTerm *ast.Term) error {
```
anderseknert: You can use resultObj.Foreach here instead, as there's no error returned in the callback

internal/lsp/semantictokens/semantictokens.go
```diff
@@ -0,0 +1,204 @@
...
+	Length    uint
+	Type      uint
+	Modifiers uint
+}
+
+func Full(module *ast.Module) (*types.SemanticTokens, error) {
+
+	tokens := make([]Token, 0)
+
+	if module.Package != nil && module.Package.Path != nil {
```
charlieegan3, line 57: The Full function is quite long now, let's break it down into some smaller functions that get the different types of tokens we support. that will also help reduce the nesting here. Each function's result, if present, can be appended to the list of tokens we return in the end.

internal/lsp/semantictokens/semantictokens.go
```diff
@@ -0,0 +1,204 @@
...
+				Type:      TokenTypePackage,
+				Modifiers: 0,
+			})
+
+		}
+	}
+
+	ctx := context.Background()
+
+	roastInput, err := transform.ToAST("semantictokens.rego", "", module, true)
```
charlieegan3:
```suggestion
	roastInput, err := transform.ToAST("policy.rego", "", module, true)
```

I think in such cases we tend to use policy.rego as the more generic filename for 'some rego file'.

internal/lsp/semantictokens/semantictokens.go
```diff
@@ -0,0 +1,204 @@
...
+	return result, nil
+}
+
+func encodeTokens(tokens []Token) *types.SemanticTokens {
+	if len(tokens) == 0 {
+		return &types.SemanticTokens{Data: []uint{}}
+	}
+
+	// Sort tokens by position (line first, then column)
+	sort.Slice(tokens, func(i, j int) bool {
```
charlieegan3: Is the sort needed for tests only? I am just interested as I would hope that the client doesn't really mind what order test are in? if for tests only, we could sort the results in the test rather than here in the implementation.

internal/lsp/semantictokens/semantictokens.go
```diff
@@ -0,0 +1,204 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Failed to evaluate ogre query: %w", err)
+	}
+
+	result := encodeTokens(tokens)
+
+	return result, nil
+}
+
+func encodeTokens(tokens []Token) *types.SemanticTokens {
```
charlieegan3, line 142: This function's return type has to be pretty opaque since that's what's in the LSP api. However, we might be able to have some test helper that takes a uint slice and makes some nice structure that we can use to make the test comparisons easier to make.

internal/lsp/semantictokens/semantictokens.go
charlieegan3: E.g. 

in the test file we might have a type for this


```go
data = append(data,
			deltaLine,
			deltaCol,
			tok.Length,
			tok.Type,
			tok.Modifiers,
		)
```

with fields for line, col, Len, type, modifiers

and then in tests, we can check each part and make the errors easier to read.

internal/lsp/semantictokens/semantictokens.go
SeanLedford: Addressed!

internal/lsp/semantictokens/semantictokens.go
SeanLedford: Gotcha, moved the policy under the bundle and added a test. When you mention parsing here, are you talking about the policy being evaluated? That would still need to happen per request to keep the token info up to date, no?

internal/lsp/semantictokens/semantictokens.go
SeanLedford: Addressed!

internal/lsp/semantictokens/semantictokens.go
SeanLedford: The sorting is done purely for the implementation itself! I imagined there would be cases where the ast would return tokens to be parsed that aren't in the same order they come in the file. Which means we'd need to sort the tokens, because what we're actually passing back in the response is the relative distance between each token, with the first in each column being the one that the distances are based off of. It's a bit difficult to explain succinctly here 😅, I found this link to be super helpful understanding it visually:

https://pygls.readthedocs.io/en/latest/protocol/howto/interpret-semantic-tokens.html

internal/lsp/semantictokens/semantictokens.go
SeanLedford: Yeah the fields of the return type would definitely be hard to interpret in the future, added some clarity to the tests comparisons and error messages.

internal/lsp/semantictokens/semantictokens.go
SeanLedford: Good idea, split the function into two for starters.

bundle/regal/lsp/semantictokens/semantictokens.rego
```diff
@@ -0,0 +1,42 @@
+# METADATA
+# description: |
+#   Returns location of variables to be highlighted via semantic tokens. Currently returns:
+#     - declarations of function args in text documents
+#     - variable references that are used in function calls
+#     - variable references that are used in expressions
+# related_resources:
+#   - https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_semanticTokens
+package regal.lsp.semantictokens
```
anderseknert, line 9: Clean Rego in this package, and well-documented!

internal/lsp/semantictokens/semantictokens.go
```diff
@@ -0,0 +1,181 @@
...
+
+func extractPackageTokens(module *ast.Module) ([]Token, error) {
+	var tokens []Token
+
+	if module.Package != nil && module.Package.Path != nil {
+		for _, term := range module.Package.Path {
+			packageString := term.Value.String()
+			if packageString == "data" {
+				continue
+			}
```
anderseknert: `data` will always be the first element of the path, so you could instead do:

```go
for _, term := range module.Package.Path[1:] {
```

and skip this check.

internal/lsp/semantictokens/semantictokens.go
```diff
@@ -0,0 +1,181 @@
...
+				Modifiers: 0,
+			})
+		}
+	}
+
+	return tokens, nil
+}
+
+func extractVariableTokens(module *ast.Module) ([]Token, error) {
+	ctx := context.Background()
```
anderseknert: We should pass the context as the first arg to `Full` and onwards here, so that these functions will be cancelable, etc.

internal/lsp/semantictokens/semantictokens.go
```diff
@@ -0,0 +1,181 @@
...
+	var tokens []Token
+
+	roastInput, err := transform.ToAST("policy.rego", "", module, true)
+	if err != nil {
+		return nil, fmt.Errorf("failed to transform to roast format: %w", err)
+	}
+
+	query := ast.MustParseBody(`arg_tokens = data.regal.lsp.semantictokens.arg_tokens`)
```
anderseknert: Move this out to a package level `var` so we only parse this once.

internal/lsp/semantictokens/semantictokens.go
```diff
@@ -0,0 +1,181 @@
...
+
+	resultHandler := func(result ast.Value) error {
+		resultObj := result.(ast.Object)
+
+		resultObj.Foreach(func(varTerm, typeTerm *ast.Term) {
+			varObj := varTerm.Value.(ast.Object)
+			typeStr := string(typeTerm.Value.(ast.String))
+
+			locationStr := string(varObj.Get(ast.StringTerm("location")).Value.(ast.String))
+			varName := string(varObj.Get(ast.StringTerm("value")).Value.(ast.String))
```
anderseknert: This is OK, but check out the `rast` package for some helpers you could use here to make this more compact/readable.

internal/lsp/semantictokens/semantictokens.go
```diff
@@ -0,0 +1,181 @@
...
+
+			trimmedVarName := strings.Trim(varName, `"' `)
+
+			parts := strings.Split(locationStr, ":")
+			row, _ := strconv.Atoi(parts[0])
+			col, _ := strconv.Atoi(parts[1])
```
anderseknert: If we think performance matters here and want to avoid creating an intermediate slice:

```go
rowStr, rest, _ := strings.Cut(locationStr, ":")
colStr, _, _ := strings.Cut(rest, ":")

row, _ := strconv.Atoi(rowStr])
col, _ := strconv.Atoi(colStr)
```

internal/lsp/semantictokens/semantictokens.go
anderseknert: Those docs are great!

`slices.SortFunc` is the modern / more performant alternative to `sort.Slice`, so consider using that 🙂

PR #1849, opened 2026-02-01
docs(README): add mise alternative method installation

jylenhof:


srenatus:
Had a quick look, because I like mise, too: 
* https://mise-versions.jdx.dev/tools/regal this is their package site, we could link to that. It even got download stats!
* It's based on the aqua package, https://github.com/aquaproj/aqua-registry/blob/main/pkgs/open-policy-agent/regal/registry.yaml
* It still has some Styra mentions, but in general seems to [have followed the transition smoothly](https://github.com/aquaproj/aqua-registry/commit/20cf78aac1627e30b5db32d06a24612ccdefca34)

charlieegan3:
👍 let's do something similar to what OPA does, which is out of scope for this PR. We have regal packages in various other places, this is just one of those.

Looks like we might want to update https://github.com/jdx/mise/blob/main/registry/regal.toml too, also not something that needs to happen for this PR.

I'm personally happy to merge this as is, and then we will just make some updates once it's in to the various other known registries with the disclaimer that the GH releases are the official source etc.

anderseknert:
Feel free to merge and edit post merge @charlieegan3 😊

charlieegan3:
I have followed up on this one here with some website and readme adjustments for installation instructions. https://github.com/open-policy-agent/regal/pull/1861

PR #1854, opened 2026-02-06
Adding `Format stages` option to compiler explorer

johanfylling:
Found this very useful when adding template-string support to Rego.

Also surfacing previously hidden `Enable print` option; but can back out of that if the reason for hiding this option still holds.


srenatus:
If there was a reason it probably was that I couldn't get htmx to work as I wanted it to work. 💦

johanfylling:
> couldn't get htmx to work as I wanted it to work

Because it didn't send the full state of every input on form post? Don't know if it's the recommended pattern, but `hx-include="form"` seems to have done the trick.

PR #1857, opened 2026-02-09
Lint performance Rego refactor

anderseknert:
- Rewrite `walk` calls that needed the `path` from the walk so that they no longer do. This is by far the most impactful change here, accounting for 90% of the reduces allocs
- Inline and remove a few custom functions that only had a single dependent
- Avoid `count(x) == 0` and `count(x) > 0` in favor of `x == []` and `x != []` (or `""`, `set()`, etc) when the type is known
- Rewrite some `count([x | some x in y]) == z` constructs to instead use `every`
- Avoid costly path in rule-length check (counting comments) if the rule length wasn't exceeded to begin with
- Replace `some i, x in y; i > 0` with `some x in array.slice(y, 1, max)` to avoid repeated gt call
- Rewrite hot path function `ref_to_string`  to allocate much less by calling less custom functions, and replacing an `else` condition with a map lookup

Reducing almost 1/4 of all memory (B/op) previously allocated!

**BenchmarkRegalLintingItselfPrepareOnce**
```
486491514 ns/op    1906786789 B/op    45640947 allocs/op // OPA v1.13.1 + fixes
420959403 ns/op    1534395760 B/op    36917131 allocs/op // Performance refactor
```

### DIFF
--- a/bundle/regal/ast/ast.rego
+++ b/bundle/regal/ast/ast.rego
@@ -68,10 +68,7 @@ builtin_namespaces contains regex.replace(name, `\..*`, "") if some name in buil
 # description: |
 #   provides the package path values (strings) as an array starting _from_ "data":
 #   package foo.bar -> ["foo", "bar"]
-package_path := [path.value |
-	some i, path in input.package.path
-	i > 0
-]
+package_path := [path.value | some path in array.slice(input.package.path, 1, 100)]
 
 # METADATA
 # description: |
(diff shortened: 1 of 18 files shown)


bundle/regal/ast/ast.rego
```diff
@@ -258,26 +246,24 @@ is_terms_subset(terms1, terms2) if {
...
-}
+} else := sprintf(
+	{
+		"string": `["%v"]`,
+		"var": `[%v]`,
+		"number": `[%d]`,
+		"boolean": `[%v]`,
+	}[part.type],
+	[part.value],
+)
```
srenatus, line 266: Ewww

anderseknert: Lol, yes, but so much cheaper than branching with custom functions. Maybe one day it won't be 🥲

PR #1863, opened 2026-02-11
internal/explorer: remove web-page related bits

srenatus:
This now all moved into the VSCode extension. It previously only was supported for VSCode, too, so nothing was lost.

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -2360,6 +2360,10 @@ func (l *LanguageServer) handleExplorerCommand(ctx context.Context, args types.E
 	baseName := filepath.Base(uri.ToPath(args.Target))
 	baseName = strings.TrimSuffix(baseName, ".rego")
 
+	var previousOutput string
+
+	filesToOpen := make([]string, 0)
+
 	for i, cs := range compileResults {
 		var output string
 
@@ -2387,6 +2391,14 @@ func (l *LanguageServer) handleExplorerCommand(ctx context.Context, args types.E
 			continue
 		}
 
+		// Only open stages where output differs from previous stage
+		if output != previousOutput {
+			filesToOpen = append(filesToOpen, filename)
+			previousOutput = output
+		}
+	}
+
+	for _, filename := range filesToOpen {
 		takeFocus := false
 		showParams := types.ShowDocumentParams{
 			URI:       uri.FromPath(l.client.Identifier, filename),


charlieegan3:
There are some things we were interested to use the web server for, e.g. https://github.com/open-policy-agent/regal/issues/1598, perhaps https://github.com/open-policy-agent/regal/issues/1529.

There isn't an issue for it but I had the idea that we might one day have a UI that one could use to browse internal server state.

So I understand we need to remove the bit we have in the client, but I think there is still some value in the server generally. Even if after this PR it becomes a but useless for a time.

srenatus:
The server isn't entirely removed, though: https://github.com/open-policy-agent/regal/pull/1863/changes#diff-158eb27e417c4efa881226ba59673d0f8da48575bed38390b2278285fe2a3779R24

I can dial back the deletions if you'd like me to, though, of course.

charlieegan3:
Ahh, sorry! I misread that in that case. This is fine.

charlieegan3:
There seems to be a small issue here however: https://github.com/open-policy-agent/regal/actions/runs/21942818953/job/63372671994?pr=1863#step:9:118

anderseknert:
> This is fine.

this-is-fine-1169276240

But yes, seems fine with me. I'm thinking we should reinstate the Source Action as one that simply dumps the compiler stage outputs into files in a temporary directory, and redirects the client there. That would mean we provide this feature to all clients, even if not as polished. WDYT?

srenatus:
OK the non-vscode fallback is a bit crude, but not entirely useless:


https://github.com/user-attachments/assets/7df677af-dea2-4b72-9d78-e9f74d896f5b

bundle/regal/lsp/codeaction/codeaction.rego
```diff
@@ -95,24 +95,19 @@ actions contains action if {
...
 	action := {
 		"title": "Explore compiler stages for this policy",
 		"kind": "source.explore",
 		"command": {
 			"title": "Explore compiler stages for this policy",
-			"command": "vscode.open",
+			"command": "regal.explorer",
 			"tooltip": "Explore compiler stages for this policy",
-			"arguments": [explorer_url],
+			"arguments": [{"target": input.params.textDocument.uri, "format": true}],
```
srenatus, line 110: 🔍 We could send JSON for all the code actions we have -- the spec allows it. I have not changed all the other places, though. This one, however, makes sense to use like that because the VScode extension sends its payload as JSON object as well.

charlieegan3:
> OK the non-vscode fallback is a bit crude, but not entirely useless:


I wonder if it makes sense, if every step is being opened, that we only open ones that have changes? Otherwise it's a lot of files by default.

srenatus:
Now it's only opening changed stages for the fallback mode. ✨ Really much better 😅

anderseknert:
That's awesome! Thanks for sticking around @srenatus 👏

srenatus:
I'm just glad opa-explorer comes closer to what I had imagined originally :sunglasses:

anderseknert:
Would that be the VS Code menu or "a bunch of files in a temp dir"? 😆

PR #1865, opened 2026-02-11
Optimize Semantic Token Language Server Request

SeanLedford:
These changes are meant to optimize the request for semantic tokens between the client and the language server. Instead of preparing the module required to parse the textDocument prior to every request, the module is now loaded from the regal bundle.


bundle/regal/lsp/semantictokens/semantictokens.rego
```diff
@@ -6,37 +6,58 @@
...
 
 # METADATA
-# description: finds declarations of function args in text documents
-arg_tokens[var] := "declaration" if {
-	some rule_index, contexts in ast.found.vars
-	some var in contexts.args
+# description: Extract function argument declarations
+arg_tokens.declaration contains var if {
+	some rule in module.rules
+	some var in rule.head.args
```
anderseknert: It's an advanced topic sure, but args aren't necessarily vars! If they aren't they'll be pattern matched for equality in a call. So for example this definition:

```rego
f("foo", y) if {
    y > 100
}
```
Would be the same as:
```rego
f(x, y) if {
    x == "foo"
    y > 100
}
```

Here's [one example](https://github.com/open-policy-agent/regal/blob/main/bundle/regal/util/util.rego#L219C1-L226C29) of where we use this in Regal to describe an "or switch" pretty much.

So while it's not too common, you'll want to ensure the type is `var` here:
```rego
arg_tokens.declaration contains arg if {
    some rule in module.rules
    some arg in rule.head.args
    arg.type == "var"
}
```

internal/embeds/schemas/regal/lsp/semantictokens.json
```diff
@@ -0,0 +1,31 @@
...
+          "type": "string",
+          "description": "URI of the text document"
+        }
+      },
+      "required": [
+        "uri"
+      ]
+    }
+  }
+}
```
anderseknert: It's just complaining about not having a last newline in the file.

bundle/regal/lsp/semantictokens/semantictokens.rego
```diff
@@ -6,37 +6,58 @@
...
 
 # METADATA
-# description: finds variable references that are used in function calls
-arg_tokens[var] := "reference" if {
-	some rule_index, calls in ast.function_calls
-	some call in calls
-	some var in call.args
+# description: Extract variable references in function calls
+arg_tokens.reference contains var if {
+	some rule in module.rules
```
anderseknert, line 36: Add `count(rule.head.args) > 0` below or else you'll be running this on all rules (not just functions)

bundle/regal/lsp/semantictokens/semantictokens.rego
```diff
@@ -6,37 +6,58 @@
...
+	walk(rule.body, [_, expr])
+	expr.terms[0].type == "ref"
+	some var in array.slice(expr.terms, 1, count(expr.terms))
 	var.type == "var"
 
-	arg_names := {v.value | some v in ast.found.vars[rule_index].args}
+	arg_names := {v.value | some v in rule.head.args}
```
anderseknert: You'll want to do the same check for type == "var" here.

Also, this will be the same for all vars in the rule, so this should be moved up above the `walk`

bundle/regal/lsp/semantictokens/semantictokens.rego
```diff
@@ -6,37 +6,58 @@
...
+	walk(rule.body, [_, expr])
+
+	some term in expr.terms
+	term.type == "call"
+
+	some var in term.value
 	var.type == "var"
 
-	arg_names := {v.value | some v in ast.found.vars[rule_index].args}
+	arg_names := {v.value | some v in rule.head.args}
```
anderseknert: Same as above applies here.

bundle/regal/lsp/semantictokens/semantictokens.rego
SeanLedford: Ah gotcha, addressed! ~~Went ahead and changed the arg_token naming to var_token as well to be more exact.~~ On second thought, we're still checking for function args only here, just checking that they're used as variables. So I'll leave it as is

bundle/regal/lsp/semantictokens/semantictokens.rego
SeanLedford: Ah good idea, added.

bundle/regal/lsp/semantictokens/semantictokens.rego
SeanLedford: Addressed the location of the function arg name check. I did already have a type check for the arg I was pulling from the array of terms, so just want to verify that's what you meant.

PR #1867, opened 2026-02-12
lsp: Expose implemented features in server caps

charlieegan3:
This will allow clients to determine what UI functionality to load. These caps are sent to the clients as part of the init and so we can use the data to decide what to do.

### DIFF
--- a/internal/lsp/types/types.go
+++ b/internal/lsp/types/types.go
@@ -70,26 +70,41 @@ type (
 	}
 
 	ServerCapabilities struct {
-		CodeLensProvider           ResolveProviderOption   `json:"codeLensProvider"`
-		Workspace                  WorkspaceOptions        `json:"workspace"`
-		DiagnosticProvider         DiagnosticOptions       `json:"diagnosticProvider"`
-		CodeActionProvider         CodeActionOptions       `json:"codeActionProvider"`
-		ExecuteCommandProvider     ExecuteCommandOptions   `json:"executeCommandProvider"`
-		TextDocumentSyncOptions    TextDocumentSyncOptions `json:"textDocumentSync"`
-		CompletionProvider         CompletionOptions       `json:"completionProvider"`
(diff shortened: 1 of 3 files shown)


charlieegan3:
Yes indeed. This is mostly done with VSCode in mind and future regal / client changes being able to sync up. 

Hopefully other editors will follow suit, and make this more generally valuable.

internal/lsp/types/types.go
```diff
@@ -70,26 +70,41 @@ type (
...
+	ExperimentalCapabilities struct {
+		// ExplorerProvider indicates whether the server supports the regal.explorer
+		// command and regal/showExplorerResult notification.
+		ExplorerProvider bool `json:"explorerProvider"`
+		// EvalProvider indicates whether the server supports the regal.eval
+		// command and regal/showEvalResult request.
+		EvalProvider bool `json:"evalProvider"`
+		// DebugProvider indicates whether the server supports the regal.debug
+		// command and regal/startDebugging request.
+		DebugProvider bool `json:"debugProvider"`
```
anderseknert, line 107: I'm thinking we'll most likely want to move as much of the init handler into Rego sooner rather than later .. which makes me wonder if we want to user snake_case for serialized names instead? 🤔 There's no way to do have this done in Rego without actually naming rules like `debugProvider`, etc.. and that'll need inline ignores, not to mention of course that it won't be very Rego'y :)

(aside, it'd be pretty cool if metadata annotations could provide hints about how a Rego object should be serialized!)

charlieegan3: Perhaps, and we do have control of these for the client and the server.

However, others like `hoverProvider` will need to remain the same regardless of the Go/Rego implementation as that's the spec.

I think my preference for now is to use the same style as the existing field names.

anderseknert: Fair point!

PR #1870, opened 2026-02-14
Add Semantic Token Functionality for Imports

SeanLedford:
- Adds semantic token functionality for imports
- Addresses bug related to semantic token values for quoted package strings

### DIFF
--- a/bundle/regal/lsp/semantictokens/semantictokens.rego
+++ b/bundle/regal/lsp/semantictokens/semantictokens.rego
@@ -20,6 +20,16 @@ module := data.workspace.parsed[input.params.textDocument.uri]
 result.response := {
 	"arg_tokens": arg_tokens,
 	"package_tokens": package_tokens,
+	"import_tokens": import_tokens,
+}
+
+# METADATA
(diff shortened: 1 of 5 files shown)


anderseknert (approved):
Looks good to me!

As often (always?), there's a special case that we might want to consider too, which is aliased imports. I.e:

```rego
import data.foo.bar as baz
```

This can be useful sometimes when an imported identifier clashes with names in the package, or perhaps something built-in. Aliases are really not nice to deal with in the AST though!  See for example:

- https://github.com/open-policy-agent/regal/issues/1723
- https://github.com/open-policy-agent/regal/issues/1145

That's really an original sin carried on from OPA, but one we should improve here at some point. As you'll notice reading what's described there, aliases are represented as nothing but strings! Like, they have **no location** even though they are clearly placed somewhere 😅 So we've come up with all sorts of hacks to sort of deal with them in the past, and we'll likely need to do that here to. 

I don't mean that you need to do this as part of this work — only mention that for your awareness and perhaps to give some thought to how we could improve that situation  going forward.

PR #1871, opened 2026-02-14
test: Add assert/must helper modules

anderseknert:
I've had enough of simple assertions spanning 3 lines, and the resulting scroll fest. We don't need anything advanced, so rather than adding a dependency I've just added the most commonly needed assertions, split up across an `assert` module and a `must` module.

Nothing fancy here, just 1300 lines of boilerplate code removed, and a couple of issues fixed in our tests while I worked on this.

This isn't complete by any means, and I think some of the assert functions can have their output improved, some can be named better, and so on. But that'll be improved as we work with these I'm sure.


charlieegan3:
Happy to have this, but the issue with testing libraries is that people forget to use them (especially new contributors). So let's try and make sure that we A) remind ourselves / them and B) move what we can to Rego tests eventually.

PR #1872, opened 2026-02-16
add initial rq engine support to regal

charlesdaniels:
* Update `do.rq fetch` to download capabilities files from the rq git repository, in a similar manner to how it is already done for EOPA.
* Update embedded capabilities so that a new `rq` engine is recognized, in a similar manner to the existing `eopa` engine.
    * Added test cases for the same.
* Update `no-whitespace-comment` rule so that comments on the first line of the file beginning with `#!` are ignored, so that shebangs do not trip the linter rule. This applies to all Rego files, not only rq.
    * Added test cases for the same.
* Update `NotRego()` filter to include `.rq` files as valid Rego files. This applies globally, not only when the `rq` engine is selected. 
* **known limitation**: `rq` scripts which elide the `package` statement (allowed by `rq`, but not mainline OPA) cannot be parsed correctly.
    * Initially, I experimented with injecting the `package` statement in a similar manner to `rq`, but it seems that I would have to make modifications in several places. It seemed to produce slightly wrong output. We can revisit this in the future. For now, requiring `package` statements to be able to use Regal seems like a reasonable compromise.
* Bumped the rq version used in the update-caps workflow from 0.0.9 to 0.0.15.

Happy to discuss if changes are needed. 

Demo video:

https://github.com/user-attachments/assets/7ac8eb6e-2ee9-4eed-90de-d2f8f7f4e6df

### DIFF
--- a/docs/configuration/capabilities.md
+++ b/docs/configuration/capabilities.md
@@ -79,4 +79,4 @@ Regal includes capabilities files for the following engines:
 | `eopa` | [Enterprise OPA website](https://www.styra.com/enterprise-opa/) | Styra Enterprise OPA |
 | `rq`   | [rq website](https://git.sr.ht/~charles/rq)                     | Rego Query (`rq`)    |
 
-**`rq` support note**: `rq` scripts must include `package` statements to be compatible with Regal. 
+**`rq` support note**: `rq` scripts must include `package` statements to be compatible with Regal.


.github/workflows/update-caps.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       contents: write
       pull-requests: write
     env:
-      RQ_VERSION: v0.0.9
+      RQ_VERSION: v0.0.15
```
anderseknert: Let's use v0.0.16 since that seems to be tracking a more recent version of OPA, and importantly one that supports string interpolation. Or any reason not to?

build/do.rq
```diff
@@ -303,6 +307,86 @@ fetch_eopa_caps if {
...
+	{"local": p, "remote": r} |
+		t := rq_tags[_]
+		p := rq.joinpath([rq_caps_dir, sprintf("%s.json", [t])])
+
+		not rq_caps_tree[p]
+
+		r := rq.template("https://git.sr.ht/~charles/rq/blob/{{.tag}}/capabilities.json", {"tag": t})
+	}
+
+	print(sprintf("fetching %d capabilities files missing locally", [count(missing_locally)]))
```
anderseknert:
```suggestion
print("fetching", count(missing_locally), "capabilities files missing locally")

File: build/do.rq [hunk 3]
```diff
@@ -303,6 +307,86 @@ fetch_eopa_caps if {
...
+
+	new_caps := {
+	{"local": m.local, "content": c} |
+		m := missing_locally[_]
+		print("\tfetching ", m.remote)
+		resp := http.send({"url": m.remote, "method": "GET"})
+
+		{rq.error(sprintf("non-200 status code '%d' for URL '%s'", [resp.status_code, m.remote])) | resp.status_code != 200}
```
anderseknert, line 379: No `rq.errorf`? 🙂

.github/workflows/update-caps.yml
charlesdaniels: For sure, 0.0.16 was released after this PR.

charlesdaniels: 85addc3e

charlesdaniels:
I thought it might be wise to document the limitation that rq scripts must have `package` statements in the actual docs and not only in this PR description, so I added it in baf27797.

build/do.rq
charlesdaniels: https://todo.sr.ht/~charles/rq/21

anderseknert: Neat!

anderseknert:
@charlesdaniels thanks again! I'd like to also see a PR to have the OPA VS Code extension recognize .rq for syntax highlighting, as the second it takes to manually tell it too always bothers me a little :)

PR #1873, opened 2026-02-16
Add `use-array-flatten` and `use-object-union-n` rules

### LINKED ISSUE #1855: Rule: use-array-flatten
It's been a while since we had a new rule for recommending the use of a built-in function. But it struck me now that since my rationale for adding `array.flatten` to OPA was to be able to replace code like this..

```rego
flat := array.concat([a], array.concat(b, [c]))
```
with this..
```rego
flat := array.flatten([a, b, c])
```
..we should of course let our users know that they too can do this! Let's create a rule that recognizes nested `array.concat` calls and suggests using `array.flatten` instead — unless of course that built-in is not in the target OPA's capabilities.

Thinking about it, we can probably do this for a few more functions, like recommending `object.union_n` if we find nested calls to `object.union`. I also looked at suggesting to replace repeated `replace` calls -> `strings.repeat_n`... but AFAICS those aren't interchangeable.Oh well, that should be another issue/rule.

anderseknert:
With a few config knobs for the config inclined.

Fixes #1855

### DIFF
--- a/bundle/regal/ast/keywords.rego
+++ b/bundle/regal/ast/keywords.rego
@@ -88,14 +88,14 @@ _keywords_with_location(value) := keywords if {
 	value.terms.symbols
 
 	location := util.to_location_object(value.terms.location)
-	keywords := array.concat([{"name": "some", "location": location}], _in_on_row(location.row))
+	keywords := array.flatten([{"name": "some", "location": location}, _in_on_row(location.row)])
 }
 
 _keywords_with_location(value) := keywords if {
 	value.domain
 
 	location := util.to_location_object(value.location)
-	keywords := array.concat([{"name": "every", "location": location}], _in_on_row(location.row))
+	keywords := array.flatten([{"name": "every", "location": location}, _in_on_row(location.row)])
 }
 
 _in_on_row(row) := [keyword |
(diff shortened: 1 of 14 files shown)


bundle/regal/ast/keywords.rego
```diff
@@ -88,14 +88,14 @@ _keywords_with_location(value) := keywords if {
 	value.terms.symbols
 
 	location := util.to_location_object(value.terms.location)
-	keywords := array.concat([{"name": "some", "location": location}], _in_on_row(location.row))
+	keywords := array.flatten([{"name": "some", "location": location}, _in_on_row(location.row)])
```
charlieegan3, line 91: Very good!

bundle/regal/rules/idiomatic/use-array-flatten/use_array_flatten.rego
```diff
@@ -0,0 +1,91 @@
...
+	violation := result.fail(rego.metadata.chain(), result.location(call))
+}
+
+report contains violation if {
+	not config.rules.idiomatic["use-array-flatten"]["flag-all-concat"]
+	config.rules.idiomatic["use-array-flatten"]["flag-wrapped-concat"] == true
+
+	some call in wrapped_concat_calls
+
+	violation := json.patch(result.fail(rego.metadata.chain(), result.location(call)), [{
```
charlieegan3, line 31: We do this in naming_convention.rego too with a hhelper `_with_description(violation, description)`. Which these, that makes three rules now. Might it be useful in other rules, or in new rules that need to have the desc configured.

anderseknert: Oh, I didn't realize. That's a good call, yes! Will update in the next PR (have code lined up already :))

PR #1880, opened 2026-02-17
lsp: Make capabilities consistent with client

charlieegan3:
This PR follows https://github.com/open-policy-agent/regal/pull/1867, and is related to https://github.com/open-policy-agent/vscode-opa/pull/406

To make better support mismatched server and client versions.

https://github.com/user-attachments/assets/ea8ea58e-599f-4f34-ab88-4ec2843f6d82

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -94,6 +94,16 @@ var (
 	fixRedundantExistence     = &fixes.RedundantExistenceCheck{}
 )
 
+// DefaultServerFeatureFlags returns the default feature flags with all
+// custom features enabled.
+func DefaultServerFeatureFlags() *types.ServerFeatureFlags {
+	return &types.ServerFeatureFlags{
+		ExplorerProvider:         true,
+		InlineEvaluationProvider: true,
+		DebugProvider:            true,
+	}
+}
+
 type LanguageServerOptions struct {
 	// Logger is the logger to use for the language server.
 	Logger *log.Logger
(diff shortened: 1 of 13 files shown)


internal/lsp/server.go
```diff
@@ -1880,6 +1892,29 @@ func (l *LanguageServer) handleInitialize(ctx context.Context, params types.Init
 	regoFilter := types.FileOperationFilter{Scheme: "file", Pattern: types.FileOperationPattern{Glob: "**/*.rego"}}
 	fileOpOpts := types.FileOperationRegistrationOptions{Filters: []types.FileOperationFilter{regoFilter}}
 
+	// set enabled commands based on server feature flags
+	enabledCommands := []string{
+		"regal.eval",
```
charlieegan3, line 1897: eval command is always on, just diff behaviour

internal/lsp/server_test.go
```diff
@@ -87,7 +91,18 @@ func createAndInitServer(t *testing.T, ctx context.Context, tempDir string, clie
...
 
-	request := types.InitializeParams{RootURI: rootURI, ClientInfo: types.ClientInfo{Name: "go test"}}
+	// all client features enabled by default
+	flagState := true
+
+	request := types.InitializeParams{
+		RootURI:    rootURI,
+		ClientInfo: types.ClientInfo{Name: "go test"},
+		InitializationOptions: &types.InitializationOptions{
+			EnableDebugCodelens:       &flagState,
```
srenatus: [nit] It would be storing two extra booleans, but we can now do something [like this](https://go.dev/play/p/CJprY-wl9Yc):

```suggestion
			EnableDebugCodelens:       new(true),
```

charlieegan3: Nice, have updated!

PR #1883, opened 2026-02-19
Semantic Token Support for Comprehensions and Constructs

SeanLedford:
Adds functionality for syntax highlighting of comprehensions and constructs via semantic tokens! Also fixes some bugs introduced by the original framework, such as empty files causing the language server to error.

### DIFF
--- a/bundle/regal/lsp/semantictokens/vars/comprehensions/comprehensions.rego
+++ b/bundle/regal/lsp/semantictokens/vars/comprehensions/comprehensions.rego
@@ -11,26 +11,18 @@ import data.regal.ast
 # METADATA
 # description: Extract comprehension variable declarations from array/set/object comprehensions
 result.declaration contains var if {
-	# regal ignore:prefer-some-in-iteration
-	comprehensions := ast.found.comprehensions[_]
-	some comprehension in comprehensions
+	comprehension := ast.found.comprehensions[_][_]
 
-	comp_vars := {v |
-		some term in comprehension.value.body
-		term.terms.symbols
-		some symbol in term.terms.symbols
-		some v in array.slice(symbol.value, 1, count(symbol.value) - 1)
-		v.type == "var"
-	}
-	some var in comp_vars
+	some expr in comprehension.value.body
+	some symbol in expr.terms.symbols
+	some var in array.slice(symbol.value, 1, count(symbol.value) - 1)
+	var.type == "var"
 }
 
 # METADATA
 # description: Extract comprehension variable references in the output
 result.reference contains var if {
-	# regal ignore:prefer-some-in-iteration
-	comprehensions := ast.found.comprehensions[_]
-	some comprehension in comprehensions
+	comprehension := ast.found.comprehensions[_][_]
 
 	output_vars := array.flatten([
 		_comprehension_key(comprehension),
--- a/bundle/regal/lsp/semantictokens/vars/function_args/function_args.rego
+++ b/bundle/regal/lsp/semantictokens/vars/function_args/function_args.rego
@@ -24,16 +24,14 @@ result.declaration contains arg if {
 # description: Extract variable references in function calls
 result.reference contains arg if {
 	some rule in module.rules
+	some rule_index
 
 	rule.head.args
 
 	arg_names := ast.function_arg_names(rule)
 
-	walk(rule.body, [_, expr])
-
-	expr.terms[0].type == "ref"
-
-	some arg in array.slice(expr.terms, 1, count(expr.terms))
+	call := ast.found.calls[rule_index][_]
+	some arg in array.slice(call, 1, 100)
 
 	arg.type == "var"
 	arg.value in arg_names
@@ -43,7 +41,7 @@ result.reference contains arg if {
 # description: Extract variable references in call expressions
 result.reference contains arg if {
 	some rule in module.rules
-	arg_names := {v.value | some v in rule.head.args}
+	arg_names := ast.function_arg_names(rule)
 	walk(rule.body, [_, expr])
 
 	some term in expr.terms
--- a/bundle/regal/lsp/semantictokens/vars/imports/imports.rego
+++ b/bundle/regal/lsp/semantictokens/vars/imports/imports.rego
@@ -10,8 +10,9 @@ package regal.lsp.semantictokens.vars.imports
 # description: Get the module from workspace
 module := data.workspace.parsed[input.params.textDocument.uri]
 
-# METADATA
-# description: Extract import tokens - return only last term of the path
+# METADATA:
+# description: |
+#   Extract import tokens - return only last term of the path. Does not currently highlight aliased identifiers
 result contains last_term if {
 	some import_statement in module.imports
 	import_path := import_statement.path.value
--- a/bundle/regal/lsp/semantictokens/vars/packages/packages.rego
+++ b/bundle/regal/lsp/semantictokens/vars/packages/packages.rego
@@ -12,8 +12,4 @@ module := data.workspace.parsed[input.params.textDocument.uri]
 
 # METADATA
 # description: Extract package tokens - return full package path
-result contains last_term if {
-	package_path := module.package.path
-
-	last_term := regal.last(package_path)
-}
+result contains regal.last(module.package.path)
--- a/internal/lsp/semantictokens/semantictokens.go
+++ b/internal/lsp/semantictokens/semantictokens.go
@@ -24,11 +24,11 @@ const (
 )
 
 type Token struct {
-	Line      uint
-	Col       uint
-	Length    uint
-	Type      uint
-	Modifiers uint
+	Line      uint32
+	Col       uint32
+	Length    uint32
+	Type      uint32
+	Modifiers uint32
 }
 
 // Represents location data from the AST
@@ -37,10 +37,10 @@ type ASTLocation struct {
 }
 
 type LocationInfo struct {
-	Line        uint
-	StartColumn uint
-	EndColumn   uint
-	Length      uint
+	Line        uint32
+	StartColumn uint32
+	EndColumn   uint32
+	Length      uint32
 }
 
 func (loc *ASTLocation) UnmarshalJSON(data []byte) error {
@@ -69,10 +69,10 @@ func (loc *ASTLocation) UnmarshalJSON(data []byte) error {
 	}
 
 	loc.Location = LocationInfo{
-		Line:        uint(row),
-		StartColumn: uint(startcol),
-		EndColumn:   uint(endcol),
-		Length:      uint(endcol - startcol),
+		Line:        uint32(row),
+		StartColumn: uint32(startcol),
+		EndColumn:   uint32(endcol),
+		Length:      uint32(endcol - startcol),
 	}
 
 	return nil
@@ -204,7 +204,7 @@ func processImportTokens(importTokens []ASTLocation) ([]Token, error) {
 	return tokens, nil
 }
 
-func extractTokens(astLoc ASTLocation, tokenType uint, modifiers uint) (Token, error) {
+func extractTokens(astLoc ASTLocation, tokenType uint32, modifiers uint32) (Token, error) {
 	return Token{
 		Line:      astLoc.Location.Line - 1,
 		Col:       astLoc.Location.StartColumn - 1,
@@ -216,7 +216,7 @@ func extractTokens(astLoc ASTLocation, tokenType uint, modifiers uint) (Token, e
 
 func encodeTokens(tokens []Token) *types.SemanticTokens {
 	if len(tokens) == 0 {
-		return &types.SemanticTokens{Data: []uint{}}
+		return &types.SemanticTokens{Data: []uint32{}}
 	}
 
 	// Sort tokens by position (line first, then column)
@@ -227,9 +227,9 @@ func encodeTokens(tokens []Token) *types.SemanticTokens {
 		return int(a.Col) - int(b.Col)
 	})
 
-	data := make([]uint, 0, len(tokens)*5)
+	data := make([]uint32, 0, len(tokens)*5)
 
-	var prevLine, prevCol uint
+	var prevLine, prevCol uint32
 
 	for _, tok := range tokens {
 		deltaLine := tok.Line - prevLine
--- a/internal/lsp/semantictokens_test.go
+++ b/internal/lsp/semantictokens_test.go
@@ -208,14 +208,14 @@ some_one_var_construct if {
 // semanticTokenInstance adds structure to uint data stream in the SemanticToken
 // return type making it more readable for error messages and comparisons in tests
 type semanticTokenInstance struct {
-	DeltaLine uint
-	DeltaCol  uint
-	Length    uint
-	Type      uint
-	Modifier  uint
+	DeltaLine uint32
+	DeltaCol  uint32
+	Length    uint32
+	Type      uint32
+	Modifier  uint32
 }
 
-func uintsToTestTokens(data []uint) []semanticTokenInstance {
+func uintsToTestTokens(data []uint32) []semanticTokenInstance {
 	if len(data)%5 != 0 {
 		panic("invalid token data length, must be multiple of 5")
 	}
--- a/internal/lsp/types/types.go
+++ b/internal/lsp/types/types.go
@@ -583,8 +583,8 @@ type (
 	}
 
 	SemanticTokens struct {
-		ResultID *string `json:"resultId,omitempty"`
-		Data     []uint  `json:"data"`
+		ResultID *string  `json:"resultId,omitempty"`
+		Data     []uint32 `json:"data"`
 	}
 
 	SemanticTokensParams struct {


bundle/regal/lsp/semantictokens/semantictokens.rego
```diff
@@ -15,12 +15,18 @@ package regal.lsp.semantictokens
 # description: Get the module from workspace
 module := data.workspace.parsed[input.params.textDocument.uri]
 
+# This is handling the case where the module from the parsed workspace is empty
+
```
charlieegan3:
```suggestion
```

bundle/regal/lsp/semantictokens/semantictokens.rego
```diff
@@ -78,3 +84,130 @@ arg_tokens.reference contains arg if {
...
+} else := value if {
+	comprehension.type == "objectcomprehension"
+	value := comprehension.value.value
+} else := set()
+
+# Helper to get variables from differing comprehensions
+_get_comprehension_key(comprehension) := value if {
+	comprehension.type == "objectcomprehension"
+	value := comprehension.value.key
+} else := set()
```
charlieegan3: I think it might be nice to have comprehensions and 'constructs' in their own files in the same package. this is only going to get longer as we add more.

bundle/regal/lsp/semantictokens/semantictokens.rego
```diff
@@ -78,3 +84,130 @@ arg_tokens.reference contains arg if {
 # METADATA
 # description: Extract package tokens - return full package path
 package_tokens := module.package.path
+
+# METADATA
+# description: Extract comprehension variable declarations from array/set/object comprehensions
+comprehension_tokens.declaration contains var if {
```
charlieegan3: comprehension_tokens and 'construct' tokens walk over each rule in the module.rules, find one or two vars, and then return.

This is similar to a lot of what we have in regal's ast package. Let's see what we can reuse from that.

bundle/regal/lsp/semantictokens/semantictokens.rego
```diff
@@ -78,3 +84,130 @@ arg_tokens.reference contains arg if {
...
+	some rule in module.rules
+	walk(rule.head.value, [_, comprehension])
+	comp_vars := _comprehension_vars(comprehension)
+	some var in comp_vars
+}
+
+# METADATA
+# description: Extract comprehension variable references in the output
+comprehension_tokens.reference contains var if {
```
charlieegan3: I think that if we can, ast.found.comprehensions might be good to use here.

bundle/regal/lsp/semantictokens/semantictokens.rego
```diff
@@ -78,3 +84,130 @@ arg_tokens.reference contains arg if {
...
+	walk(rule.body, [_, term])
+
+	declared_vars := _get_construct_vars(term.terms)
+
+	some var in declared_vars
+}
+
+# METADATA
+# description: Extract variable references in every and some constructs
+construct_tokens.reference contains var if {
```
charlieegan3: similarly, we might be able to get some and every like this

https://github.com/open-policy-agent/regal/blob/main/bundle/regal/rules/style/unnecessary-some/unnecessary_some.rego#L12-L15

bundle/regal/lsp/semantictokens/semantictokens.rego
SeanLedford: For sure, good idea. Went ahead and added them to their own separate files, with their own test files as well!

bundle/regal/lsp/semantictokens/semantictokens.rego
SeanLedford: I've used ast package helpers `found.comprehensions` and `found.vars["every"]`, to tidy up some of the code here. Happy to try other optimizations if you have any in mind!

bundle/regal/lsp/semantictokens/semantictokens.rego
SeanLedford: Yep, that helped tidy up some of the code here

bundle/regal/lsp/semantictokens/semantictokens.rego
SeanLedford: I used found.vars to get the declarations in the some/every contexts! This worked pretty well for my use case, let me know what you think.

bundle/regal/lsp/semantictokens/vars/comprehensions/comprehensions.rego
```diff
@@ -0,0 +1,74 @@
...
+# schemas:
+#   - input:        schema.regal.lsp.common
+#   - input.params: schema.regal.lsp.semantictokens
+package regal.lsp.semantictokens.vars.comprehensions
+
+import data.regal.ast
+
+# METADATA
+# description: Extract comprehension variable declarations from array/set/object comprehensions
+result.declaration contains var if {
```
charlieegan3, line 13: I think this result.declaration and reference is nice!

bundle/regal/lsp/semantictokens/vars/comprehensions/comprehensions.rego
```diff
@@ -0,0 +1,74 @@
...
+} else := value if {
+	comprehension.type == "objectcomprehension"
+	value := comprehension.value.value
+} else := set()
+
+# Helper to get variables from differing comprehensions
+_get_comprehension_key(comprehension) := value if {
+	comprehension.type == "objectcomprehension"
+	value := comprehension.value.key
+} else := set()
```
charlieegan3: might be nice to have a default rule instead of the else here.

charlieegan3: same for the above.

bundle/regal/lsp/semantictokens/vars/comprehensions/comprehensions_test.rego
```diff
@@ -0,0 +1,39 @@
+package regal.lsp.semantictokens.vars.comprehensions_test
+
+import data.regal.lsp.semantictokens.vars.comprehensions
+
+test_comprehensions if {
```
charlieegan3: I would make this three separate tests. At the moment the locations are going to be pain to update if there is a small change and there are many assertions on the data.

Have a look at https://www.openpolicyagent.org/docs/policy-testing#parameterized-tests-and-data-driven-testing, it might be good to use here with three different mini modules for each type.

bundle/regal/lsp/semantictokens/vars/every_expr/every_expr.rego
```diff
@@ -0,0 +1,36 @@
...
+
+# METADATA
+# description: Get the module from workspace
+module := data.workspace.parsed[input.params.textDocument.uri]
+
+# METADATA
+# description: Extract variable declarations from every keyword domains
+result.declaration contains var if {
+	some rule_index in ast.rule_index_strings
+	some var in ast.found.vars[rule_index]["every"]
```
charlieegan3:
```suggestion
	some rule_index
	some var in ast.found.vars[rule_index]["every"]
```

I think you can just do this.

bundle/regal/lsp/semantictokens/vars/every_expr/every_expr_test.rego
```diff
@@ -0,0 +1,30 @@
+package regal.lsp.semantictokens.vars.every_expr_test
+
+import data.regal.lsp.semantictokens.vars.every_expr
+
+test_every_vars if {
+	policy := `package regal.woo
+	
+every_two_vars_construct if {
```
charlieegan3: I would be included to make this two tests again, perhaps using https://www.openpolicyagent.org/docs/policy-testing#parameterized-tests-and-data-driven-testing too.

I'd also avoid using 'construct' here, just call them "single every", "nested every" or similar I think.

bundle/regal/lsp/semantictokens/vars/function_args/function_args_test.rego
```diff
@@ -0,0 +1,41 @@
...
+
+	{"location": "3:15:3:21", "value": "param1", "type": "var"} in tokens.declaration
+	{"location": "3:23:3:29", "value": "param2", "type": "var"} in tokens.declaration
+
+	{"location": "4:11:4:17", "value": "param1", "type": "var"} in tokens.reference
+	{"location": "5:11:5:17", "value": "param2", "type": "var"} in tokens.reference
+	{"location": "9:11:9:17", "value": "param1", "type": "var"} in tokens.reference
+}
+
+test_function_args_declarations_only if {
```
charlieegan3, line 27: Nice, I like this better as two tests like this

bundle/regal/lsp/semantictokens/vars/imports/imports.rego
```diff
@@ -0,0 +1,20 @@
...
+# description: Get the module from workspace
+module := data.workspace.parsed[input.params.textDocument.uri]
+
+# METADATA
+# description: Extract import tokens - return only last term of the path
+result contains last_term if {
+	some import_statement in module.imports
+	import_path := import_statement.path.value
+
+	last_term := import_path[count(import_path) - 1]
```
charlieegan3: We actually have a custom built in called `regal.last` for this that you might want to use here.

charlieegan3: This is more performant than getting the final elem in rego

bundle/regal/lsp/semantictokens/vars/packages/packages.rego
```diff
@@ -0,0 +1,19 @@
...
+# METADATA
+# description: Get the module from workspace
+module := data.workspace.parsed[input.params.textDocument.uri]
+
+# METADATA
+# description: Extract package tokens - return full package path
+result contains last_term if {
+	package_path := module.package.path
+
+	last_term := package_path[count(package_path) - 1]
```
charlieegan3: Same comment as above about regal.last.

bundle/regal/lsp/semantictokens/vars/some_expr/some_expr_test.rego
```diff
@@ -0,0 +1,28 @@
...
+some_two_vars_construct if {
+	some i, item in input.array   
+	i < 10                        
+	item > 0                        
+}
+
+some_one_var_construct if {
+	some i in input.array   
+	i < 10                                              
+}`
```
charlieegan3, line 38:
```suggestion
}`

```

bundle/regal/lsp/semantictokens/vars/some_expr/some_expr_test.rego
```diff
@@ -0,0 +1,28 @@
+package regal.lsp.semantictokens.vars.some_expr_test
+
+import data.regal.lsp.semantictokens.vars.some_expr
+
+test_some_vars if {
```
charlieegan3, line 5: Again, might be nice broken down into two smaller tests.

bundle/regal/lsp/semantictokens/vars/some_expr/some_expr.rego
```diff
@@ -0,0 +1,35 @@
...
+import data.regal.ast
+
+# METADATA
+# description: Get the module from workspace
+module := data.workspace.parsed[input.params.textDocument.uri]
+
+# METADATA
+# description: Extract variable declarations from some keyword domains
+result.declaration contains var if {
+	some context in ["somein", "some"]
```
charlieegan3: I'd personally use a set here, rather than an array since the order does not matter.

bundle/regal/lsp/semantictokens/vars/some_expr/some_expr.rego
```diff
@@ -0,0 +1,35 @@
...
+	some rule_index in ast.rule_index_strings
+	some var in ast.found.vars[rule_index][context]
+}
+
+# METADATA
+# description: Extract variable references in some keyword domains
+result.reference contains var if {
+	some rule_index in ast.rule_index_strings
```
charlieegan3: I'd move this nearer the declared_var_names and just use `some rule_index` as above.

bundle/regal/lsp/semantictokens/semantictokens.rego
```diff
@@ -4,77 +4,38 @@
...
-
-	some term in expr.terms
-	term.type == "call"
-
-	some arg in term.value
-	arg.type == "var"
-
-	arg.value in arg_names
+# entrypoint: true
+result["response"] := {
```
charlieegan3, line 32: I think this is a good improvement, nice to have that structure as this list grows.

bundle/regal/lsp/semantictokens/vars/function_args/function_args.rego
```diff
@@ -0,0 +1,54 @@
...
+}
+
+# METADATA
+# description: Extract variable references in function calls
+result.reference contains arg if {
+	some rule in module.rules
+
+	rule.head.args
+
+	arg_names := {v.value | some v in rule.head.args}
```
charlieegan3: We also have a function called function_arg_names, perhaps we might be able o use that?

bundle/regal/lsp/semantictokens/vars/function_args/function_args.rego
```diff
@@ -0,0 +1,54 @@
...
+
+	walk(rule.body, [_, expr])
```
charlieegan3: and then @anderseknert might have better idea, but might ast.find_vars_in_local_scope work here, and then just match with the vars from the args?

bundle/regal/lsp/semantictokens/vars/comprehensions/comprehensions.rego
SeanLedford: Yeah good idea, addressed!

bundle/regal/lsp/semantictokens/vars/comprehensions/comprehensions_test.rego
SeanLedford: Gotcha, yeah I went ahead and adjusted this test using the format here. Wanted to get your feedback before I updated the other tests that you commented on.

bundle/regal/lsp/semantictokens/vars/imports/imports.rego
SeanLedford: Ah nice, addressed!

bundle/regal/lsp/semantictokens/vars/function_args/function_args.rego
SeanLedford: Nice yeah that cleaned things up a bit, addressed!

bundle/regal/ast/search.rego
```diff
@@ -209,6 +297,15 @@ found.symbols[rule_index] contains value.symbols if {
 	walk(_rules[i], [_, value])
 }
 
+# METADATA
+# description: all every constructs found in module
+found.every[rule_index] contains terms if {
```
charlieegan3, line 214: nice addition here. I wonder if there are other places we can use in rules etc. (bonus side quest, not a main quest issue at all and not needed before merge)

bundle/regal/lsp/semantictokens/semantictokens_test.rego
```diff
@@ -2,71 +2,37 @@ package regal.lsp.semantictokens_test
...
-	count(tokens.declaration) == 0
-	count(tokens.reference) == 0
+	result == {"response": {
+		"imports": {{"location": "3:19:3:22", "type": "string", "value": "ast"}},
+		"packages": {{"location": "1:15:1:18", "type": "string", "value": "woo"}},
+		"vars": {
+			"comprehensions": {
+				"declaration": set(),
+				"reference": set(),
+			},
```
charlieegan3, line 23: So we need the

```
{
				"declaration": set(),
				"reference": set(),
			}
```

part, as it's expected somewhere else, but if we didn't expect it, then this test could be cleaner without the empty parts...

Perhaps there was a good reason to always have these set?

internal/lsp/rego/router.go
```diff
@@ -261,7 +261,7 @@ func semanticTokensHandler(ctx context.Context, rctx *RegalContext, req *jsonrpc
 		return nil, err
 	}
 
-	result, err := QueryEval[types.SemanticTokensParams, map[string]any](
+	result, err := QueryEval[types.SemanticTokensParams, semantictokens.SemanticTokensResult](
```
charlieegan3, line 264: nice, we like types.

internal/lsp/semantictokens/semantictokens.go
```diff
@@ -30,164 +30,187 @@ type Token struct {
...
+type SemanticTokensResult struct {
+	PackageTokens []ASTLocation `json:"packages"`
+	ImportTokens  []ASTLocation `json:"imports"`
+	Vars          VarsSection   `json:"vars"`
+	DebugInfo     interface{}   `json:"debug_info"`
+}
 
-					rowStr, rest, _ := strings.Cut(locationStr, ":")
-					colStr, _, _ := strings.Cut(rest, ":")
+func Full(ctx context.Context, result SemanticTokensResult) (*types.SemanticTokens, error) {
```
charlieegan3, line 103: This function is long enough that it might be nice to break down now, we're going to get a maintidx lint error pretty soon at this rate and we do plan to add more here.

internal/lsp/semantictokens/semantictokens.go
```diff
@@ -30,164 +30,187 @@ type Token struct {
...
+	Reference   []ASTLocation `json:"reference,omitempty"`
+}
 
-	if packageTokens, ok := queryResult["package_tokens"]; ok {
+// Represents the vars section containing different token categories
+type VarsSection struct {
+	ArgTokens           ArgTokenCategory `json:"function_args"`
+	ComprehensionTokens ArgTokenCategory `json:"comprehensions"`
+	EveryTokens         ArgTokenCategory `json:"every_expr"`
+	SomeTokens          ArgTokenCategory `json:"some_expr"`
```
charlieegan3, line 92:
```suggestion
	FunctionArgTokens           ArgTokenCategory `json:"function_args"`
	ComprehensionTokens ArgTokenCategory `json:"comprehensions"`
	EveryExprTokens         ArgTokenCategory `json:"every_expr"`
	SomeExprTokens          ArgTokenCategory `json:"some_expr"`
```

Just an idea, to make the mapping / naming more consistent.

SeanLedford: Yeah I like this!

bundle/regal/lsp/semantictokens/vars/comprehensions/comprehensions.rego
```diff
@@ -0,0 +1,78 @@
...
+
+import data.regal.ast
+
+# METADATA
+# description: Extract comprehension variable declarations from array/set/object comprehensions
+result.declaration contains var if {
+	comprehensions := [c |
+		c := ast.found.comprehensions[_][_]
+	]
+	some comprehension in comprehensions
```
charlieegan3:
```suggestion
    some foo, bar
	some comprehension in ast.found.comprehensions[foo][bar]
```

I think we can do it more like this, or similar to avoid allocating a new list of comprehensions

bundle/regal/lsp/semantictokens/vars/comprehensions/comprehensions.rego
```diff
@@ -0,0 +1,78 @@
...
+	comp_vars := _comprehension_vars(comprehension)
+	some var in comp_vars
+}
+
+# METADATA
+# description: Extract comprehension variable references in the output
+result.reference contains var if {
+	comprehensions := [c |
+		c := ast.found.comprehensions[_][_]
+	]
```
charlieegan3: Same as above.

bundle/regal/lsp/semantictokens/vars/comprehensions/comprehensions.rego
```diff
@@ -0,0 +1,78 @@
...
+		_get_comprehension_value(comprehension),
+		_get_comprehension_vars(comprehension.value.body),
+	])
+	some var in output_vars
+	var.type == "var"
+	var.value in {v.value | some v in comp_vars}
+}
+
+# Helper function to get declared variables in a comprehension
+_comprehension_vars(comprehension) := comp_vars if {
```
charlieegan3: It'd be nice to make this a rule rather than a function since we are calling it twice for each comprehension at the moment.

internal/lsp/semantictokens/semantictokens.go
```diff
@@ -30,164 +30,175 @@ type Token struct {
...
-					row, _ := strconv.Atoi(rowStr)
-					col, _ := strconv.Atoi(colStr)
+	return tokens, nil
+}
 
-					length, err := getTokenLengthFromLocation(locationStr)
-					if err != nil {
-						return nil, fmt.Errorf("failed to get token length: %w", err)
-					}
+func extractTokens(astLoc ASTLocation, tokenType uint, modifiers uint) (Token, error) {
```
charlieegan3: I feel like some of this logic and perhaps the ASTLocation type might be best in the roast package instead. This is a roast function and I feel handling roast's format would be best done in that package.

internal/lsp/semantictokens/semantictokens.go
```diff
@@ -231,7 +242,6 @@ func encodeTokens(tokens []Token) *types.SemanticTokens {
 	return &types.SemanticTokens{Data: data}
 }
 
-// getTokenLengthFromLocation calculates token length from location span
 func getTokenLengthFromLocation(locationStr string) (uint, error) {
```
charlieegan3: Similar for this function, I think it'd be nice for this to be a method on the ASTLocation type or similar.

Perhaps we can just go straight from roast string location to an ast.Location and then we don't need to process its length more than once etc.

bundle/regal/lsp/semantictokens/vars/comprehensions/comprehensions.rego
```diff
@@ -0,0 +1,78 @@
...
+	comp_vars := {v |
+		some term in comprehension.value.body
+		term.terms.symbols
+		some symbol in term.terms.symbols
+		some v in array.slice(symbol.value, 1, count(symbol.value) - 1)
+		v.type == "var"
+	}
+}
+
+default _get_comprehension_value(comprehension) := set()
```
charlieegan3: https://www.openpolicyagent.org/projects/regal/rules/style/avoid-get-and-list-prefix

internal/lsp/semantictokens/semantictokens.go
charlieegan3: @anderseknert unsure if you have any feelings about where code to convert `N:N:N:N` into locations should live.

bundle/regal/lsp/semantictokens/semantictokens_test.rego
SeanLedford: Hmm yeah I honestly just printed what was returned in the case where the rule returned nothing, and made it the expected output. I don't think I have default values set for this rule though, is an empty set unexpected?

bundle/regal/lsp/semantictokens/vars/comprehensions/comprehensions.rego
```diff
@@ -0,0 +1,71 @@
...
+package regal.lsp.semantictokens.vars.comprehensions
+
+import data.regal.ast
+
+# METADATA
+# description: Extract comprehension variable declarations from array/set/object comprehensions
+result.declaration contains var if {
+	# regal ignore:prefer-some-in-iteration
+	comprehensions := ast.found.comprehensions[_]
+	some comprehension in comprehensions
```
charlieegan3: Hmm, so ast.found.comprehensions is an object. So it's like foobar here:

```rego
package play

foobar[ri] contains value if {
	some ri in [1, 2, 3]

	some value in [4, 5, 6]
}
```

and then we want the values from the obj, and don't care for the key, we can just do:

```rego
barfoo contains var if {
	some var in foobar
}
```

charlieegan3: https://play.openpolicyagent.org/p/g_NTAzYmYzMzc3ODAwZTMxNDhkZTk3ODM1MDBjMTVlYmJfXwBrRerG-w4OobPH4NcFyzNARGs

bundle/regal/lsp/semantictokens/vars/comprehensions/comprehensions.rego
```diff
@@ -0,0 +1,71 @@
...
+	comp_vars := result.declaration
+
+	output_vars := array.flatten([
+		_comprehension_key(comprehension),
+		_comprehension_value(comprehension),
+		_comprehension_body_vars(comprehension.value.body),
+	])
+	some var in output_vars
+	var.type == "var"
+	var.value in {v.value | some v in comp_vars}
```
charlieegan3:
```suggestion
	var.value in {v.value | some v in result.declaration}
```

bundle/regal/lsp/semantictokens/vars/comprehensions/comprehensions.rego
```diff
@@ -0,0 +1,71 @@
...
+	some var in comp_vars
+}
+
+# METADATA
+# description: Extract comprehension variable references in the output
+result.reference contains var if {
+	# regal ignore:prefer-some-in-iteration
+	comprehensions := ast.found.comprehensions[_]
+	some comprehension in comprehensions
+	comp_vars := result.declaration
```
charlieegan3:
```suggestion
```

bundle/regal/lsp/semantictokens/vars/every_expr/every_expr.rego
```diff
@@ -0,0 +1,39 @@
...
+# METADATA
+# description: Get the module from workspace
+module := data.workspace.parsed[input.params.textDocument.uri]
+
+# METADATA
+# description: Extract variable declarations from every keyword domains
+result.declaration contains var if {
+	some rule_index
+	declared_vars := ast.found.vars[rule_index]["every"]
+	some var in declared_vars
```
charlieegan3, line 20:
```suggestion
	var := ast.found.vars[_]["every"][_]
```

Writing rego like the olde times!

bundle/regal/lsp/semantictokens/vars/some_expr/some_expr.rego
```diff
@@ -0,0 +1,74 @@
...
+}
+
+# METADATA
+# description: Extract variable references in some keyword domains
+result.reference contains var if {
+	some rule_index, startpoint in _some_start_points
+	some context in {"somein", "some"}
+
+	# regal ignore:prefer-some-in-iteration
+	declared_vars := ast.found.vars[rule_index][context]
```
charlieegan3, line 34: Similar to above comments.

bundle/regal/ast/search.rego
anderseknert: There are other locations for sure! Will make a note to get back to this once merged.

bundle/regal/lsp/semantictokens/vars/comprehensions/comprehensions.rego
```diff
@@ -0,0 +1,70 @@
...
+
+# METADATA
+# description: Extract comprehension variable declarations from array/set/object comprehensions
+result.declaration contains var if {
+	# regal ignore:prefer-some-in-iteration
+	comprehensions := ast.found.comprehensions[_]
+	some comprehension in comprehensions
+
+	comp_vars := {v |
+		some term in comprehension.value.body
```
anderseknert: This should be named `expr` as that's what you'll get iterating over a body.

bundle/regal/lsp/semantictokens/vars/comprehensions/comprehensions.rego
```diff
@@ -0,0 +1,70 @@
...
+		term.terms.symbols
```
anderseknert: This is a [redundant-existence-check](https://www.openpolicyagent.org/projects/regal/rules/bugs/redundant-existence-check) as the same ref is on the line below as well, and evaluation will fail there if this undefined. Unfortunately that linter rule is not yet competent enough to catch this in nested context such as comprehensions. But you can safely remove this line.

bundle/regal/lsp/semantictokens/vars/comprehensions/comprehensions.rego
```diff
@@ -0,0 +1,70 @@
...
+		some symbol in term.terms.symbols
+		some v in array.slice(symbol.value, 1, count(symbol.value) - 1)
```
anderseknert: What's the last item in `symbol.value`? I.e. the one we discard here?

bundle/regal/lsp/semantictokens/vars/comprehensions/comprehensions.rego
anderseknert: Could this not be just `comprehension := ast.found.comprehensions[_][_]`

That should also allow removing `# regal ignore:prefer-some-in-iteration`

bundle/regal/lsp/semantictokens/vars/comprehensions/comprehensions.rego
```diff
@@ -0,0 +1,70 @@
...
+import data.regal.ast
+
+# METADATA
+# description: Extract comprehension variable declarations from array/set/object comprehensions
+result.declaration contains var if {
+	# regal ignore:prefer-some-in-iteration
+	comprehensions := ast.found.comprehensions[_]
+	some comprehension in comprehensions
+
+	comp_vars := {v |
```
anderseknert: Why are we building this intermediate `comp_vars` set only to then iterate over it? We could just do what's in the body here right in the top-level AFAICS.

bundle/regal/lsp/semantictokens/vars/comprehensions/comprehensions.rego
SeanLedford: Was seeing issues with my output when trying to adopt this format, doing `some comprehension in ast.found.comprehensions`. I understand this conceptually though, so I'm curious what the difference is here.

bundle/regal/lsp/semantictokens/vars/comprehensions/comprehensions_test.rego
```diff
@@ -0,0 +1,76 @@
...
+			{"location": "4:10:4:11", "type": "var", "value": "x"},
+			{"location": "4:7:4:8", "type": "var", "value": "i"},
+		},
+		"references": {
+			{"location": "3:26:3:27", "type": "var", "value": "x"},
+			{"location": "5:2:5:3", "type": "var", "value": "i"},
+		},
+	}}
+	array_comp_tokens := comprehensions.result with input as {"params": {"textDocument": {"uri": "file://p.rego"}}}
+		with data.workspace.parsed["file://p.rego"] as regal.parse_module("p.rego", policy_one)
```
anderseknert, line 23: You could use the `ast.policy` helper here and in most locations in place of `regal.parse_module` if you want. Then you don't need to include `package regal.woo` in your example as a package line is prepended to your snippet. But that's optional, and might mess up your locations.

bundle/regal/lsp/semantictokens/vars/every_expr/every_expr.rego
anderseknert: No preference on style for me, but see my comment above about rule_index

bundle/regal/lsp/semantictokens/vars/function_args/function_args.rego
```diff
@@ -0,0 +1,56 @@
...
+
+	rule.head.args
+
+	arg_names := ast.function_arg_names(rule)
+
+	walk(rule.body, [_, expr])
+
+	expr.terms[0].type == "ref"
+
+	some arg in array.slice(expr.terms, 1, count(expr.terms))
```
anderseknert: You could use the usual Regal cheap trick here and just do `array.slice(expr.terms, 1, 100)` to avoid the count call. If someone managed to somehow cram a 100 terms into a single expression (if that's even possible), Regal can't save them. They'd need a priest or something.

bundle/regal/lsp/semantictokens/vars/imports/imports.rego
```diff
@@ -0,0 +1,20 @@
...
+module := data.workspace.parsed[input.params.textDocument.uri]
+
+# METADATA
+# description: Extract import tokens - return only last term of the path
+result contains last_term if {
+	some import_statement in module.imports
+	import_path := import_statement.path.value
+
+	last_term := regal.last(import_path)
+}
```
anderseknert, line 21: We should at least add a note in this package about us not handling `alias`ed identifiers yet.

bundle/regal/lsp/semantictokens/vars/packages/packages.rego
```diff
@@ -0,0 +1,19 @@
...
+#   - input.params: schema.regal.lsp.semantictokens
+package regal.lsp.semantictokens.vars.packages
+
+# METADATA
+# description: Get the module from workspace
+module := data.workspace.parsed[input.params.textDocument.uri]
+
+# METADATA
+# description: Extract package tokens - return full package path
+result contains last_term if {
```
anderseknert:
```rego
result contains regal.last(module.package.path)
```

internal/lsp/semantictokens/semantictokens.go
```diff
@@ -30,166 +31,189 @@ type Token struct {
...
+// Represents location data from the AST
+type ASTLocation struct {
+	Location LocationInfo `json:"location"`
+}
+
+type LocationInfo struct {
+	Line        uint
+	StartColumn uint
+	EndColumn   uint
+	Length      uint
```
anderseknert: I like the use of uint over int here! But uint32 should suffice for our purposes

internal/lsp/semantictokens/semantictokens.go
anderseknert: We have plenty of helpers to do this in Rego already. The reason we do it on this side is that it's more performant, or something else?

I say fine to leave it here for now, and we can consider whether this should be in Rego or Go as a next step.

internal/lsp/semantictokens/semantictokens.go
```diff
@@ -30,166 +31,189 @@ type Token struct {
...
 }
 
+func extractTokens(astLoc ASTLocation, tokenType uint, modifiers uint) (Token, error) {
+	return Token{
+		Line:      astLoc.Location.Line - 1,
+		Col:       astLoc.Location.StartColumn - 1,
+		Length:    astLoc.Location.Length,
+		Type:      tokenType,
+		Modifiers: modifiers,
+	}, nil
```
anderseknert, line 214: Looking at this data structure has me more convinced we should move pretty much all of this over to Rego eventually. We have all the data that we need there already, and should be able to return tokens in this shape from there without intermediate formats and collections. But let's leave it be like this for now. Whatever works is good for the first implementation. 

The only function I think must remain in Go even later is likely the one that turns tokens into deltas. That would be hard to do in Rego, I think.

bundle/regal/lsp/semantictokens/vars/every_expr/every_expr.rego
```diff
@@ -0,0 +1,39 @@
...
+
+import data.regal.ast
+
+# METADATA
+# description: Get the module from workspace
+module := data.workspace.parsed[input.params.textDocument.uri]
+
+# METADATA
+# description: Extract variable declarations from every keyword domains
+result.declaration contains var if {
```
anderseknert, line 17: It looks to me like on the receiving side (i.e. in Go) we only ever care about the location of tokens. Not the type, not the value. If that is so, why don't we return locations only? i.e. `var.location` and so on.

bundle/regal/lsp/semantictokens/vars/function_args/function_args.rego
anderseknert: I haven't looked into it. But one thing I'm not seeing is us looking for var refs in the rule.head as well.  meaning we'll miss things like

```
add(a, b) := a + b
```

Or no?

Seeing as we do `expr.terms[0].type == "ref"` below, can we not replace the `walk` and look/walk for vars in ast.found.refs ? That'd give us vars from both rule refs, rule heads and bodies.

bundle/regal/lsp/semantictokens/vars/comprehensions/comprehensions.rego
anderseknert: The compiler explorer should be helpful here :) Let's have a look together next week,

bundle/regal/lsp/semantictokens/vars/comprehensions/comprehensions.rego
SeanLedford: Ah good catch, addressed.

bundle/regal/lsp/semantictokens/vars/comprehensions/comprehensions.rego
SeanLedford: This is just me converting from the length to the index here. But I am discarding the first item in `symbol.value`, as the variables start from there in the ast.

bundle/regal/lsp/semantictokens/vars/comprehensions/comprehensions.rego
SeanLedford: Ah yeah, I believe this was leftover from me messing around with the `mixed_iteration` violation

bundle/regal/lsp/semantictokens/vars/comprehensions/comprehensions.rego
SeanLedford: Yeah I was able to get that syntax working, but it would be great to take a look still.

SeanLedford:
> I haven't looked into it. But one thing I'm not seeing is us looking for var refs in the rule.head as well. meaning we'll miss things like

Yes good catch, I'm instead using ast.found.calls now, which does a similar walk, just not limited to the scope of the rule bodies.

bundle/regal/lsp/semantictokens/vars/every_expr/every_expr.rego
SeanLedford: Yep the Go side only uses the location, but I thought it was neat to use my declaration finding rules in my reference rules, instead of duplicating the code within that rule. For that I would need to match the value, so I left the object as is for now. Happy to hear your opinion on this though.

bundle/regal/lsp/semantictokens/vars/function_args/function_args.rego
SeanLedford: Haha very good point, will add this.

bundle/regal/lsp/semantictokens/vars/comprehensions/comprehensions.rego
SeanLedford: Did not directly state this in my original comment so just clarifying: addressed this in my latest commit

PR #1886, opened 2026-02-21
Fix false positive in use-some-for-output-vars

### LINKED ISSUE #1876: False positive in `use-some-for-output-vars` rule
This is reported as not using `some` although clearly it is:
```rego
aliased_imports contains input.imports[x] if {
    some x
    input.imports[x].alias
}
```

anderseknert:
Fixes #1876

### DIFF
--- a/bundle/regal/rules/idiomatic/use-some-for-output-vars/use_some_for_output_vars.rego
+++ b/bundle/regal/rules/idiomatic/use-some-for-output-vars/use_some_for_output_vars.rego
@@ -13,6 +13,7 @@ report contains violation if {
 	not startswith(term.value, "$")
 	not term.value in ast.imported_identifiers
 	not term.value in ast.rule_names
+	not true in {true | some v in ast.found.vars[rule_index].some; v.value == term.value}
 
 	rule := input.rules[to_number(rule_index)]
 
@@ -20,7 +21,7 @@ report contains violation if {
 
 	walk(rule, [path, term.location])
 
-	not _var_in_ref_head_declared(path, rule_index, term.value)
+	not _var_in_ref_head_declared(path[0], rule_index, term.value)
 	not _var_in_comprehension_body(path, term.value, rule)
 
 	violation := result.fail(rego.metadata.chain(), result.location(term))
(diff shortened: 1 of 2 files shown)


charlieegan3 (approved):

bundle/regal/rules/idiomatic/use-some-for-output-vars/use_some_for_output_vars.rego
```diff
@@ -13,14 +13,15 @@ report contains violation if {
...
+	not true in {true | some v in ast.found.vars[rule_index].some; v.value == term.value}
 
 	rule := input.rules[to_number(rule_index)]
 
 	not ast.is_in_local_scope(rule, term.location, term.value)
 
 	walk(rule, [path, term.location])
 
-	not _var_in_ref_head_declared(path, rule_index, term.value)
+	not _var_in_ref_head_declared(path[0], rule_index, term.value)
```
charlieegan3, line 24: It's not immediately clear to me why this works/is needed

bundle/regal/rules/idiomatic/use-some-for-output-vars/use_some_for_output_vars.rego
```diff
@@ -13,14 +13,15 @@ report contains violation if {
 	not startswith(term.value, "$")
 	not term.value in ast.imported_identifiers
 	not term.value in ast.rule_names
+	not true in {true | some v in ast.found.vars[rule_index].some; v.value == term.value}
```
charlieegan3, line 16: I understand this bit.

bundle/regal/rules/idiomatic/use-some-for-output-vars/use_some_for_output_vars.rego
anderseknert: After `_var_in_ref_head_declared` got simplified below, it now only ever references the first path component, so Regal suggested we'd [narrow](https://www.openpolicyagent.org/projects/regal/rules/custom/narrow-argument) down what we passed.

charlieegan3: Ahh very nice!

PR #1888, opened 2026-02-23
lsp: Implement LSP opaTestProvider feature in language server

charlieegan3:
Please see https://github.com/open-policy-agent/vscode-opa/pull/408 for client implementation and video.

~Draft until we get https://github.com/open-policy-agent/opa/pull/8362 in as it's manually referenced atm.~


docs/readme-sections/badges.md
```diff
@@ -1,6 +1,6 @@
 <!-- markdownlint-disable MD041 -->
 
 [![Build Status](https://github.com/open-policy-agent/regal/workflows/Build/badge.svg)](https://github.com/open-policy-agent/regal/actions)
-![OPA v1.13.1](https://www.openpolicyagent.org/badge/v1.13.1)
+![OPA v1.13.1-0.20260224105757-9b2143cdf99c](https://www.openpolicyagent.org/badge/v1.13.1-0.20260224105757-9b2143cdf99c)
```
srenatus, line 4: What do we need? Ah, the runtime bits. 💡

srenatus (approved):
LGTM. Parameterized tests don't happen to "just work", do they? 🤔

charlieegan3:
> Parameterized tests don't happen to "just work", do they?

Afraid not yet, I need to dig a little deeper to understand how to reliably detect these. 

There is also a bit of an issue where each test is run one after the other rather than all at once. I need to understand how to have 'test run all' return acceptable results to update the single test rule state.

charlieegan3:
That's the reason there is no test entire package / file, since the client does that.

PR #1891, opened 2026-02-25
Fix broken input.json completion provider

### LINKED ISSUE #1890: No completions from input.json
Doing some work on `main` and not getting any completion suggestions from my `input.json` file.

I have rebuilt Regal, but the issue remains. Other completion providers seem to work as expected. I hope I'm holding it wrong somehow, but it seems like something has broken this provider. If that's the case, we need to have it fixed before the next release.

anderseknert:
Not sure when it happened, but the struct -> ast.Value transform failed to take into account struct field pointing to existing ast.Value values.

Fixes #1890

### DIFF
--- a/pkg/roast/rast/rast.go
+++ b/pkg/roast/rast/rast.go
@@ -206,16 +206,21 @@ func StructToValue(input any) ast.Value {
 
 		value := v.Field(i)
 
-		if strings.Contains(tag, ",") {
-			parts := strings.Split(tag, ",")
-			tag = parts[0]
+		if comma := strings.Index(tag, ","); comma != -1 {
+			rest := tag[comma+1:]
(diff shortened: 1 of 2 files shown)


pkg/roast/rast/rast.go
```diff
@@ -206,16 +206,21 @@ func StructToValue(input any) ast.Value {
 
 		value := v.Field(i)
 
-		if strings.Contains(tag, ",") {
-			parts := strings.Split(tag, ",")
-			tag = parts[0]
+		if comma := strings.Index(tag, ","); comma != -1 {
```
srenatus, line 209: Looks like a nice place for a strings.Cut 😉

srenatus (approved):

PR #1892, opened 2026-02-25
Various AST transform improvements

anderseknert:
Reduce allocations, fix some TODOs, and so on. Not much to see, just me doing anything but work on what I'm supposed to work on.

### DIFF
--- a/internal/roast/transforms/module/module.go
+++ b/internal/roast/transforms/module/module.go
@@ -1,6 +1,7 @@
 package module
 
 import (
+	"bytes"
 	"encoding/base64"
 
 	"github.com/open-policy-agent/opa/v1/ast"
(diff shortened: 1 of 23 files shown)


internal/roast/transforms/module/module.go
```diff
@@ -44,7 +51,11 @@ func ToValue(mod *ast.Module) (ast.Value, error) {
 		comments := make([]*ast.Term, len(mod.Comments))
 
 		for i, comment := range mod.Comments {
-			encoded := base64.StdEncoding.EncodeToString(comment.Text)
+			encoded := "IE1FVEFEQVRB" // " METADATA"
```
srenatus, line 54: 😄

internal/roast/transforms/module/module.go
```diff
@@ -10,10 +11,16 @@ import (
...
+var metadataBytes = []byte(" METADATA")
+
 // ToValue converts an AST module to RoAST value representation.
 // This is much more efficient than using a JSON encode/decode round trip.
 func ToValue(mod *ast.Module) (ast.Value, error) {
-	value := ast.NewObject()
+	value := ast.NewObjectWithCapacity(1 +
+		min(1, len(mod.Imports)) +
+		min(1, len(mod.Rules)) +
+		min(1, len(mod.Comments)),
```
srenatus, line 22: I see what you did there... OK... Perhaps this would warrant a small (to-be-inlined) helper?

```go
func selector[T ~[]](s T) int { // 1 or 0
  return min(1, len(T))
}
```
(untested)

internal/roast/transforms/module/module.go
anderseknert: Hah! Like 500+ of them in Regal alone :P

internal/roast/transforms/module/module.go
anderseknert: I'm already well over my time budget, but noted for the next round 👍

PR #1897, opened 2026-02-25
Improved redundant-existence-check

### LINKED ISSUE #1805: Make redundant-existence-check care less about where a check happens
This is currently considered a [redundant-existence-check](https://www.openpolicyagent.org/projects/regal/rules/bugs/redundant-existence-check):
```rego
f(obj) if {
    obj.type
    obj.type == "object"	
}
```
But this isn't:
```rego
f(obj) if {
    obj.type == "object"	
    obj.type
}
```
Whether the intent was an existence check or not, the latter should also be considered a violation of this rule.

I actually found a case of this in our own code [here](https://github.com/open-policy-agent/regal/blob/dec83ea62b334de6b7b521df552d34534b1b960c/bundle/regal/result/result.rego#L187) 😅

anderseknert:
By looking in both directions. Still not exhaustive, but that's way too expensive.. and most redundant checks are adjacent to the place of use.

Fixes #1805

### DIFF
--- a/bundle/regal/rules/bugs/redundant-existence-check/redundant_existence_check.rego
+++ b/bundle/regal/rules/bugs/redundant-existence-check/redundant_existence_check.rego
@@ -8,18 +8,13 @@ import data.regal.result
 # METADATA
 # description: check rule bodies for redundant existence checks
 report contains violation if {
-	some rule_index, rule in input.rules
-	some expr_index, expr in _exprs[rule_index]
+	some rule_index, expr_index
+	expr := _static_ref_exprs[rule_index][expr_index]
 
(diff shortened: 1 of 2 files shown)


charlieegan3:
I guess the look ahead behind could be configurable.

anderseknert:
@charlieegan3 yes, but what if you find:

```rego
foo.bar
super_expensive_call()
foo.bar.baz == 1
```

Is the foo.bar check redundant? Logically yes it is, as the rule will evaluate the same without it. But it’s also a reasonable way to guard the more expensive computation below, making it not redundant. Like a poor man’s indexer. That’s why being exhaustive here is next to impossible 😅

PR #1909, opened 2026-03-04
lsp: Wait for client to send initialized at boot

charlieegan3:
Now the server will wait for the client to send the initialized (empty) message before starting work. Previously we got away with sending diagnostics but now we have a custom test location handler, we need to ensure that we wait for that to be registered before sending test location notifications.

So we have moved the load workspace contents to the post initialized state, and added a gate to block the test locations worker from running if the client does not support it.

### DIFF
--- a/cmd/languageserver.go
+++ b/cmd/languageserver.go
@@ -42,10 +42,16 @@ func init() {
 				verbose = true
 			}
 
+			logLevel := log.LevelMessage
+			if verbose {
+				logLevel = log.LevelDebug
+			}
+
 			opts := &lsp.LanguageServerOptions{
-				Logger:       log.NewLogger(log.LevelMessage, os.Stderr),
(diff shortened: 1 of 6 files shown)


charlieegan3:
I think this is what we should have done all the time and it is what others like gopls do. That said, it's not super clear, and is phrased more in terms of client responsibility than server.

> The initialized notification is sent from the client to the server after the client received the result of the initialize request but before the client is sending any other request or notification to the server. The server can use the initialized notification, for example, to dynamically register capabilities. The initialized notification may only be sent once.

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialized

charlieegan3:
It so happened that clients were able to accept diagnostics before the initialized message was sent back, so yes, this is a little slower.

anderseknert:
Thanks @charlieegan3 👍

PR #1911, opened 2026-03-07
Allow naming-convention rule to take list of names (not only a pattern)

### LINKED ISSUE #1493: Establish terminology for commonly used AST elements
While building Regal in the span of the last 2 years, we haven't given much thought about how we name things even in the AST. 15K+ lines of Rego dealing with that AST later, and this is slowly starting to be an annoyance for me. Like how we use `ref` in to refer to a whole bunch of (by all means related) things, or `path`, `term`, `value` etc. Or how `location` sometimes means the compact string format from roast, sometimes the resulting object you get from parsing that, and something else entirely in the context of LSP policies.

Most often it's pretty easy to understand what something actually means by looking at the code, but consistent naming would speed up that process a lot, and likely make the project more approachable to others.

I think adding a document under `docs` to explain the terminology we use would be a good first step, and then to make an effort to ensure we follow that consistently.

(Ideally, we'd be able to have linter rules to enforce at least parts of that ... but any such rule would very much be a best effort that probably wouldn't amount to anything near good coverage)

anderseknert:
While the title of this PR is what we'll want in the release notes, this is 95% an internal change meant to improve Regal for ourselves. That a custom rule got better is mostly a happy coincidence :)

Since we started working on Regal, we have learnt a lot about both OPA, the AST and of course our own domains — static analysis, linting and language servers. The terminology used in the project has grown organically along with that, and with time we've found a shared vocabulary for the project. There's however been no lack of code written before we knew better, and as this PR highlights, there have been many places where names have been picked rather arbitrarily.

Additionally, the lack of a documented vocabulary meant it wasn't accessible outside of our own internal know-how. This change aims to fix that by:

1. Documenting the terminlogy and vocabulary of Regal and our domain
2. Provide the means to help check how well we follow our stated naming conventions
3. Update our Rego code to comply with these naming conventions

While having this goes a long way to unify our terminology, there are things that won't be caught by a simple "allow list" like this. As an example, our use of the name `ref` sometimes mean the "outer" type (the ref term), and sometimes it refers to only the the term slice of the ref (which should rather be called `terms`). That's out of scope for this PR, but could be something to look into later.

As noted in the change, we will not enforce anything related to naming conventions, but occasionally check this and make fixes accordingly (and without ever getting in the way of anyone contributing to the project).

Fixes #1493

### DIFF
--- a/.regal/config.yaml
+++ b/.regal/config.yaml
@@ -16,6 +16,101 @@ rules:
       ignore:
         files:
           - "*_test.rego"
+    naming-convention:
+      # Regal naming conventions configuration. Note that this is disabled by default, and
+      # is meant only to be run every once in a while to ensure we stick to our conventions.
+      # These conventions are not static, and the correct "fix" to a naming violation may
+      # very well be to update this configuration rather than the code.
+      #
+      # To check compliance:
+      # regal lint --enable naming-convention bundle
+      #
+      # See naming.md for rationale and examples of these conventions.
+      # NOTE: that this check is not automated, but intended to be used occasionally.
+      level: ignore
+      conventions:
+        - pattern: '\b[a-z_]{5,}\b'
+          names:
+            # indices
+            - i
+            - j
+            - k
+            # count / length
+            - n
+            # rego types
+            - arr
+            - obj
+            - set
+            - str
+            - num
+            # imaginary 'types'
+            - coll
+            - seq
+            - item
+            # ast types
+            - pkg
+            - imp
+            - ref
+            - refs
+            - var
+            - vars
+            - fun
+            - arg
+            - args
+            - term
+            - terms
+            - expr
+            - exprs
+            - rule
+            - head
+            - body
+            - call
+            - comp
+            - node
+            # ast attributes and related
+            - key
+            - keys
+            - val
+            - vals
+            - type
+            # common
+            - file
+            - text
+            - line
+            - rows
+            - cols
+            - loc
+            - end
+            - col
+            - row
+            - agg
+            - aggs
+            - cfg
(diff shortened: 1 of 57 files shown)


.regal/config.yaml
```diff
@@ -16,6 +16,101 @@ rules:
...
+            # count / length
+            - n
+            # rego types
+            - arr
+            - obj
+            - set
+            - str
+            - num
+            # imaginary 'types'
+            - coll
```
charlieegan3, line 47: Collection?

anderseknert: Yep! Basically the “any” type of sets, arrays, objects

bundle/regal/rules/imports/ignored-import/ignored_import.rego
```diff
@@ -18,11 +18,11 @@ report contains violation if {
...
 	ref.value[0].value in {"data", "input"}
 
-	most_specific_match := regal.last(sort([ip |
-		ref_path := [p.value | some p in ref.value]
+	most_specific_match := regal.last(sort([import_path |
+		ref_path := [term.value | some term in ref.value]
 
-		some ip in _import_paths
-		array.slice(ref_path, 0, count(ip)) == ip
+		some import_path in _import_paths
```
charlieegan3, line 24: Nice, I like these improvements a lot!

docs/naming.md
```diff
@@ -0,0 +1,140 @@
+# Naming Convention
```
charlieegan3, line 1: if we had a templating language, we might be able to generate parts of this file...

anderseknert: If only!!

PR #1913, opened 2026-03-10
lsp/testing: Nested test implementation 

charlieegan3:
This supports the following nested test view in the client:

Screenshot 2026-03-10 at 10 15 46

### DIFF
--- a/internal/lsp/eval.go
+++ b/internal/lsp/eval.go
@@ -47,7 +47,7 @@ type PrintHook struct {
 func (l *LanguageServer) Eval(
 	ctx context.Context, query string, input map[string]any, printHook print.Hook,
 ) (rego.ResultSet, error) {
-	regoArgs := prepareRegoArgs(ast.MustParseBody(query), l.assembleEvalBundles(), printHook, l.getLoadedConfig())
+	regoArgs := prepareRegoArgs(ast.MustParseBody(query), l.assembleBundles(), printHook, l.getLoadedConfig())
 
 	// TODO: Let's try to avoid preparing on each eval, but only when the contents
 	// of the workspace modules change, and before the user requests an eval.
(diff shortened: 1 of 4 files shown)


internal/lsp/eval.go
```diff
@@ -93,13 +93,13 @@ func (l *LanguageServer) EvalInWorkspace(ctx context.Context, query string, inpu
 
 func prepareRegoArgs(
 	query ast.Body,
-	bundles map[string]bundle.Bundle,
+	bundles map[string]*bundle.Bundle,
 	printHook print.Hook,
 	cfg *config.Config,
 ) []func(*rego.Rego) {
 	bundleArgs := make([]func(*rego.Rego), 0, len(bundles))
-	for key, b := range bundles { //nolint:gocritic // expensive copy, but I don't think we can avoid it
```
charlieegan3, line 101: Hmm, is there something I have missed here by using pointers to bundles? I just wanted to be able to do `SetBundles(l.assembleBundles())` in the test runner to copy what we did in eval.

anderseknert: Nothing wrong. IIRC it's just that the various OPA APIs we use return `Bundle` rather than `*Bundle`, so it'll be copied either way. Doesn't hurt to copy less though.

internal/lsp/server_testing.go
```diff
@@ -19,14 +22,41 @@ func (l *LanguageServer) handleRunTests(
...
+	txn, err := testStore.NewTransaction(ctx, storage.WriteParams)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create transaction: %w", err)
 	}
 
-	defer l.regoStore.Abort(ctx, txn)
+	defer testStore.Abort(ctx, txn)
+
+	// TODO: make the loading of config and regal built-ins only happen
+	// in Regal repo
```
anderseknert, line 39: repo?

PR #1918, opened 2026-03-25
Fail with error when --profile used without JSON format

### LINKED ISSUE #1542: Have `regal lint --profile` show an error message if not `--format json` is set
Since supporting other formats here would have meant copying a lot of code from OPA, we opted to only support the JSON format for profiling. However, users have no way of knowing this unless we tell them. So whenever the `--profile` flag is used and the format is not `--json`, we should fail with an error message telling them to use `--format json` to see profiling data.

mvanhorn:
## Summary

When `--profile` is used without `--format json`, regal now returns an error instead of silently producing no profiling data.

## Why this matters

Profiling only works with JSON output. Without this check, users get no feedback when they forget `--format json` - the command succeeds but profiling data is missing. The error message tells them exactly what to do.

## Changes

- `cmd/lint.go`: Added validation at the start of `lint()` that checks `params.profile && params.format != formatJSON` and returns a descriptive error.

## Testing

- `go build ./...` passes
- Manual verification: `regal lint --profile .` now returns the error message instead of silently dropping profiling data

Fixes #1542

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/cmd/lint.go
+++ b/cmd/lint.go
@@ -172,8 +172,9 @@ func init() {
 
 func lint(args []string, params *lintParams) (result report.Report, err error) {
 	if params.profile && params.format != formatJSON {
-		return report.Report{}, fmt.Errorf("--profile requires --format json to display profiling data")
+		return report.Report{}, errors.New("--profile requires --format json to display profiling data")
 	}
+
 	ctx, cancel := getLinterContext(params.lintAndFixParams)
 	defer cancel()
 


mvanhorn:
Fair point - this was definitely not Claude's finest hour. DCO and lint fixed in 1baf26b.

anderseknert:
Thanks 👍

mvanhorn:
Thanks for the review and merge!

PR #1929, opened 2026-04-01
[feature] Add Popup Prompt for Creating input.json If Not Present When Evaluate is Run

SeanLedford:
Uses lsp capability to provide a popup prompt to the user when they run evaluate from the IDE with no input.json present. The user can close the prompt, ignore it for as long as the language server runs, or have an empty input.json file created automatically. 

image

A follow up PR is in progress containing testing and generation of an input.json skeleton based on what fields from the input are used!

### DIFF
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -165,6 +165,9 @@ type LanguageServer struct {
 
 	workspaceRootURI         string
 	workspaceDiagnosticsPoll time.Duration
+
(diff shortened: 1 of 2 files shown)


anderseknert:
> generation of an input.json skeleton based on what fields from the input are used!

Great! See the discussion [here](https://github.com/open-policy-agent/regal/issues/1928#issuecomment-4166102470) for some code that does something similar, and that you could use at least parts from.

PR #1932, opened 2026-04-02
Rule: superflous-object-get

### LINKED ISSUE #1926: Rule: superfluous-object-get
Reviewing some obviously generated Rego recently, I was surprised to see it made use of the `object.get` built-in function for pretty much all comparisons. I was less surprised to see it did so incorrectly, with most of the calls doing something like this:

```rego
object.get(input, ["foo", "bar"], "") == "baz"
```

This of course would be better expressed without `object.get`:

```rego
input.foo.bar == "baz"
```

Whether this is common in generated Rego or not, I have no idea. But I could easily see a human author make the same mistake based on not fully understanding undefined, so let's add a rule to identify this. 

(Trivia: we already have rules that include "redundant", "pointless" and "unnecessary" in their names. This would be the first to use "superfluous"!)

anderseknert:
Let's help our deceiving robot friends write better Rego!

Fixes #1926

### DIFF
--- a/bundle/regal/config/provided/data.yaml
+++ b/bundle/regal/config/provided/data.yaml
@@ -101,6 +101,8 @@ rules:
       level: error
     single-item-in:
       level: error
+    superfluous-object-get:
+      level: error
     use-array-flatten:
       level: error
     use-contains:
--- a/bundle/regal/rules/idiomatic/superfluous-object-get/superfluous_object_get.rego
+++ b/bundle/regal/rules/idiomatic/superfluous-object-get/superfluous_object_get.rego
@@ -1,0 +1,58 @@
+# METADATA
+# description: Superfluous `object.get` call
+package regal.rules.idiomatic["superfluous-object-get"]
+
+import data.regal.ast
+import data.regal.result
+
+# object.get(a, "b", "c") == "violation"
+report contains violation if {
+	# check the most unlikely case first in order to fail fast
+	some rule_index, expr
+	ast.found.expressions[rule_index][expr].terms[1].type == "call"
+
+	terms := ast.found.expressions[rule_index][expr].terms
+
+	terms[1].value[0].value[0].value == "object"
+	terms[1].value[0].value[1].value == "get"
+
+	terms[0].type == "ref"
+	terms[0].value[0].type == "var"
+	terms[0].value[0].value == "equal"
+
+	def_value := terms[1].value[3]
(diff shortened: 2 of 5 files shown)


bundle/regal/config/provided/data.yaml
```diff
@@ -101,6 +101,8 @@ rules:
       level: error
     single-item-in:
       level: error
+    superfluous-object-get:
```
charlieegan3, line 104: In other rules, we tend to use redundant to mean much the same thing. perhaps that would feel more familiar to existing users?

bundle/regal/rules/idiomatic/superfluous-object-get/superfluous_object_get.rego
```diff
@@ -0,0 +1,58 @@
+# METADATA
+# description: Superfluous `object.get` call
+package regal.rules.idiomatic["superfluous-object-get"]
+
+import data.regal.ast
+import data.regal.result
+
+# object.get(a, "b", "c") == "violation"
+report contains violation if {
+	# check the most unlikely case first in order to fail fast
```
SeanLedford, line 10: You have comment here stating you're checking the most unlikely case first, is this referring to the checking that the expression is a call, or the ordering of the the left-hand versus right-hand side? If it's the latter I'd imagine you don't want the comments here in both cases

bundle/regal/config/provided/data.yaml
anderseknert: Just for posterity: this was mentioned in the [original issue](https://github.com/open-policy-agent/regal/issues/1926#issue-4173661940). We can afford to be a little quirky as long as it's not annoying. And I think in this case, it's not.

bundle/regal/rules/idiomatic/superfluous-object-get/superfluous_object_get.rego
anderseknert: You're right, I could have been clearer. It's really all these lines:

```
ast.found.expressions[rule_index][expr].terms[1].type == "call"

terms := ast.found.expressions[rule_index][expr].terms

terms[1].value[0].value[0].value == "object"
terms[1].value[0].value[1].value == "get"
```

i.e.  checking that it's a call to object.get specifically, which is of course typically not the case 🙂

I'll update the comments in a later change to better describe that. Thanks!

PR #1934, opened 2026-04-02
[feature] Create input.json Skeleton if User Would Like When One Doesn't Exist

SeanLedford:
Recent PR #1929  introduced a prompt to the user that allowed them create an empty input.json file if one didn't exist. This PR creates a proper input.json file based on its references.

Also adds a fix to stop the command worker thread from being blocked if the user doesn't respond to the prompt, and testing for the creation of the input file.

Addresses the [following issue](https://github.com/open-policy-agent/vscode-opa/issues/425) that was raised in [vscode-opa](https://github.com/open-policy-agent/vscode-opa)

### DIFF
--- a/internal/lsp/eval.go
+++ b/internal/lsp/eval.go
@@ -204,24 +204,6 @@ func (h PrintHook) Print(ctx print.Context, msg string) error {
 	return nil
 }
 
-func ruleHasInputRefs(rule *ast.Rule) bool {
-	if rule == nil {
-		return false
-	}
-
-	found := false
-
-	ast.WalkRefs(rule, func(ref ast.Ref) bool {
-		if len(ref) >= 2 && ref[0].Equal(ast.InputRootDocument) {
-			found = true
-		}
-
-		return found
-	})
-
-	return found
-}
-
 func inputSkeletonFromRule(rule *ast.Rule, compiler *ast.Compiler) map[string]any {
 	root := map[string]any{}
 
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -2543,6 +2543,101 @@ func (l *LanguageServer) handleExplorerCommand(ctx context.Context, params types
 	return nil
 }
 
+func (l *LanguageServer) handleInputSkeletonPrompt(
+	ctx context.Context,
+	target, ruleName string,
+	row int,
+) (bool, error) {
+	compiler := compile.NewCompilerWithRegalBuiltins()
+	compiler.Compile(l.cache.GetAllModules())
+
+	if compiler.Failed() {
+		l.log.Message("failed to compile workspace modules for input skeleton: %v", compiler.Errors)
+	}
+
+	// Using the compiled modules to parse the rules. The dependencies package used in inputSkeletonFromRule
+	// relies on compiled modules to resolve transitive dependencies.
+	var compiledRule *ast.Rule
+
+	if compiledModule, ok := compiler.Modules[target]; ok {
+		for _, rule := range compiledModule.Rules {
+			if rule.Head.Name.String() == ruleName && rule.Location.Row == row {
+				compiledRule = rule
+
+				break
+			}
+		}
+	}
+
+	if compiledRule == nil {
+		return false, nil
+	}
+
+	skeleton := inputSkeletonFromRule(compiledRule, compiler)
+	if len(skeleton) == 0 {
+		return false, nil
+	}
+
+	var action types.MessageActionItem
+
+	showMsgCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	defer cancel()
+
+	if err := l.conn.Call(showMsgCtx, "window/showMessageRequest", types.ShowMessageRequestParams{
+		Type: 3, // info
+		Message: "No input.json/yaml file was found. " +
+			"This file is used to provide input data for rule evaluation. " +
+			"Would you like to create one?",
+		Actions: []types.MessageActionItem{
+			{Title: "Yes"},
+			{Title: "No"},
+			{Title: "Ignore"},
+		},
+	}, &action); err != nil {
+		return false, fmt.Errorf("window/showMessageRequest failed: %w", err)
+	}
+
+	switch action.Title {
+	case "Yes":
+		data, err := json.MarshalIndent(skeleton, "", "  ")
+		if err != nil {
+			return false, fmt.Errorf("failed to marshal input skeleton: %w", err)
+		}
+
+		inputFile := filepath.Join(l.workspacePath(), "input.json")
+		if err = os.WriteFile(inputFile, append(data, '\n'), 0o600); err != nil {
+			return false, fmt.Errorf("failed to create input.json: %w", err)
+		}
+
+		var openAction types.MessageActionItem
+		if err = l.conn.Call(ctx, "window/showMessageRequest", types.ShowMessageRequestParams{
+			Type:    3,
+			Message: "input.json created successfully! Running Evaluate will now pull from this file.",
+			Actions: []types.MessageActionItem{{Title: "Open"}},
+		}, &openAction); err != nil {
+			return true, fmt.Errorf("window/showMessageRequest failed: %w", err)
+		}
+
+		if openAction.Title == "Open" {
+			takeFocus := false
+
+			var showResult types.ShowDocumentResult
+			if err = l.conn.Call(ctx, "window/showDocument", types.ShowDocumentParams{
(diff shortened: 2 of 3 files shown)


internal/lsp/eval.go
```diff
@@ -202,3 +202,37 @@ func (h PrintHook) Print(ctx print.Context, msg string) error {
...
+			}
+			// If the object exists, make it the starting point for the next check
+			if child, ok := node[key].(map[string]any); ok {
+				node = child
+			}
+		}
+
+		leaf := strings.Trim(ref[len(ref)-1].Value.String(), `"`)
+		if _, ok := node[leaf]; !ok {
+			node[leaf] = "EXAMPLE"
```
anderseknert: Let's lowercase this :) And perhaps use "changeme" instead. What's more complicated though is how we don't have any type information here. Perhaps "changeme" is good enough for now, but input.foo.bar might just as well point to an integer or boolean value, and later we might want to try and extract type information from the compiler, or user-provided JSON schemas if any are found.

internal/lsp/server.go
anderseknert: Perhaps "to your rules" could be replaced with "for evaluation" as it's more direct?

internal/lsp/server_command_test.go
```diff
@@ -237,3 +238,92 @@ allow if {
...
+		},
+	}); err != nil {
+		t.Fatalf("failed to send didOpen: %s", err)
+	}
+
+	timeout := time.NewTimer(determineTimeout())
+	defer timeout.Stop()
+
+	// Needed to add a polling loop to wait for the module to be parsed before running the eval command.
+	// Possibly a better way to do this, but wasn't sure.
```
anderseknert: I suppose you could 
- ls.cache.SetModule
- ls.cache.SetFileContents

beforehand

internal/lsp/server_command_test.go
```diff
@@ -237,3 +238,92 @@ allow if {
...
+	select {
+	case <-inputJSONCreated:
+		for {
+			if _, err := os.Stat(filepath.Join(tempDir, "input.json")); err == nil {
+				contents, err := os.ReadFile(filepath.Join(tempDir, "input.json"))
+				if err != nil {
+					t.Fatalf("failed to read input.json: %s", err)
+				}
+
+				must.Equal(t, "{\n  \"foo\": \"EXAMPLE\"\n}\n", string(contents))
```
anderseknert: It would be good with a few more paths here, and some of them nested a few levels.

internal/lsp/eval.go
```diff
@@ -202,3 +202,37 @@ func (h PrintHook) Print(ctx print.Context, msg string) error {
 
 	return nil
 }
+
+func inputSkeletonFromRule(rule *ast.Rule) map[string]any {
+	root := map[string]any{}
+
+	ast.WalkRefs(rule, func(ref ast.Ref) bool {
```
anderseknert: This works, but only for direct references in a rule and not deeper, so this would miss that the `allow` rule below depends on `input.foo` via the reference to `deny`:

```rego
allow if not deny

deny if input.foo == "bar"
```

The EOPA code that I pointed to before resolved deps transitively, and I'd suggest we do the same. Luckily you won't have to write it from scratch 🙂 https://github.com/open-policy-agent/opa/blob/main/v1/dependencies/deps.go

internal/lsp/server.go
```diff
@@ -2614,12 +2615,31 @@ func (l *LanguageServer) handleEvalCommand(ctx context.Context, args types.Comma
...
-				if err = os.WriteFile(inputFile, []byte("{}\n"), 0o600); err != nil {
-					l.log.Message("failed to create input.json: %v", err)
+				skeleton := map[string]any{}
+
+				ruleName := strings.TrimPrefix(args.Query, module.Package.Path.String()+".")
+				for _, rule := range module.Rules {
+					if rule.Head.Name.String() == ruleName {
+						skeleton = inputSkeletonFromRule(rule)
+
+						break
```
charlieegan3: https://github.com/user-attachments/assets/186f7dc9-ff38-4d42-b870-6d7f8e1277b0

I think this is causing an issue where input.* references elsewhere in the rule are not picked up. This might best, not all keys are needed, but when it's different from the one the user clicks it's a bit unexpected.

internal/lsp/eval.go
```diff
@@ -202,3 +202,37 @@ func (h PrintHook) Print(ctx print.Context, msg string) error {
 
 	return nil
 }
+
+func inputSkeletonFromRule(rule *ast.Rule) map[string]any {
```
charlieegan3: I had a go at writing this in Rego:

```rego
package play

obj := json.patch({}, [p| some p in patches])

value(true) := "EXAMPLE"
value(false) := {}


patches contains {"op": "add", "value": value(c == end), "path": str_path} if {
	some ref in refs
    c := count(ref)
    some end in numbers.range(2, c)
	path := array.slice(ref, 1, end)
	str_path := $"/{concat("/", path)}"
}


refs := [
	["input", "bar"],
	["input", "foo", "bar"],
	["input", "wow", "bar-baz"],
]
```

I am unsure if you have already chatted about this and are keen to do v1 in Go, but sharing for fun :)

charlieegan3: obj is the template, once you have a list of ref arrays

internal/lsp/server.go
SeanLedford: Good call, thoughts on "for rule evaluation"?

internal/lsp/server_command_test.go
SeanLedford: Ah, yes definitely much cleaner, addressed!

internal/lsp/eval.go
SeanLedford: Ah yeah makes sense, didn't fully understand what I was missing by pulling only from direct refs. Updated!

internal/lsp/server.go
SeanLedford: Ah good catch, my check wasn't accounting for duplicate rule names. Checking the row now as well!

internal/lsp/eval.go
```diff
@@ -203,13 +204,45 @@ func (h PrintHook) Print(ctx print.Context, msg string) error {
...
+func inputSkeletonFromRule(rule *ast.Rule, compiler *ast.Compiler) map[string]any {
+	root := map[string]any{}
+
+	refs, err := dependencies.Base(compiler, rule)
+	if err != nil {
+		return root
+	}
+
+	// The logic that resolves dependencies in Base doesn't find refs in the rule head.
+	// So, passing that in individually.
```
anderseknert, line 216: Interesting, and good find. We should fix that in OPA. Perhaps something for you to look into later 😊

internal/lsp/server.go
```diff
@@ -2621,16 +2634,10 @@ func (l *LanguageServer) handleEvalCommand(ctx context.Context, args types.Comma
...
-				for _, rule := range module.Rules {
-					if rule.Head.Name.String() == ruleName {
-						skeleton = inputSkeletonFromRule(rule)
+				compiler := compile.NewCompilerWithRegalBuiltins()
+				compiler.Compile(l.cache.GetAllModules())
 
-						break
-					}
-				}
+				skeleton := inputSkeletonFromRule(matchedRule, compiler)
```
anderseknert: Nice

internal/lsp/server_command_test.go
```diff
@@ -263,35 +265,23 @@ func TestExecuteCommandEvalCreatesInputJSON(t *testing.T) {
...
-			Text: "package test\nallow if { input.foo == \"bar\" }\n",
-		},
-	}); err != nil {
-		t.Fatalf("failed to send didOpen: %s", err)
-	}
+allow if {
+	input.foo.bar == "woo"
+	input.foo.wee.age == 24
+	input.superfoo.wee == "fun"
+	input.superbar.tee == "run"
```
anderseknert, line 290: What happens if there are array paths in the input? Like
```
input.list[0].property == "something"
```

It's quite common to have something like `input.user.roles` where those are provided in array form.
OK if we don't handle it yet, but we must ensure that we don't error out on it.

SeanLedford: Yeah good to note this, I'm not handling arrays here currently, or at least not explicitly. I just map everything as an object, even array indices like in your example. Took inspiration from Philip here: https://github.com/open-policy-agent/eopa/blob/e6f35559495dc0a5140e1fbb51e30fd010d7b4f4/pkg/test_bootstrap/testcase_gen.go#L97-L110

I should definitely add this to the test though!

PR #1935, opened 2026-04-07
Include rule docs links in metadata

### LINKED ISSUE #1927: Add related_resources metadata to all rule packages linking to rule docs
The link to the docs for any linter rule is currently generated dynamically on request, such as when the linter reports a violation, on a hover event in the language server, etc. While this works well, it would be better if the rules themselves included pointers to their docs, as having them be more self-contained would make it easier for anyone trying to understand the source code of a rule to quickly jump to documentation describing what it does

Instead of this:

```rego
# METADATA
# description: Redundant count before loop
package regal.rules.bugs["redundant-loop-count"]
```

We should do this:

```rego
# METADATA
# description: Redundant count before loop
# related_resources:
#   - description: documentation
#     ref: https://www.openpolicyagent.org/projects/regal/rules/bugs/redundant-loop-count
package regal.rules.bugs["redundant-loop-count"]
```

And have existing logic around building doc links updated to instead fetch links from a rule's package metadata. We should also update `regal new rule` to automatically render this metadata for new rules.

anderseknert:
This makes rules more self-contained / declarative, and will benefit a future "regodoc" renderer. Related to that, also replace use of config.docs.resolve_url in tests in favor of static URLs. Dynamically resolving doc URLs wasn't ever needed tbh, and having tests compare against static data feels much better.

Also some style fixes to unify how location data is presented in tests.

While there are many changes here, it's mostly repeating the same changes everywhere, so don't spend much time reviewing this.

Fixes #1927

### DIFF
--- a/bundle/regal/ast/keywords_test.rego
+++ b/bundle/regal/ast/keywords_test.rego
@@ -121,13 +121,19 @@ test_keywords_some_no_body if {
 	_keyword_on_row(
 		kwds,
 		6,
-		{"name": "some", "location": {"row": 6, "col": 2, "end": {"col": 6, "row": 6}, "text": "some"}},
+		{
+			"name": "some",
+			"location": {"row": 6, "col": 2, "end": {"col": 6, "row": 6}, "text": "some"},
+		},
 	)
 
 	_keyword_on_row(
 		kwds,
 		6,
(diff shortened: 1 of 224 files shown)


SeanLedford (approved):
LGTM

anderseknert:
It would have been nice to wipe config.docs.resolve_url, but since it's been mentioned in the custom rules docs, there are likely users that depend on it.

PR #1937, opened 2026-04-07
Rule: invalid-regexp

### LINKED ISSUE #1933: Rule: invalid-regex
Like any issue in built-in functions, this is a runtime error / undefined in OPA. We could use OPA's own regex.is_valid function to make this a lint time error in all locations where a regex literal string is provided.

anderseknert:
Simple, but a nice addition!

Also added some "Related Resources" section to rule docs where missing.

Fixes #1933

### DIFF
--- a/bundle/regal/ast/regexp.rego
+++ b/bundle/regal/ast/regexp.rego
@@ -1,0 +1,49 @@
+package regal.ast
+
+import data.regal.util
+
+# METADATA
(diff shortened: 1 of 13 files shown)


SeanLedford (approved):
LGTM! Definitely a neat rule. AST helpers for regex are a nice addition as well.

anderseknert:
Thanks @SeanLedford!

PR #1938, opened 2026-04-07
Cover more `non-loop-expression`s

### LINKED ISSUE #1445: Missing assignment case in non-loop-expression
Could be that we left this for later, I don't remember. But noting it down here so that we'll get to it eventually. It would be good if we flagged assignment expressions too in case that expression doesn't depend on the loop. This is currently not a violation:

```rego
package p

r if {
    some volume in data.volumes
    x := f(input)
}

f(_)
```

While changing `x := f(input)` to `x == f(input)` makes it one.

anderseknert:
We now also flag assignment in iteration which could be moved out of it. Additionally fixed issues with `every` flagging vars outside of its body, so that it now only considers variables and expressions in its own scope.

Fixes #1445


SeanLedford (approved):
Looks good. Would be nice to go over this one together sometime, still haven't fully wrapped my head around it 😅

anderseknert:
For sure! Remind me tomorrow and we’ll find some time.

PR #1943, opened 2026-04-09
testing: Improve lsp shutdown logic & address test race conditions

charlieegan3:
I have been trying to make test suite flakes less common. I am not 100% sure I have got all the issues, but these changes make the flakes and races appear much less often on my local testing. The main things done:

- Add proper shutdown with timeout and worker synchronization. Refactor server test helper to manage the shutdown of the server instances centrally. This helps address issues where workers were continuing to write to pipes after closing.
- Address Aggregate Data races (ast.Object sharing), uses defensive copies when setting aggregate data. ast.Object.Insert() contains non-atomic map/slice operations.
- Address channel blocking issues and timeouts, e.g. in TestTemplateWorkerRaceCondition
- Config access races, locking when accessing workspaceRootURI.
- Fixer state mutation race, create local copy of OPAFmtOpts before mutation.

I have been testing with:

```
bash -c 'i=1; while go test -race ./... -count=1; do echo "Completed iteration $i"; i=$((i+1)); done'
```


internal/lsp/server.go
```diff
@@ -348,451 +336,356 @@ func (l *LanguageServer) SetConn(conn *jsonrpc2.Conn) {
 	l.conn = conn
 }
 
-func (l *LanguageServer) StartDiagnosticsWorker(ctx context.Context) {
-	var wg sync.WaitGroup
+// Shutdown waits for all worker goroutines to complete. The context can be
+// used to set a timeout or cancel the wait if workers take too long to exit.
+// The context passed to workers should be cancelled before calling this method.
+func (l *LanguageServer) Shutdown(ctx context.Context) error {
```
charlieegan3, line 353: might be a better name for this...

internal/lsp/server.go
```diff
@@ -348,451 +336,356 @@ func (l *LanguageServer) SetConn(conn *jsonrpc2.Conn) {
...
+	case <-done:
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
 
-					continue
-				}
+func (l *LanguageServer) StartDiagnosticsWorker(ctx context.Context) {
```
charlieegan3, line 370: see new [diagnosticworker.go](https://github.com/open-policy-agent/regal/pull/1943/changes#diff-542b85a06e241db858a65d398df88b789059f9cd70d11869875b78db2415cfb4)

internal/lsp/diagnosticworker.go
```diff
@@ -0,0 +1,176 @@
...
+
+			l.log.Debug("linting file %s done", job.URI)
+		}
+	}
+}
+
+// startWorkspaceJobRouter routes workspace linting jobs with rate limiting.
+// It listens on l.lintWorkspaceJobs and forwards to workspaceLintRuns,
+// implementing backpressure for aggregate-only reports to prevent performance degradation.
+func startWorkspaceJobRouter(ctx context.Context, l *LanguageServer, workspaceLintRuns chan<- lintWorkspaceJob) {
```
anderseknert: "WorkspaceJobRouter" is so generic I'd have to look up what this does. Could we name it something that includes lint/linting?

internal/lsp/diagnosticworker.go
```diff
@@ -0,0 +1,176 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case job := <-l.lintWorkspaceJobs:
+			// AggregateReportOnly is set when updating aggregate
+			// violations on character changes. Since these happen so
+			// frequently, we stop adding to the channel if there already
+			// jobs set to preserve performance
+			if job.AggregateReportOnly && len(workspaceLintRuns) > 10/2 {
```
anderseknert:
```
> 10/2
```
🤨 
What do these numbers represent? And why not write it as 5?

internal/lsp/diagnosticworker.go
```diff
@@ -0,0 +1,176 @@
...
+		select {
+		case <-ctx.Done():
+			return
+		case job := <-workspaceLintRuns:
+			l.log.Debug("linting workspace: %#v", job)
+
+			// if there are no parsed modules in the cache, then there is
+			// no need to run the aggregate report. This can happen if the
+			// server is very slow to start up.
+			if len(l.cache.GetAllModules()) == 0 {
```
anderseknert: May be worthy of a debug log event

internal/lsp/diagnosticworker.go
```diff
@@ -0,0 +1,176 @@
+package lsp
```
anderseknert: Very nice to move these out!

internal/lsp/diagnosticworker.go
```diff
@@ -0,0 +1,176 @@
...
+				AggregateReportOnly: job.AggregateReportOnly,
+				UpdateForRules:      targetRules,
+				CustomRulesPath:     l.getCustomRulesPath(),
+			})
+			if err != nil {
+				l.log.Message("failed to update all diagnostics: %s", err)
+			}
+
+			for fileURI := range l.cache.GetAllFiles() {
+				l.sendFileDiagnostics(ctx, fileURI)
```
anderseknert: I suppose this isn't new code, but what's the difference between **updating** diagnostics — like we just did above — and sending **diagnostics** like we do here? 🤔

internal/lsp/server_config_test.go
```diff
@@ -70,14 +66,7 @@ allow := true
...
 
-	for success := false; !success; {
-		select {
-		case requestData := <-receivedMessages:
-			success = testRequestDataCodes(t, requestData, mainRegoFileURI, []string{"opa-fmt"})
-		case <-timeout.C:
-			t.Fatalf("timed out waiting for file diagnostics to be sent")
-		}
-	}
+	waitForDiagnostics(t, receivedMessages, mainRegoFileURI, []string{"opa-fmt"}, timeout)
```
anderseknert: Much better!

internal/lsp/server_multi_file_test.go
```diff
@@ -19,8 +17,6 @@ import (
 // workspace diagnostics are run, this test validates that the correct diagnostics are sent to the client in this
 // scenario.
 func TestLanguageServerMultipleFiles(t *testing.T) {
-	// TODO: this test has been flakey and we need to skip it until we have time to look deeper into why
-	t.Skip()
```
anderseknert, line 23: Opening up the sarcophagus!

pkg/roast/util/concurrent/object.go
```diff
@@ -103,6 +104,20 @@ func (co *Object) RenameKey(oldKey, newKey string) {
...
 
+// GetObject returns a copy of the entire object with proper locking.
+func (co *Object) GetObject() ast.Object {
+	co.mu.RLock()
+	defer co.mu.RUnlock()
+
+	// Return a defensive copy to prevent a time-of-check-time-of-use (TOCTOU) race.
+	// Without this copy, callers could access the returned object while another
+	// goroutine calls Set(), which mutates co.o via Insert().
+	return co.o.Copy()
```
anderseknert, line 115: This will be incredibly expensive for large sets of aggregated data. Have you run any benchmarks on this? I get why not doing it could cause a race, but perhaps we can find a way to ensure we avoid it that doesn't involve deep-copying all aggregates. Without having looked into the details, I wonder if we could use the OPA store for this instead of a custom cache. It's made for AST objects after all, and ensures safety through transactions 🤔

anderseknert (commented):
Some great work here Charlie! 👏 

Dropped a few comments, but mainly the aggregate copying I'm worried about. Besides that, huge improvement!

internal/lsp/diagnosticworker.go
charlieegan3: Yeah that used to be `workspaceLintRunBufferSize/2`, I have used a fixed value in 420dc8d3ea7b4ade6c9e1af3defc927390db33c4 to make it clearer.

internal/lsp/diagnosticworker.go
charlieegan3: Yeah that's fair, I have renamed in 12a1dc017d2751eefb3807bf4bc1f371049a9ef7

internal/lsp/diagnosticworker.go
charlieegan3: Added one in 12a1dc017d2751eefb3807bf4bc1f371049a9ef7

pkg/roast/util/concurrent/object.go
charlieegan3: https://github.com/open-policy-agent/regal/pull/1943/commits/b6c498a495023184aaf79ad636f311a8d263dfdc is an attempt to make a step in the right direction here. it's still copy on write, but makes some improvements by keeping the aggregates in the store for linter runs.

anderseknert (approved):
Absolutely stellar!  ⭐️ 🚀 

This solves so many long-standing issues, and many I've had to back off from fixing myself in the past as I couldn't figure it out. Can't thank you enough for this.

My only nit here is the cyclic dependency introduced when other modules import `main`, but that should be easy to resolve later. Let's get this in!!

PR #1951, opened 2026-04-14
Rewrite hover provider in Rego

### LINKED ISSUE #1905: Replace remaining Styra references
With the exception of links to the Styra blog, we should replace links pointing to Styra resources with the equivalent OPA doc pointer. I noticed today as I clicked to view "Examples" after hovering a `contains` keyword that I got to sent to the Styra docs on the topic. These docs are in OPA now though, and we should point to them instead. I am however not 100% sure how it's all wired together, so would be helpful to pair up on this with someone who knows.

See:
* https://github.com/open-policy-agent/regal/blob/main/internal/lsp/examples/examples.go#L16
* https://github.com/open-policy-agent/regal/blob/main/build/workflows/update_example_index.rego#L61

I have replaced other references in #1904

anderseknert:
And get rid of all related code previously written in Go. The biggest win here is IMO how this allows us to entirely get rid of one of our continuous "watcher" processes, and instead use only data from input and storage, like we do in all Rego handlers. Quite an improvement!

Notable other changes:
- Move ast.keywords to lsp.hover.keywords as they're only used in hover
- Move lsp.completion.location to lsp.location as useful outside completion handlers

Besides the tests, also manually tested hovering both built-ins and keywords and all seem to work well!

Fixes #1905


charlieegan3 (approved):
Nice work on the lines removed 👏

PR #1952, opened 2026-04-15
lsp: Gate workspace lint jobs to only run after workspace content has loaded

SeanLedford:
Addresses an error on main where workspace lint jobs are being fired before the workspace contents have a chance to finish loading. 

Changes here block workspace lint jobs until the cache has been set, which solve the issue. But need to verify if this is the right approach, and also how this approach might fit into current work that's aimed at improving lsp shutdown and testing logic.


anderseknert:
I think we can close this after Charlies massive PR, right? :)

PR #1956, opened 2026-04-16
Semantic package tokens in Rego

anderseknert:
Finally had a moment to play some with @SeanLedford's great work on semantic tokens, and I could not resist seeing what moving a first part of the token creation over to Rego. Feels pretty straightforward, although a bit of a pain updating the token tests. But that's a good problem to have!

### DIFF
--- a/internal/lsp/semantictokens/semantictokens.go
+++ b/internal/lsp/semantictokens/semantictokens.go
@@ -12,16 +12,35 @@ import (
 	"github.com/open-policy-agent/regal/internal/lsp/types"
 )
 
-const (
-	TokenTypePackage  = 0
-	TokenTypeVariable = 1
-	TokenTypeImport   = 2
+type (
+	TokenType     = uint32
+	TokenModifier = uint32
 )
 
 const (
-	ModifierDeclaration = 1 << 0
-	ModifierReference   = 1 << 1
+	Namespace TokenType = iota
+	Variable
+	Import
+	Keyword
 )
+
+const (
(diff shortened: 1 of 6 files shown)


internal/lsp/semantictokens/semantictokens.go
```diff
@@ -12,17 +12,36 @@ import (
...
+	Namespace TokenType = iota
+	Variable
+	Import
+	Keyword
 )
 
 const (
-	ModifierDeclaration = 1 << 0
-	ModifierReference   = 1 << 1
+	ModifierDeclaration TokenModifier = iota
```
SeanLedford: Didn't know about this iota counter, pretty cool! I don't think this works for this case though, we want to populate TokenModifier with individual bit values. This is how the LSP defines them, basically just gives us the ability to create a bunch of different modifiers using only one field. We could do something like this though:

```suggestion
	ModifierDeclaration TokenModifier = 1 << iota
```

anderseknert: TIL! Happy you caught that, as I wrongly assumed the modifiers worked the same way as the types. Indeed we can use  `iota` as you suggested. I'll fix this tomorrow. Besides that, looks OK?

SeanLedford (approved):

internal/lsp/semantictokens/semantictokens.go
SeanLedford: Yep looks good! Just stamped.

PR #1960, opened 2026-04-19
Semantic import tokens Rego rewrite

anderseknert:
As with the package rewrite, also return a token for the `import` keyword.

Now also handles aliased imports, and will treat the alias as the imported identifier rather than the last path component. We really ought to have the RoAST encoder encode an alias as a proper term with location, but that will have to come as a separate change as it touches quite many components.


charlieegan3 (approved):

SeanLedford (approved):
LGTM! Great work Anders, the alias support is nice too

PR #1965, opened 2026-04-21
Minor fixes

anderseknert:
- Add missing docs on `names` config option for `naming-convention` rule
- Fix case of [repeated computation](https://github.com/open-policy-agent/regal/issues/1820)

### DIFF
--- a/bundle/regal/rules/custom/one-liner-rule/one_liner_rule.rego
+++ b/bundle/regal/rules/custom/one-liner-rule/one_liner_rule.rego
@@ -60,10 +60,12 @@ _rule_body_brackets(lines) if {
 }
 
 _comment_in_body(rule_row, comments, lines) if {
+	num_lines := count(lines)
+
 	some comment in comments
 
 	comment_row := to_number(util.substring_to(comment.location, 0, ":"))
 
 	comment_row > rule_row
-	comment_row < rule_row + count(lines)
+	comment_row < rule_row + num_lines
 }
--- a/docs/rules/custom/naming-convention.md
+++ b/docs/rules/custom/naming-convention.md
@@ -32,18 +32,27 @@ rules:
       # one of "error", "warning", "ignore"
       level: error
       conventions:
-          # allow only "private" rules and functions, i.e. those starting with
-          # underscore, or rules named "deny" or "allow"
-        - pattern: '^_[a-z]+$|^deny$|^allow$'
+        # allow only "private" rules and functions, i.e. those starting with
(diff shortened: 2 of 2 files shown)


docs/rules/custom/naming-convention.md
```diff
@@ -32,18 +32,27 @@ rules:
...
+        # underscore, or rules named "deny" or "allow"
+        - pattern: "^_[a-z]+$|^deny$|^allow$"
           # one of "package", "rule", "function", "variable"
           targets:
             - rule
             - function
         # any number of naming rules may be added
         # package names must start with "acmecorp" or "system"
-        - pattern: '^acmecorp|^system'
+        - pattern: "^acmecorp|^system"
```
anderseknert, line 44: I really don't like how both the markdown and YAML formatter does this.

srenatus: Does what? Ah, replace the characters? 👍

srenatus (approved):

bundle/regal/rules/custom/one-liner-rule/one_liner_rule.rego
```diff
@@ -60,10 +60,12 @@ _rule_body_brackets(lines) if {
...
 _comment_in_body(rule_row, comments, lines) if {
+	num_lines := count(lines)
+
 	some comment in comments
 
 	comment_row := to_number(util.substring_to(comment.location, 0, ":"))
 
 	comment_row > rule_row
-	comment_row < rule_row + count(lines)
+	comment_row < rule_row + num_lines
```
srenatus, line 70: Took me a moment to understand that this _is indeed_ a repeated computation. 👍

docs/rules/custom/naming-convention.md
anderseknert: Indeed. It _will_ use single quotes when certain characters are present as that's required (IIRC) but that obviously makes quoting inconsistent through the document. Infuriating.

bundle/regal/rules/custom/one-liner-rule/one_liner_rule.rego
anderseknert: Yes, clearly we need that linter rule :) 

Should perhaps be done by the compiler too/instead... not sure how far we want to go with optimizations there though.

srenatus: I'd love to have the compiler deal with this.

PR #1967, opened 2026-04-22
Update ast.json else to use "#/$defs/rule"

sspaink:
follow up of: https://github.com/open-policy-agent/opa/pull/8542

### DIFF
--- a/internal/embeds/schemas/regal/ast.json
+++ b/internal/embeds/schemas/regal/ast.json
@@ -127,33 +127,7 @@
           "$ref": "#/$defs/body"
         },
         "else": {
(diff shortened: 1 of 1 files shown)


anderseknert:
Thanks! Definitely seeing the overflow here, so will be good to see it fixed by rebasing once we've bumped OPA. We have so many things depending on that now that I think I'm going to pull opa `main` as our OPA dep until the release, so that we get some time to test things like this. Keeping you posted!

srenatus (approved):

anderseknert:
Beautiful 🥲

PR #1976, opened 2026-04-28
feat(pre-commit): add regal-fix hooks for auto-fixing rules

### LINKED ISSUE #1847: Pre-commit hook for fixing rules
Currently, there is a pre-commit hook that runs regal lint, but I believe there is no way to make it autofix the issues it finds via pre-commit. Now I have to have this in my pre-commit:
```

  - repo: local # The regal fix command doesn't seem to have a pre-commit hook in the regal repo.
    hooks:
      - id: regal-fmt
        language: system
        name: regal format
        entry: regal fix --force .
  - repo: https://github.com/open-policy-agent/regal
    rev: v0.38.1
    hooks:
      - id: regal-lint
```

It would be great if the regal-lint hook took an argument to make it autofix what it can, or a different hook is added for this functionality. Not sure what you'd prefer. Thanks!

mvanhorn:
## Summary

Closes #1847.

The existing pre-commit hooks (`regal-lint`, `regal-lint-use-path`, `regal-download`) all run `regal lint`, which only reports. The reporter wanted pre-commit to also auto-fix what `regal fix` can repair, and was working around the gap with a `repo: local` hook that called `regal fix --force .` directly.

This adds three hook IDs that mirror the lint trio but invoke `regal fix --force`:

- `regal-fix` - `language: golang`, builds Regal from source
- `regal-fix-use-path` - `language: system`, uses `regal` on `$PATH`
- `regal-fix-download` - `language: script`, runs `build/download-and-run.sh fix --force`

`--force` is required because pre-commit invokes hooks against files staged in a dirty git tree, which `regal fix` would otherwise refuse without `--force` (per `cmd/fix.go`).

## Files

- `.pre-commit-hooks.yaml` - three new hook entries.
- `docs/pre-commit-hooks.md` - documentation sections mirroring the lint sections.

## Usage

```yaml
repos:
  - repo: https://github.com/open-policy-agent/regal
    rev: v0.40.0
    hooks:
      - id: regal-fix       # auto-fix what can be fixed
      - id: regal-lint      # then lint anything left
```

The `regal-download` script (`build/download-and-run.sh`) already accepts the subcommand as `$@`, so the new download hook works with no script changes.

## Test plan

- [x] `.pre-commit-hooks.yaml` parses as valid YAML and lists 6 hooks.
- [x] Each new hook entry has matching `id`, `language`, `entry`, `files` fields consistent with its lint counterpart.
- [x] Documentation entries mirror the existing lint sections.
- [x] Manual smoke test by a maintainer in their own pre-commit config (no automated test infra here).

No new dependencies. No code changes to Regal itself.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -18,3 +18,24 @@
   description: policy file linting with Regal downloaded from Github
   entry: build/download-and-run.sh lint
   files: (\.rego)$
+
+- id: regal-fix
+  language: golang
+  name: policy file fixing
+  description: auto-fix policy files with Regal from Styra built from source
+  entry: regal fix --force
+  files: (\.rego)$
+
+- id: regal-fix-use-path
+  language: system
+  name: policy file fixing
+  description: auto-fix policy files with Regal from Styra using system $PATH
+  entry: regal fix --force
+  files: (\.rego)$
+
+- id: regal-fix-download
+  language: script
+  name: policy file fixing
+  description: auto-fix policy files with Regal downloaded from Github
+  entry: build/download-and-run.sh fix --force
+  files: (\.rego)$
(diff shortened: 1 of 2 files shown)


anderseknert (approved):
LGTM!

PR #1977, opened 2026-04-28
Move initialize handler to Rego router

anderseknert:
It's leaner, it's faster, it's Rego! I have wanted to do this for a long time, but it wasn't prioritized. But what are weekends for anyway.

Also:
- Remove types that were only used in Go initialize
- Remove Go functions no longer used
- Handle *json.RawMessage in rast.AnyToValue


internal/lsp/types/internal.go
```diff
@@ -44,41 +50,44 @@ type CommandArgs struct {
...
 func NewGenericClient() Client {
 	return Client{Identifier: clients.IdentifierGeneric}
 }
 
-func (c Client) SupportsExplorer() bool {
-	return c.InitOptions != nil &&
-		c.InitOptions.EnableExplorer != nil &&
-		*c.InitOptions.EnableExplorer
-}
+func (c *Client) UnmarshalJSON(data []byte) (err error) {
```
charlieegan3, line 62: I wonder if it makes sense to continue to use the enum here? replicating the values into the rego policy feels a bit leaky and makes this unmarshalling more complicated.

I don't have a strong opinion, just a thought.

bundle/regal/lsp/initialize/initialize_test.rego
```diff
@@ -0,0 +1,46 @@
...
+
+	response.capabilities.experimental.explorerProvider == true
+	response.capabilities.experimental.inlineEvalProvider == true
+	response.capabilities.experimental.debugProvider == true
+	response.capabilities.experimental.opaTestProvider == true
+}
+
+test_initialize_client_identifier[client_name] if {
+	some identifier, client_name in [
+		"Space Editor!!1!",
```
charlieegan3, line 27: 😆 

This might be a little clearer as something like 'previously unseen client name' etc.

charlieegan3 (approved):

bundle/regal/lsp/initialize/initialize_test.rego
anderseknert: Keep Regal Quirky!! ;)

internal/lsp/types/internal.go
anderseknert: I think it's a temporary measure. Once we have all handlers moved we won't need structs like Client at all, but can have those be ast.Value's too, and the handlers that manage workspace state changes simply be an orchestrating layer on top of storage. A little more to do before we're there though ☺️

PR #1978, opened 2026-04-29
Cleanup

anderseknert:
Just a few followup fixes from the last PR

- Move most of the command handlers code into a separate file to further reduce the size of server.go. Not exactly proper separation / isolation yet, but it's a first step.
- Remove parser code for command argument format no longer used


bundle/regal/lsp/initialize/initialize_test.rego
```diff
@@ -24,7 +24,7 @@ test_initialize_experimental_capabilities if {
 
 test_initialize_client_identifier[client_name] if {
 	some identifier, client_name in [
-		"Space Editor!!1!",
+		"Space Editor!!1! (unknown client)",
```
anderseknert, line 27: For you @charlieegan3 :)

charlieegan3: appreciated!

charlieegan3 (approved):

PR #1979, opened 2026-04-30
Move folding range provider to Rego handlers

anderseknert:
Like the previous version, the handler provides folding ranges for:

- Comment blocks
- Imports
- Rules, functions, collections and comprehensions that span more than a single line

The new implementation additionally honors client capabilities like `lineFoldingOnly` and `rangeLimit`, and the documentation has been updated to cover this.

Also:
- Remove scanner code previously borrowed from OPA as it was only ever used for this purpose
- Use a single schema for all textdocument handlers
- Fix bug in AnyToValue that turned `json.Number`s into strings
- Fix potential concurrrency issue using shared oracle
- Move client information from input to store as we don't need to recompute that per request


internal/lsp/server.go
```diff
@@ -893,15 +895,15 @@ func (l *LanguageServer) handleTextDocumentDefinition(params types.DefinitionPar
...
-	query := oracle.DefinitionQuery{
-		// The value of Filename is used if the defn in the current buffer.
-		Filename: l.toRelativePath(params.TextDocument.URI),
-		Pos:      positionToOffset(contents, params.Position),
-		Modules:  modules,
-		Buffer:   outil.StringToByteSlice(contents),
-	}
-
-	definition, err := orc.WithCompiler(compile.NewCompilerWithRegalBuiltins()).FindDefinition(query)
+	definition, err := oracle.New().
```
philipaconrad, line 898: I was curious about the shared oracle bits, and this makes sense as a solution. Just mint a new oracle on every `handleTextDocument` call!

philipaconrad (approved):
I looked over most of this PR yesterday afternoon, and took a second pass this morning for the remaining bits. I can't comment much on the implementation of the folding range providers, except that it looks like they now exist for all the listed types with reasonable implementations, and have tests.

I focused most of my review time on the "other" bits that got mixed in, and can confirm all of the items listed were done, and have reasonable-looking implementations. 🙂

PR #1982, opened 2026-05-05
[feature] First Pass at Automatic Test Generation in Regal

### LINKED ISSUE #1928: Add code action for generating a test from rule evaluation
The `Evaluate` feature enables exploratory policy development, where changes to either policy, input or data immediately are reflected in the evaluation result. This goes a long way to build understanding and confidence in the policy as it's being developed. Ideally, we'd be able to transfer this confidence into something that persists past the development session! I believe one way to do this would be to provide an option to generate a test from the same state that the workspace is in when the user clicks `Evaluate`. 

Given a (overly trivial) workspace like this:

**policy.rego**
```rego
package policy

allow if {
    input.foo == "foo"
    data.bar == "bar"
}
```

**input.json**
```json
{"foo": "foo"}
```

**data.json**
```json
{"bar": "bar"}
```

Clicking Evaluate would of course return `true` for the given workspace state. Translating the evaluation to a test from that would then render something like:

**policy_test.rego**
```rego
package policy_test

import data.policy

test_allow {
    policy.allow with input.foo as "foo" with data.bar as "bar"
}
```

Changing e.g. `input.json` to `{"foo": "bar"}` would make the result undefined, and we could easily add a test for that condition too:

**policy_test.rego**
```rego
package policy_test

import data.policy

test_allow {
    policy.allow with input.foo as "foo" with data.bar as "bar"
}

test_not_allow {
    not policy.allow with input.foo as "bar" with data.bar as "bar"
}
```

Exactly how test names should be generated will be left as an exercise to the implementor... but the name should most likely be provided / refined by the user anyway.

This would definitely not be a trivial task with all things considered (identifying and narrowing down dependencies to mock, for one), but I imagine we can do this gradually, and perhaps have a first version copy the whole input document to the test, not consider data.json files, and so on. Even in a crude form requiring some manual edits would still save time compared to writing all tests from scratch.

SeanLedford:
A new lsp feature for automatically creating tests for rules in a policy, based on what's provided in input/data.json, and what is actually referenced from those files in the rules.

Currently facing several limitations, mainly how to handle more complex dependencies like array access, but will work to support those eventually as well!

Fixes #1928

### DIFF
--- a/bundle/regal/lsp/codeaction/codeaction.rego
+++ b/bundle/regal/lsp/codeaction/codeaction.rego
@@ -127,10 +127,7 @@ actions contains action if {
 			"title": "Create tests for this file",
 			"command": "regal.createTest",
 			"tooltip": "Create test cases for all rules in this file",
-			"arguments": [{
-				"target": input.params.textDocument.uri,
-				"row": input.params.range.start.line,
-			}],
+			"arguments": [{"target": input.params.textDocument.uri}],
 		},
 	}
 }
--- a/bundle/regal/lsp/codeaction/codeaction.rego
+++ b/bundle/regal/lsp/codeaction/codeaction.rego
@@ -115,6 +115,27 @@ actions contains action if {
 }
 
 # METADATA
+# description: |
+#   Code action to create a test from the current rule evaluation state.
+actions contains action if {
+	strings.any_prefix_match("source.createTest", only)
+
+	action := {
+		"title": "Create tests for this file",
+		"kind": "source.createTest",
+		"command": {
+			"title": "Create tests for this file",
+			"command": "regal.createTest",
+			"tooltip": "Create test cases for all rules in this file",
+			"arguments": [{
+				"target": input.params.textDocument.uri,
+				"row": input.params.range.start.line,
+			}],
+		},
+	}
+}
+
+# METADATA
 # description: All code actions for fixing reported diagnostics
 rules := {
 	"opa-fmt": ["Format using opa-fmt", ["target"]],
@@ -137,6 +158,6 @@ rules := {
 #   be hierarchical — if only contains "source" it matches all source actions,
 #   while "source.foo" matches only source actions with a "foo" prefix.
 # scope: document
-default only := ["quickfix", "source.explore"]
+default only := ["quickfix", "source.explore", "source.createTest"]
 
 only := input.params.context.only if input.params.context.only != []
--- a/bundle/regal/lsp/initialize/initialize.rego
+++ b/bundle/regal/lsp/initialize/initialize.rego
@@ -67,25 +67,17 @@ _capabilities.workspace.fileOperations[operation].filters := filters if {
 ## *there is no way* to support it here.
 _capabilities.workspace.workspaceFolders.supported := true
 
-_capabilities.inlayHintProvider := {
-	# inlayHint/resolve request supported
-	"resolveProvider": true,
-}
+_capabilities.inlayHintProvider := {"resolveProvider": true} # inlayHint/resolve request supported
 
 _capabilities.hoverProvider := true
 
-_capabilities.signatureHelpProvider := {
-	# In additional to the client's default trigger characters for signature help
-	"triggerCharacters": ["(", ","]
-}
+# In additional to the client's default trigger characters for signature help
+_capabilities.signatureHelpProvider := {"triggerCharacters": ["(", ","]}
 
-_capabilities.codeActionProvider := {
-	# Currently supported code action kinds
-	"codeActionKinds": [
-		"quickfix",
-		"source"
-	],
-}
+_capabilities.codeActionProvider := {"codeActionKinds": [
+	"quickfix", # Currently supported code action kinds
+	"source",
+]}
 
 _capabilities.executeCommandProvider.commands := _commands
 
@@ -102,6 +94,7 @@ _commands contains "regal.fix.redundant-existence-check"
 _commands contains "regal.config.disable-rule"
 _commands contains "regal.explorer" if data.server.feature_flags.explorer_provider
 _commands contains "regal.debug" if data.server.feature_flags.debug_provider
+_commands contains "regal.createTest" if data.server.feature_flags.test_creation_provider
 
 _capabilities.documentFormattingProvider := true
 
@@ -122,15 +115,9 @@ _capabilities.completionProvider := {
 	"completionItem": {"labelDetailsSupport": true},
 }
 
-_capabilities.codeLensProvider := {
-	# codeLens/resolve to be implemented
-	"resolveProvider": false,
-}
+_capabilities.codeLensProvider := {"resolveProvider": false} # codeLens/resolve to be implemented
 
-_capabilities.documentLinkProvider := {
-	# documentLink/resolve to be implemented
-	"resolveProvider": false,
-}
+_capabilities.documentLinkProvider := {"resolveProvider": false} # documentLink/resolve to be implemented
 
 _capabilities.documentHighlightProvider := true
 
--- a/bundle/regal/lsp/initialize/initialize.rego
+++ b/bundle/regal/lsp/initialize/initialize.rego
@@ -67,17 +67,25 @@ _capabilities.workspace.fileOperations[operation].filters := filters if {
 ## *there is no way* to support it here.
 _capabilities.workspace.workspaceFolders.supported := true
 
-_capabilities.inlayHintProvider := {"resolveProvider": true} # inlayHint/resolve request supported
+_capabilities.inlayHintProvider := {
+	# inlayHint/resolve request supported
+	"resolveProvider": true,
+}
 
 _capabilities.hoverProvider := true
 
-# In additional to the client's default trigger characters for signature help
-_capabilities.signatureHelpProvider := {"triggerCharacters": ["(", ","]}
+_capabilities.signatureHelpProvider := {
+	# In additional to the client's default trigger characters for signature help
+	"triggerCharacters": ["(", ","]
+}
 
-_capabilities.codeActionProvider := {"codeActionKinds": [
-	"quickfix", # Currently supported code action kinds
-	"source",
-]}
+_capabilities.codeActionProvider := {
+	# Currently supported code action kinds
+	"codeActionKinds": [
+		"quickfix",
+		"source"
+	],
+}
 
 _capabilities.executeCommandProvider.commands := _commands
 
@@ -115,9 +123,15 @@ _capabilities.completionProvider := {
 	"completionItem": {"labelDetailsSupport": true},
 }
 
-_capabilities.codeLensProvider := {"resolveProvider": false} # codeLens/resolve to be implemented
+_capabilities.codeLensProvider := {
+	# codeLens/resolve to be implemented
+	"resolveProvider": false,
+}
 
-_capabilities.documentLinkProvider := {"resolveProvider": false} # documentLink/resolve to be implemented
+_capabilities.documentLinkProvider := {
+	# documentLink/resolve to be implemented
+	"resolveProvider": false,
+}
 
 _capabilities.documentHighlightProvider := true
 
--- a/internal/lsp/server.go
+++ b/internal/lsp/server.go
@@ -112,6 +112,7 @@ func DefaultServerFeatureFlags() *types.ServerFeatureFlags {
 		InlineEvaluationProvider: true,
 		DebugProvider:            true,
 		OPATestProvider:          true,
+		TestCreationProvider:     true,
 	}
 }
 
--- a/internal/lsp/server_command.go
+++ b/internal/lsp/server_command.go
@@ -560,9 +560,8 @@ func (l *LanguageServer) handleCreateTestCommand(ctx context.Context, params typ
 		return errors.New("target file URI is required")
 	}
 
-	filePath := uri.ToPath(args.Target)
-
-	inputPath, inputValue := rio.FindInput(filePath, l.workspacePath())
+	inputPath := l.input.FindForPath(args.Target)
+	inputValue := l.input.Get(ctx, inputPath)
 
 	inputEmpty := inputValue == nil
 	if obj, ok := inputValue.(ast.Object); ok {
@@ -570,14 +569,13 @@ func (l *LanguageServer) handleCreateTestCommand(ctx context.Context, params typ
 	}
 
 	if inputPath == "" || inputEmpty {
-		if err := l.conn.Notify(ctx, "window/showMessage", types.ShowMessageParams{
-			Type:    3,
-			Message: "No input.json or input.yaml file found. Create one to provide test input data.",
-		}); err != nil {
-			l.log.Message("failed to show input.json missing message: %v", err)
-		}
+		l.window.ShowMessage(
+			ctx,
+			types.InfoMessage,
+			"No input.json or input.yaml file found. Create one to provide test input data.",
+		)
 
-		return errors.New("no input.json found in workspace")
+		return nil
 	}
 
 	_, module, ok := l.cache.GetContentAndModule(args.Target)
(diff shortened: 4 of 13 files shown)


bundle/regal/lsp/codeaction/codeaction.rego
```diff
@@ -112,6 +112,27 @@ actions contains action if {
...
 }
 
+# METADATA
+# description: |
+#   Code action to generate a test from the current rule evaluation state.
+actions contains action if {
+	strings.any_prefix_match("source.generateTest", only)
+
+	action := {
+		"title": "Generate tests for this file",
```
charlieegan3: I wonder if template might be a less surprising word than generate, in the era of gen ai.

internal/lsp/server.go
```diff
@@ -1677,29 +2254,327 @@ func (l *LanguageServer) parseOpts(fileURI string, bis map[string]*ast.Builtin)
...
+		return errors.New("target file URI is required")
+	}
+
+	filePath := uri.ToPath(args.Target)
+
+	inputPath, inputMap := rio.FindInput(filePath, l.workspacePath())
+	if inputPath == "" || len(inputMap) == 0 {
+		if err := l.conn.Notify(ctx, "window/showMessage", types.ShowMessageParams{
+			Type:    3,
+			Message: "No input.json or input.yaml file found. Create one to provide test input data.",
```
charlieegan3: I seem to remember we have this elsewhere as well, I wonder if that can be shared?

internal/lsp/server.go
```diff
@@ -1677,29 +2254,327 @@ func (l *LanguageServer) parseOpts(fileURI string, bis map[string]*ast.Builtin)
...
+		return errors.New("no rules found in this file")
+	}
+
+	allModules := l.cache.GetAllModules()
+
+	var testFunctions []string
+
+	packagePath := module.Package.Path.String()
+
+	for _, rule := range module.Rules {
```
charlieegan3: I feel like module -> combinedTest could be in the testgen package rather than the server.go file. (which is pretty massive)

internal/lsp/server_command_test.go
```diff
@@ -336,3 +336,121 @@ allow if {
 		t.Fatal("timed out waiting for window/showDocument")
 	}
 }
+
+func TestExecuteCommandGenerateTest(t *testing.T) {
```
charlieegan3: I would also like to see some module->generated test cases in the test gen package (more unit tests for the main cases.)

bundle/regal/lsp/codeaction/codeaction.rego
anderseknert: How about "create"? OTOH it's not like we didn't generate code before LLMs.. but I if anyone understand the negative connotation you're alluding to, lol

internal/lsp/testgen/testgen.go
```diff
@@ -0,0 +1,227 @@
...
+			return "", fmt.Errorf("failed to parse even basic test: %w", err)
+		}
+	}
+
+	module, err := rparse.Module("test.rego", completeTest)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse test: %w", err)
+	}
+
+	formatted, err := format.Ast(module)
```
anderseknert: I think you could use format.Source here and avoid the parsing above

internal/lsp/server.go
```diff
@@ -1677,29 +2254,327 @@ func (l *LanguageServer) parseOpts(fileURI string, bis map[string]*ast.Builtin)
...
+
+				rpcCancel()
+			}
+		}
+	}
+
+	return nil
+}
+
+func (l *LanguageServer) handleGenerateTestCommand(ctx context.Context, params types.ExecuteCommandParams) error {
```
anderseknert: I'd suggest moving this into server_command.go

internal/lsp/server.go
anderseknert: Indeed... a `ShowMessage` function would be nice too, for getting rid of some boilerplate :)

internal/lsp/server.go
```diff
@@ -1677,29 +2254,327 @@ func (l *LanguageServer) parseOpts(fileURI string, bis map[string]*ast.Builtin)
...
+	}
+
+	showParams := types.ShowDocumentParams{
+		URI:       uri.FromPath(l.getClient().Identifier, testFileName),
+		TakeFocus: new(true),
+	}
+
+	var result types.ShowDocumentResult
+
+	rpcCtx, rpcCancel := context.WithTimeout(context.Background(), rpcTimeout)
```
anderseknert: Any reason we're not passing in the context in this function?

internal/lsp/server.go
```diff
@@ -1677,29 +2254,327 @@ func (l *LanguageServer) parseOpts(fileURI string, bis map[string]*ast.Builtin)
...
+		testFunction, err := testgen.GenerateTestFunction(testgen.TestGenerationOptions{
+			RuleName:      ruleName,
+			PackagePath:   packagePath,
+			WorkspacePath: l.workspacePath(),
+			FileURI:       args.Target,
+			Rule:          rule,
+			AllModules:    allModules,
+		})
+		if err != nil {
+			continue
```
anderseknert: Should we log something here, or this is very common?

internal/lsp/server.go
```diff
@@ -1677,29 +2254,327 @@ func (l *LanguageServer) parseOpts(fileURI string, bis map[string]*ast.Builtin)
...
+		return fmt.Errorf("could not get file contents for uri %q", args.Target)
+	}
+
+	if len(module.Rules) == 0 {
+		return errors.New("no rules found in this file")
+	}
+
+	allModules := l.cache.GetAllModules()
+
+	var testFunctions []string
```
anderseknert: A strings.Builder would be more efficient here, without too much added clutter, I think

anderseknert (commented):
Looks like an excellent start 👍 Some comments to address or look into at least, but should be good for a first iteration then. Good work!

bundle/regal/lsp/codeaction/codeaction.rego
SeanLedford: Yeah good call, updated my naming conventions to "create"

internal/lsp/server.go
SeanLedford: Good call, moved it over.

internal/lsp/server_command_test.go
SeanLedford: Yeah agreed, added a file dedicated to unit tests!

internal/lsp/testgen/testgen.go
SeanLedford: Ah yeah that works, addressed.

internal/lsp/server.go
SeanLedford: Good call, made the switch here.

internal/lsp/server.go
SeanLedford: Yeah this was an oversight, added some logging here.

internal/lsp/server.go
SeanLedford: Good call, addressed!

internal/lsp/server.go
SeanLedford: They are very similar, but the version of this that I wrote for my input.json feature uses `showMessageRequest`, which prompts the user for input. Probably could make a helper for these types of requests, but I think it's a bit out of scope here.

internal/lsp/server.go
SeanLedford: Looking at other handlers, this looked like the precedent for `showDocument` and other requests that we wanted to complete regardless of interruption. I added the comment that was present elsewhere here: `// Use a timeout context for RPC to ensure it completes during graceful shutdown`.

internal/lsp/server.go
anderseknert: You can use [this](https://github.com/open-policy-agent/regal/blob/main/internal/lsp/window/window.go) now :)

bundle/regal/lsp/initialize/initialize.rego
```diff
@@ -57,25 +57,17 @@ _capabilities.workspace.fileOperations[operation].filters := filters if {
...
-	# Currently supported code action kinds
-	"codeActionKinds": [
-		"quickfix",
-		"source"
-	],
-}
+_capabilities.codeActionProvider := {"codeActionKinds": [
+	"quickfix", # Currently supported code action kinds
+	"source",
+]}
```
anderseknert: I don't understand the above changes. Are you on an older version of OPA where the formatter did this? I believe that was fixed some time ago.

bundle/regal/lsp/initialize/initialize.rego
```diff
@@ -112,15 +105,9 @@ _capabilities.completionProvider := {
...
-	# codeLens/resolve to be implemented
-	"resolveProvider": false,
-}
+_capabilities.codeLensProvider := {"resolveProvider": false} # codeLens/resolve to be implemented
 
-_capabilities.documentLinkProvider := {
-	# documentLink/resolve to be implemented
-	"resolveProvider": false,
-}
+_capabilities.documentLinkProvider := {"resolveProvider": false} # documentLink/resolve to be implemented
```
anderseknert: Same with the above two changes.

internal/lsp/testgen/testgen.go
```diff
@@ -0,0 +1,295 @@
...
+	if len(opts.Module.Rules) == 0 {
+		return "", nil, errors.New("no rules found in this file")
+	}
+
+	packagePath := opts.Module.Package.Path.String()
+
+	var b strings.Builder
+	b.WriteString(BuildTestHeader(packagePath))
+
+	var ruleErrs []RuleError
```
anderseknert: Instead of this, could we use [errors.Join](https://pkg.go.dev/errors#Join) to concat errors into a single one? Would make for slightly nicer function signatures, I think.

internal/lsp/testgen/testgen.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+
+	refs, err := dependencies.Base(compiler, opts.Rule)
+	if err != nil {
+		return nil, fmt.Errorf("failed to analyze rule dependencies: %w", err)
+	}
+
+	headRefs, err := dependencies.Base(compiler, opts.Rule.Head)
+	if err != nil {
+		return nil, fmt.Errorf("failed to analyze rule head dependencies: %w", err)
```
anderseknert, line 104: We should fix this in OPA instead (but fine to work around here in the meantime). Have you filed an issue about this there? If not, that'd be great!

internal/lsp/testgen/testgen.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+
+	refs = append(refs, headRefs...)
+
+	filePath := uri.ToPath(opts.FileURI)
+	_, inputValue := rio.FindInput(filePath, opts.WorkspacePath)
```
anderseknert: The `FindInput` function died in #2002. You'll have to pass in an `*input.Manager` from the server instead, but should be fairly easy I hope. Let me know if you need any help with that!

internal/lsp/testgen/testgen.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+
+	testFunction := buildTestFunction(opts, withClauses)
+	completeTest := BuildTestHeader(opts.PackagePath) + "\n\n" + testFunction
+
+	formatted, err := format.Source("test.rego", []byte(completeTest))
+	if err != nil {
+		// Array access syntax like input.permissions[0] can cause parsing issues —
+		// fall back to a basic test skeleton without with-clauses.
+		// TODO: handle array access in withClause generation.
```
anderseknert: Interesting. Do you have an example of this at hand?

internal/lsp/testgen/testgen.go
```diff
@@ -0,0 +1,295 @@
...
+			return "", fmt.Errorf("failed to format basic test: %w", err)
+		}
+	}
+
+	lines := strings.Split(string(formatted), "\n")
+	for i, line := range lines {
+		if strings.HasPrefix(strings.TrimSpace(line), "test_") {
+			return strings.Join(lines[i:], "\n"), nil
+		}
+	}
```
anderseknert, line 254: Can probably avoid the Split with something this, I think:
```go
for i, line := range util.Lines(formatted) {
	if bytes.HasPrefix(bytes.TrimSpace(line), []byte("test_")) {
		return string(formatted[i:])
	}
}
```
But doesn't matter much 🙂

anderseknert (approved):
Looks great to me! Left a few comments and suggestions, but nothing blocking once you've rebased and dealt with a few recent changes.

bundle/regal/lsp/initialize/initialize.rego
SeanLedford: Hmm, yeah it's been a bit now since this commit, can't recall why I reformatted either. Regardless, reverted now!

internal/lsp/testgen/testgen.go
SeanLedford: Yeah will do!

internal/lsp/testgen/testgen.go
SeanLedford: Gotcha, made some updates here to account for this. Let me know if you notice any issues

internal/lsp/testgen/testgen.go
SeanLedford: Yeah the comment could've been worded better, I updated it. Basically, given:

```
allow if input.permissions[0] == "admin"
  ```

  The skeleton generator produces:
  ```json
  {"permissions": {"0": "changeme"}}
  ```

This is because we still haven't found a way to reliably infer what type of ref we're getting from indexed syntax(like [0])

  So in this case we'd get invalid rego:
  ```rego
  test_allow if {
      policy.allow with input.permissions.0 as "changeme"
  }
  ```

This is really just to ensure the `create input skeleton -> create test` flow creates valid rego for now, but definitely still a WIP

File: internal/lsp/testgen/testgen.go (hunk 16 above)
Comment (SeanLedford) [reply internal/lsp/testgen/testgen.go:254]: Hmm yeah I couldn't get this quite right when trying it out, might need to mess around with this util function a bit more.

File: internal/lsp/testgen/testgen.go (hunk 12 above)
Comment (SeanLedford) [reply internal/lsp/testgen/testgen.go]: Gotcha good idea, updated!

File: internal/lsp/testgen/testgen.go (hunk 16 above)
Comment (anderseknert) [reply internal/lsp/testgen/testgen.go:254]: Probably needs to b `formatted[i+1:]` or something like that. But it's fine! It's a bit of a micro-optimization anyway and I was thinking out loud :)

File: internal/lsp/testgen/testgen_test.go [hunk 17]
```diff
@@ -0,0 +1,154 @@
...
+			workspacePath := testutil.TempDirectoryOf(t, files)
+			policyPath := filepath.Join(workspacePath, "policy.rego")
+			fileURI := uri.FromPath(clients.IdentifierGoTest, policyPath)
+
+			module, err := rparse.Module("policy.rego", tc.policy)
+			if err != nil {
+				t.Fatalf("failed to parse policy: %v", err)
+			}
+
+			im := input.NewManager(store.NewRegalStore(), log.NewLogger(log.LevelDebug, os.Stderr))
```
anderseknert: Nit, and we can change it later, but better use `t.Output()` than `os.Stderr` as output in tests.

internal/lsp/server_command.go
```diff
@@ -524,3 +534,93 @@ func (l *LanguageServer) handleWorkspaceExecuteCommand(params types.ExecuteComma
...
+	}
+
+	if inputPath == "" || inputEmpty {
+		l.window.ShowMessage(
+			ctx,
+			types.InfoMessage,
+			"No input.json or input.yaml file found. Create one to provide test input data.",
+		)
+
+		return errors.New("no input.json found in workspace")
```
anderseknert: Won't this show up as two "dialogs" in the UI? One for ShowMessage and one for the error returned here? Better just return nil here if so.

internal/lsp/server_command_test.go
```diff
@@ -332,3 +332,123 @@ allow if {
...
+			}
+
+			if tc.expectFile && !fileExists {
+				t.Errorf("Expected test file to be created, but it wasn't")
+			} else if !tc.expectFile && fileExists {
+				t.Errorf("Expected no test file, but one was created")
+			}
+
+			if fileExists {
+				contents, _ := os.ReadFile(expectedTestFile)
```
anderseknert: Can use `must.ReadFile` here 🙂

internal/lsp/server_command_test.go
```diff
@@ -332,3 +332,123 @@ allow if {
...
+			expectedTestFile := filepath.Join(tempDir, "policy_test.rego")
+
+			timeout := time.NewTimer(5 * time.Second)
+			defer timeout.Stop()
+
+			var fileExists bool
+
+			if tc.expectFile {
+				for {
+					if _, statErr := os.Stat(expectedTestFile); statErr == nil {
```
anderseknert: Can use rio.Exists here and below

(from our own `io` package)

internal/lsp/server_command.go
SeanLedford: Ah yeah good catch, addressed!

PR #1983, opened 2026-05-05
regal/ast: add 'id' to valid metadata keys

srenatus:
It'll come in OPA v1.17.0: https://github.com/open-policy-agent/opa/pull/8606 🤞


bundle/regal/ast/comments.rego
anderseknert: Ideally we'd have a compiler that could concat the latter from the former at compile-time, but until then you'll want to add it here too :)

anderseknert (approved):

PR #1984, opened 2026-05-05
Introduce passthrough handler type

anderseknert:
And remove redundant Go types in the process. The idea of the passthrough handler type is essentially to avoid expensive conversions of input (params):

- `*json.RawMessage` -> `map[string]any` -> `Go struct` -> `ast.Value`

And output:

- `ast.Value` -> `map[string]any` -> `Go struct` -> `*json.RawMessage`

To instead do:

- `*json.RawMessage` -> `ast.Value`

and

- `ast.Value` -> `*json.RawMessage`

This makes these handlers **much** more efficient, at the cost of having a little less convenience when dealing with these payloads in Go. But that's what we're moving away from anyway!

All handlers will be converted to this model over time. The change was however large enough that I stopped here for now. We should also device a better test strategy for the handlers if we want to test them at a higher level than Rego, as working with byte slices isn't that convenient in tests. But I've left the few tests of the router as-is for now.

Also:

- Add new zero-alloc marshaller for AST JSON types, which efficiently encode/decode requests and responses in the language server without going via expensive roundtrips to Go types
- Remove Go types we no longer need, move their Go docs over to Rego
- Remove codeAction test written in Go now covered by Rego test, and move benchmark to router benchmarks
- Improve style used in codeAction tests
- Zero alloc implementations of `NewDefaultFixes` and `NewDefaultFormatterFixes`
- Add schemas for `data.client` and `data.server` objects


internal/lsp/rego/router.go
```diff
@@ -17,45 +20,49 @@ import (
...
 		"textDocument/codeLens":          make([]types.CodeLens, 0),
 		"textDocument/hover":             make([]types.Hover, 0),
 		"textDocument/signatureHelp":     nil,
 	}
+	errIgnored   = errors.New("ignored URI")
+	valueDecoder = encoding.OfValue()
+
+	inputValuePool = &sync.Pool{
+		New: func() any {
+			return new(ast.NewObject(
```
charlieegan3: is ast.NewObject not already a pointer?

internal/funsafe/funsafe.go
```diff
@@ -0,0 +1,50 @@
+package funsafe
```
charlieegan3, line 5: I feel like this packages warrants a comment about OPA internals dependency.

charlieegan3:
Just doing my review here but seem to get this error when testing in vscode:
 
```
[Error - 17:15:06] Request inlayHint/resolve failed.
  Message: failed querying "data.regal.lsp.main.eval": no results returned from evaluation
  Code: 0 
```

But doesn't seem to be 100% of the time.

internal/lsp/connection/connection.go
```diff
@@ -128,11 +127,10 @@ func logRequest(cfg LoggingConfig, req *jsonrpc2.Request) {
 		return
 	}
 
-	params, _ := encoding.JSON().Marshal(req.Params)
 	if req.Notif {
-		cfg.Logger.Message("--> notif: %s: %s\n", req.Method, params)
+		cfg.Logger.Message("--> notif: %s: %s\n", req.Method, req.Params)
```
charlieegan3: This seems to make the logs slightly different

before

> --> request #33: textDocument/documentSymbol: {"textDocument":{"uri":"file:///Users/charlieegan3/Code/regal/bundle/regal/ast/imports.rego"}}

after
> --> request #3: textDocument/documentSymbol: &{"textDocument":{"uri":"file:///Users/charlieegan3/Code/regal/bundle/regal/ast/imports.rego"}}

note the `&`

internal/funsafe/funsafe.go
SeanLedford: > I feel like this packages warrants a comment about OPA internals dependency.

Agreed, had to read up on this to grasp what was going on here.

internal/funsafe/funsafe_test.go
```diff
@@ -0,0 +1,22 @@
...
+
+func TestFunsafeObject(t *testing.T) {
+	t.Parallel()
+
+	obj := ast.NewObject(
+		ast.Item(ast.InternedTerm("foo"), ast.InternedTerm("bar")),
+		ast.Item(ast.InternedTerm("baz"), ast.InternedTerm("qux")),
+	)
+
+	if elm := funsafe.ObjectElems(obj); len(elm) != 2 {
```
SeanLedford, line 19: You mentioned testing the byte slices at the handler level, could we do it here instead, or until we have a better solution for that in mind? I do think we should have at least some test on the content of the mirrored object upfront.

SeanLedford (commented):
Code looks good! This looks like a nice optimization change, left a couple comments about clarity and testing, which I think will be especially important with parsing being handled like this.

internal/funsafe/funsafe.go
anderseknert: Fair enough! Adding 👍

internal/lsp/connection/connection.go
anderseknert: Ah, yes, I forgot the `*` to deref. Fixing, thanks!

internal/lsp/rego/router.go
anderseknert: It's an interface, but you're right — it wouldn't allocate when boxed to `any`. I'll simplify. Danke!

anderseknert:
@charlieegan3 good catch on the inlay hints! It turned out to be that the roundtrip we did in store.go was turning the client capabilities into RoAST format... 😅 I have fixed that now, and we should now have faster writes to the store as an added bonus.

internal/funsafe/funsafe_test.go
anderseknert: Indeed! I will follow up with another PR looking at testing specifically. I think you'll like that 😃

charlieegan3 (approved):

SeanLedford (approved):

PR #1988, opened 2026-05-11
regal/ast: update metadata fields

srenatus:
We've pivoted from `id` to `labels`. It's all within the release window so there's no reason to keep the `id` around.


anderseknert (approved):
LGTM!

PR #1992, opened 2026-05-11
New framework for testing language server handlers

anderseknert:
Or the Rego-based ones anyway.. but there's really nothing to stop us from reusing this elsewhere later. With Regal's language server moving away from a Go orchestration layer and towards passing through requests and responses straight to and from OPA wherever possible, our mostly Go-based tests of routing/handlers increasingly felt like a bad fit. But while most of our tests should be in Rego, we still need a solid suite of tests to target the server one layer above in the stack. Starting out with the Rego handlers, but again, the format should be easily reusable should we want that.

Since we need to move above Rego here, and "pure" Go isn't particularly nice for writing or reading tests dealing with `[]byte` inputs and outputs without marshalling into native Go types, this framework has test expressed as **markdown** documents. The rationale for this is mainly that I want all test data compiled in a single document, not a directory of policy + input + data + more. This makes it very easy to quickly see that given policy X and input Y, the response will be Z. I think it works very well here, but would be happy to hear thoughts from others.

Also:
- MUCH improved completion provider now able to optimize for clients (such as VS Code) capable of handling a default range for **all** items sent back in the response instead of repeating the same rather verbose range over and over. This effectively cuts the payload sent to the client in half, and results in much snappier completions.
- With that — finally seem to be able to set `isIncomplete` to `false` for at least the above mentioned clients, meaning that completion suggestions now normally only needs to be fetched once rather than at every key press. Another huge win!
- A few other improvements to slim down the size of completion items in order to avoid allocations
- Fix issue in hover provider that would provider hover tooltip for package keyword despite there being no example docs on that
- Various stability issues fixed as I encountered them while working on this.


internal/lsp/rego/router_test.go
```diff
@@ -1,199 +1,234 @@
...
+	return store
+}
 
-			sig := signatureHelp.Signatures[0]
+func handlerTests(tb testing.TB) iter.Seq2[string, testCase] {
+	tb.Helper()
+
+	return func(yield func(string, testCase) bool) {
+		dir := filepath.Join("testdata", "router")
+		for _, pattern := range []string{"**/**/*.md", "**/**/**/*.md"} {
```
charlieegan3, line 186: thought: if the tests are already at two levels, it is possible we might add more and forget to update the globs? perhaps just walk the dir and look for eligible files?

internal/lsp/rego/router_test.go
```diff
@@ -1,199 +1,234 @@
...
+		parsed  ast.Value
 	}
-}
+	jsonData struct {
+		name    string
+		content []byte
+	}
+	testCase struct {
+		method string
+		policy document
```
charlieegan3, line 47: This might be less relevant now, but we used to have a tests with many policy files for aggregate violations etc. it might be nice to be able to have that in future - or the option to add a config yaml too. Just good to know what the expansion plan is here.

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/go-git/go-git/v5 v5.18.0
 	github.com/go-viper/mapstructure/v2 v2.5.0
 	github.com/gobwas/glob v0.2.3
+	github.com/gomarkdown/markdown v0.0.0-20260417124207-7d523f7318df
```
charlieegan3, line 13: Might have been nicer if we had a separate binary to include this in since it's only for tests.

internal/lsp/semantictokens/semantictokens.go
```diff
@@ -98,3 +76,29 @@ func Full(result SemanticTokensResult) (*types.SemanticTokens, error) {
...
+	if raw, ok := result.(*json.RawMessage); ok {
+		var semTokRes SemanticTokensResult
+		// this looks like a false positive as the struct fields are tagged
+		// "the given struct should be annotated with the `json` tag"
+		//nolint: musttag
+		if err := json.Unmarshal(*raw, &semTokRes); err != nil {
+			return nil, err
+		}
+
+		full, err := Full(semTokRes)
```
charlieegan3, line 90: I guess a next step might be to have Full take the JSON and pass that to OPA without the need for the type too?

internal/lsp/rego/router.go
```diff
@@ -26,27 +25,25 @@ import (
 )
 
 var (
-	emptyResponse = map[string]any{
-		"textDocument/codeAction":        nil,
-		"textDocument/documentLink":      nil,
-		"textDocument/documentHighlight": nil,
-		"textDocument/documentSymbol":    make([]types.DocumentSymbol, 0),
```
charlieegan3, line 33: I might have misunderstood something but I think empty responses are always nil now - not arrays. Looks like this is still spec compliant but might be worth testing some other clients to confirm it's not breaking. I forget why we added these response types before.

internal/lsp/rego/router.go
```diff
@@ -360,61 +296,44 @@ func passthrough(ctx context.Context, rctx *RegalContext, req *jsonrpc2.Request)
...
 
-	defer func() {
-		inputValuePool.Put(inputValue)
-		ast.TermPtrPool.Put(paramsTerm)
-		ast.TermPtrPool.Put(regctxTerm)
-	}()
+	inputObj := cached.input.(ast.Object) //nolint:forcetypeassert
+	rast.Insert(inputObj, "method", ast.InternedTerm(req.Method))
+	rast.Insert(inputObj, "params", cached.params)
+	rast.Insert(inputObj, "regal", cached.regctx)
```
charlieegan3, line 313: Not a comment new for this PR, I am unsure when this specific use of ast.Object landed, but use of ast.Object in the server has created various race issues before - let's make sure we run the race detector a few times to check this is all still good here.

I use:

```
bash -c 'i=1; while go test -race ./... -count=3 -timeout 1m; do echo "Completed iteration $i"; i=$((i+1)); done'
```

internal/lsp/rego/router.go
```diff
@@ -360,61 +296,44 @@ func passthrough(ctx context.Context, rctx *RegalContext, req *jsonrpc2.Request)
...
 
-		if err := encoding.OfValue().Encode(buf, rsp); err != nil {
+		if err := encoding.OfValue().Encode(cached.buf, rsp); err != nil {
 			return nil, fmt.Errorf("failed to marshal response: %w", err)
 		}
 
-		raw := json.RawMessage(buf.Bytes())
+		raw := json.RawMessage(cached.buf.Bytes())
 
 		return &raw, nil
```
charlieegan3, line 331: Raw is based on the buf.Bytes, but not a copy, and then buf is returned to the pool at the end of the function. Should this be making a copy here?

charlieegan3 (approved):
I have only reviewed the code, some comments for considering and reply but nothing blocking.

internal/lsp/rego/router_test.go
anderseknert: If we extend this to run elsewhere, then yes! But these tests are structured like the LSP methods, e.g. `textDocumen/hover` or `textDocument/semanticTokens/full` .. and there are no more than 3 parts in any of those.

internal/lsp/rego/router_test.go
anderseknert: Totally. We'll want to extend this later if we think it's a nice way to run tests!

internal/lsp/semantictokens/semantictokens.go
anderseknert: That would be nice! I think the transformations it does would be very hard to do in Rego though..

internal/lsp/rego/router.go
anderseknert: Yeah, same. We definitely will want to go back to some client testing soon anyway, as it's been a year or more since we did!

PR #2002, opened 2026-05-20
Improved handling of input.(json|yaml) files

anderseknert:
**Important:**
- State of input files in a workspace now kept track of, and updates to them get parsed and cached whenever one is saved. Previously each eval/debug command invoked would involve both finding all potential input filess in the workspace, as well as to parse them *every time*, even when nothing changed. The language serer now instead includes a manager responsible for doing this only when needed.
- Parsed input data is now placed in the store rather than provided as part of `input`, making it easily available to any other handlers that may have a use for it, like e.g. the completion handlers.
- The server configuration now tells clients that it wants to be notified about input.(json|yaml) file updates in addition to `.rego` files. This allows us to track these via updates from the clients without having a watcher process running on the server as well.

**Less important:**
- Moved `initialized` handler to Rego. It doesn't do much at the moment, and we still need some Go code to persist the initial state of the server. But
- The input_dot_json completion handler was made simpler
- New `window` package providing a nicer abstraction for the various `window/*` calls the server may send the client
- Remove more Go types we no longer need

**What's left:**
- The input file generator that @SeanLedford built should be integrated into the new input manager

### DIFF
--- a/internal/lsp/input/input.go
+++ b/internal/lsp/input/input.go
@@ -1,0 +1,191 @@
+package input
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+
+	"go.yaml.in/yaml/v2"
+
+	"github.com/open-policy-agent/opa/v1/ast"
+	"github.com/open-policy-agent/opa/v1/storage"
+
+	"github.com/open-policy-agent/regal/internal/io/files"
+	"github.com/open-policy-agent/regal/internal/io/files/filter"
+	"github.com/open-policy-agent/regal/internal/lsp/log"
+	"github.com/open-policy-agent/regal/internal/lsp/store"
+	"github.com/open-policy-agent/regal/internal/lsp/uri"
+	"github.com/open-policy-agent/regal/internal/util"
+	"github.com/open-policy-agent/regal/pkg/roast/encoding"
+	"github.com/open-policy-agent/regal/pkg/roast/transform"
+)
+
+type (
+	// Manager manages input files in the workspace, allowing for fast retrieval of the most specific input for a given
+	// path, whether in Go (via [*Manager.FindForPath] and [*Manager.Get]) or in Rego (via `data.workspace.inputs`).
+	Manager struct {
+		rootPath string
+		rootFS   fs.FS
+		store    storage.Store
+		inputs   map[string]file
+		log      *log.Logger
+		mut      sync.RWMutex
+	}
+	file struct {
+		dir  string
+		raw  []byte
+		path storage.Path
+	}
+)
+
+func NewManager(store storage.Store, log *log.Logger) *Manager {
+	return &Manager{
+		store:  store,
+		inputs: make(map[string]file),
+		log:    log,
+		mut:    sync.RWMutex{},
+	}
+}
+
+func (m *Manager) LoadFromRoot(ctx context.Context, rootPath string) error {
+	root, err := os.OpenRoot(rootPath)
+	if err != nil {
+		return fmt.Errorf("failed to open workspace root: %w", err)
+	}
+
+	return m.LoadFromFS(ctx, rootPath, root.FS())
+}
+
+func (m *Manager) LoadFromFS(ctx context.Context, rootPath string, rootFS fs.FS) error {
+	m.mut.Lock()
+	m.rootPath = rootPath
+	m.rootFS = rootFS
+	m.mut.Unlock()
+
+	return files.DefaultWalker(".").
+		WithFilters(filter.Not(filter.Suffixes("input.json", "input.yaml"))).
+		WalkFS(rootFS, func(path string) error {
+			return m.Update(ctx, path, nil)
+		})
+}
+
+// FindForPath returns the most specific input path for the given path
+// relative to the workspace root, or an empty string if no input file is found.
+func (m *Manager) FindForPath(pathOrURI string) string {
+	path := m.internalPath(pathOrURI)
+
+	m.mut.RLock()
+	defer m.mut.RUnlock()
+
+	var closestDir, closestPath string
+	for inputPath, file := range m.inputs {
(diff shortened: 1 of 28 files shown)


internal/lsp/input/input.go
```diff
@@ -0,0 +1,187 @@
...
+	"github.com/open-policy-agent/regal/internal/lsp/log"
+	"github.com/open-policy-agent/regal/internal/lsp/store"
+	"github.com/open-policy-agent/regal/internal/lsp/uri"
+	"github.com/open-policy-agent/regal/internal/util"
+	"github.com/open-policy-agent/regal/pkg/roast/encoding"
+	"github.com/open-policy-agent/regal/pkg/roast/transform"
+)
+
+type (
+	Manager struct {
```
charlieegan3, line 31: I love this, but 'Manager' is quite generic, it might be good to add a comment to explain that this is tracking and caching input files in the workspace.

internal/lsp/window/window.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/sourcegraph/jsonrpc2"
+
+	"github.com/open-policy-agent/regal/internal/lsp/log"
+	"github.com/open-policy-agent/regal/internal/lsp/types"
+)
+
+// Window provides convenience methods for sending window-related notifications and requests to the client.
+// Errors are logged using the provided logger but they are *not* returned, as there is typically nothing
+// meaningful the server can do with them other than to log them, and handling them clutter the call sites.
+type Window struct {
```
charlieegan3, line 17: This is a really nice addition.

internal/lsp/server.go
```diff
@@ -1002,15 +997,22 @@ func (l *LanguageServer) handleTextDocumentDidChange(
...
-	if ignored := l.setMaybeIgnoredContents(params.TextDocument.URI, contents); !ignored {
-		parseSuccess, err := updateParse(ctx, l.parseOpts(params.TextDocument.URI, l.builtinsForCurrentCapabilities()))
-		if err != nil {
-			l.log.Message("failed to update module for %s: %s", params.TextDocument.URI, err)
-		} else if l.getClient().InitOptions.EnableServerTesting && parseSuccess {
-			l.testLocationJobs <- fileJob{URI: params.TextDocument.URI}
-		}
+	ignored := l.setMaybeIgnoredContents(params.TextDocument.URI, contents)
+	if !ignored {
+		// handle async as we want to acknowledge we handled the change to the client ASAP,
```
charlieegan3, line 1002: Is there an ordering risk there where a later edit overwritten by a slower earlier one?

internal/lsp/input/input.go
```diff
@@ -0,0 +1,187 @@
...
+			dir := strings.TrimSuffix(strings.TrimSuffix(strings.TrimSuffix(path, suffix), "input."), "/")
+
+			m.inputs[path] = file{raw: content, path: storePath, dir: dir}
+		}
+	}
+
+	return err
+}
+
+func (m *Manager) Delete(ctx context.Context, path string) error {
```
charlieegan3: Delete doesn't need `path := m.internalPath(pathOrURI)`?

internal/lsp/input/input.go
```diff
@@ -0,0 +1,187 @@
...
+		inputs: make(map[string]file),
+		log:    log,
+		mut:    sync.RWMutex{},
+	}
+}
+
+func (m *Manager) LoadFromRoot(ctx context.Context, rootPath string) {
+	root, err := os.OpenRoot(rootPath)
+	if err != nil {
+		panic(fmt.Sprintf("failed to open workspace root: %v", err))
```
charlieegan3: Might be good to leave a comment if this panic is still required.

internal/lsp/server.go
```diff
@@ -1487,7 +1481,15 @@ func (l *LanguageServer) loadWorkspaceContents(ctx context.Context, newOnly bool
 	return changedOrNewURIs, failed, nil
 }
 
-func (l *LanguageServer) handleInitialized(ctx context.Context) (any, error) {
+func (l *LanguageServer) initializedResultHandler(ctx context.Context, result any) (any, error) {
+	// If the client supports dynamic registration, register for any the Rego
+	// handler returned. Currently this is workspace/didChangeWatchedFiles only.
+	if raw, ok := result.(*json.RawMessage); ok && len(*raw) > 4 { // = len("null")
```
charlieegan3, line 1486: Would just == "null" be clearer?

internal/lsp/server.go
charlieegan3, line 1498: Not sure this should be the request context but a new one with a timeout?

charlieegan3 (approved):
I have mostly just done a high level pass of the new features, looks good to me but haven't tested myself locally. Will do so when it's on main before the release.

internal/lsp/input/input.go
anderseknert: It is only generic inside of this package though. Anywhere else it'll be referenced as `input.Manager` which was my intent :) I can add a comment still!

internal/lsp/server.go
anderseknert: I suppose so, although I'm not sure that having the client wait for parsing on each key press is how we should solve that. The change here is mainly done in order to improve completion handler responsiveness, and for that case it'll of course mostly be a case of having very few changes parse with success in the first place, and even less chance of having two quick subsequent edits both be parseable Rego.. and _even_ less risk of it now that completions are now mostly marked as `IsIncomplete: false` meaning the client won't send new requests upon further key presses...

So all in all, yes, but I think it's negligible at this point and avoiding having the client wait trumps that right now. We can definitely look into better ways to queue up and cancel jobs though.

internal/lsp/input/input.go
anderseknert: It probably does! Will fix, thanks.

internal/lsp/input/input.go
anderseknert: Yeah, I don't think we'll go far without one, but I can have the error bubble up 👍

PR #2004, opened 2026-05-20
OPA v1.16.2

anderseknert:
Also:
- Remove Codecov from CI as we haven't used it for a year or so and the less things we depend on the better
- Fix in Rego router to avoid potential reuse of a byte buffer that is still used elsewhere. Thanks @charlieegan3 for catching that!

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,14 +4,11 @@
 
 # Regal
 
-
 <!-- markdownlint-disable MD041 -->
 
 [![Build Status](https://github.com/open-policy-agent/regal/workflows/Build/badge.svg)](https://github.com/open-policy-agent/regal/actions)
-![OPA v1.16.1](https://www.openpolicyagent.org/badge/v1.16.1)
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -46,7 +41,6 @@ development, whether you're an experienced Rego developer or just starting out.
 - Provide advice on [best practices](https://www.openpolicyagent.org/docs/style-guide), coding style, and tooling
 - Allow users, teams and organizations to enforce custom rules on their policy code
 
-
```
charlieegan3, line 49: puzzled as to where these deletions from since the README is generated and the script has not changed.

charlieegan3 (approved):
LGTM, some test failures and one comment but good to go when sorted.

README.md
anderseknert: Whoa, I have no idea. This was a late edit but not _that_ late 😅 I'll fix.

anderseknert:
Closing this as I mistakenly believed the new `not` was in 1.16.2. Turns out it's not and we'll have to wait until v1.17.0 next week.

README.md
charlieegan3: The best code works so well you never knew it existed!

PR #2005, opened 2026-05-21
fix(rules): unassigned-return-value now covers namespaced builtins

### LINKED ISSUE #1999: Missed case in unassigned-return-value rule
```rego
r1 if json.match_schema({"foo": 1}, {
	"type": "object",
	"properties": {"Foo": {"type": "number"}},
	"additionalProperties": false,
})

r2 if lower(input.foo)
```

Neither built-in functions return a boolean value, but only the `lower` call is flagged here. The r1 case is an easy mistake to make, so we should make sure that's covered.

mvanhorn:
## Summary

Fix #1999: `unassigned-return-value` now flags namespaced builtins like `json.match_schema(...)` in addition to flat builtins like `lower(...)`.

## Why this matters

The reporter showed two policies that should both produce an `unassigned-return-value` violation:

```rego
r1 if json.match_schema({"foo": 1}, { "type": "object", "properties": { "Foo": {"type": "number"} } })
r2 if lower(input.foo)
```

Today the linter flags only `r2`. The root cause is in
`bundle/regal/rules/bugs/unassigned-return-value/unassigned_return_value.rego`:

```rego
terms[0].value[0].type == "var"
ref_name := terms[0].value[0].value
ref_name in ast.builtin_names
```

For a flat builtin like `lower`, `terms[0].value` is a single `var` term and `ref_name` becomes `"lower"`, which matches `ast.builtin_names`. For a namespaced builtin like `json.match_schema`, `terms[0].value` is `[var("json"), string("match_schema")]` and `ref_name` becomes `"json"`. `ast.builtin_names` (built from `object.keys(config.capabilities.builtins)` in `bundle/regal/ast/ast.rego:59`) is keyed on the full dotted name (`"json.match_schema"`, `"http.send"`, ...), so the lookup misses.

`bundle/regal/ast/ast.rego:245` already exposes `ref_to_string(ref)` which concatenates the ref terms into the dotted form. Swapping the single-segment lookup for `ast.ref_to_string(terms[0].value)` lets the rule match namespaced builtins without touching the downstream checks (`config.capabilities.builtins[ref_name].decl.result` and `ast.function_ret_in_args(ref_name, terms)` already key on the full name).

The `terms[0].value[0].type == "var"` guard is dropped because the same shape constraint is implied by `ref_to_string` producing a usable name — refs with a non-var first segment will not appear in `builtin_names` and short-circuit the rule the same way.

## Testing

- Added `test_fail_unused_return_value_namespaced` covering `json.match_schema({"foo": 1}, {})` as a naked expression — the rule now reports one violation.
- Added `test_success_namespaced_assigned` covering `x := json.match_schema({"foo": 1}, {})` — assignment continues to suppress the violation.
- `regal test bundle/regal/rules/bugs/unassigned-return-value/` reports `PASS: 8/8`.

Closes #1999

### DIFF
--- a/bundle/regal/rules/bugs/unassigned-return-value/unassigned_return_value.rego
+++ b/bundle/regal/rules/bugs/unassigned-return-value/unassigned_return_value.rego
@@ -14,6 +14,7 @@ report contains violation if {
 	terms := ast.found.expressions[_][expr].terms
 
 	not expr.interpolated
+	not expr.negated
 
 	terms[0].type == "ref"
 
--- a/bundle/regal/rules/bugs/unassigned-return-value/unassigned_return_value.rego
+++ b/bundle/regal/rules/bugs/unassigned-return-value/unassigned_return_value.rego
@@ -16,9 +16,8 @@ report contains violation if {
 	not expr.interpolated
 
 	terms[0].type == "ref"
-	terms[0].value[0].type == "var"
 
-	ref_name := terms[0].value[0].value
+	ref_name := ast.ref_to_string(terms[0].value)
 	ref_name in ast.builtin_names
 
 	# special case as the "result" of print is ""
--- a/bundle/regal/rules/bugs/unassigned-return-value/unassigned_return_value.rego
+++ b/bundle/regal/rules/bugs/unassigned-return-value/unassigned_return_value.rego
@@ -25,13 +25,6 @@ report contains violation if {
 
 	config.capabilities.builtins[ref_name].decl.result != "boolean"
 
-	# Skip builtins whose declared result is "any" (e.g. object.subset,
-	# object.union_n) - those values are commonly used as boolean
-	# predicates in `not` expressions, so flagging them produces false
-	# positives. The reporter's case (json.match_schema, declared "array")
-	# is still caught.
-	config.capabilities.builtins[ref_name].decl.result != "any"
-
 	# no violation if the return value is declared as the last function argument
 	# see the function-arg-return rule for *that* violation
 	not ast.function_ret_in_args(ref_name, terms)
--- a/bundle/regal/rules/bugs/unassigned-return-value/unassigned_return_value.rego
+++ b/bundle/regal/rules/bugs/unassigned-return-value/unassigned_return_value.rego
@@ -25,6 +25,13 @@ report contains violation if {
 
 	config.capabilities.builtins[ref_name].decl.result != "boolean"
 
+	# Skip builtins whose declared result is "any" (e.g. object.subset,
+	# object.union_n) - those values are commonly used as boolean
+	# predicates in `not` expressions, so flagging them produces false
+	# positives. The reporter's case (json.match_schema, declared "array")
+	# is still caught.
+	config.capabilities.builtins[ref_name].decl.result != "any"
+
 	# no violation if the return value is declared as the last function argument
 	# see the function-arg-return rule for *that* violation
 	not ast.function_ret_in_args(ref_name, terms)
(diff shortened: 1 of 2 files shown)


bundle/regal/rules/bugs/unassigned-return-value/unassigned_return_value.rego
```diff
@@ -16,16 +16,22 @@ report contains violation if {
...
 	# special case as the "result" of print is ""
 	ref_name != "print"
 
 	config.capabilities.builtins[ref_name].decl.result != "boolean"
 
+	# Skip builtins whose declared result is "any" (e.g. object.subset,
+	# object.union_n) - those values are commonly used as boolean
+	# predicates in `not` expressions, so flagging them produces false
+	# positives. The reporter's case (json.match_schema, declared "array")
+	# is still caught.
```
anderseknert: I don't think this holds true for anything but possibly `object.subset`. It's a good point that the signature of some of OPA's built-in functions should be more specific than defining `any` as the return type though, and I'll raise an issue for this in OPA. But let's remove this check here for now.

bundle/regal/rules/bugs/unassigned-return-value/unassigned_return_value_test.rego
```diff
@@ -64,6 +64,16 @@ test_fail_unused_return_value_nested if {
 	}}
 }
 
+test_fail_unused_return_value_namespaced if {
+	r := rule.report
+		with input as ast.with_rego_v1(`allow if {
```
anderseknert: Nit, but prefer ast.policy over ast.with_rego_v1 as the `rego.v1` import has not been required for over a year.

anderseknert (commented):
Thanks! Looks good to me with some minor adjustments suggested.

mvanhorn:
Addressed in 99da33e4. Removed the any-result skip in the rule and swapped the four `ast.with_rego_v1` helpers to `ast.policy` in the test file. Expected location rows shifted up by 2 since `ast.policy` does not inject the package + `import rego.v1` boilerplate -- updated those assertions to match.

Verification:
- `go run main.go test bundle` -- PASS 1019/1019
- `go run main.go lint --format compact bundle` now surfaces two pre-existing `object.subset` uses inside `not` predicates (the false-positive pattern you mentioned):
  - `bundle/regal/rules/idiomatic/use-array-flatten/use_array_flatten.rego:17`
  - `bundle/regal/rules/idiomatic/use-object-union-n/use_object_union_n.rego:17`

Left those out of this PR since you scoped the change to removing the check. Happy to follow up with either explicit assignments at those sites or a `# regal ignore:unassigned-return-value` annotation if you prefer to keep the bundle clean while the upstream OPA signature fix lands.

anderseknert (approved):
Thanks! 🙏

anderseknert:
Hah! Looks like the build now is failing precisely because we use `not object.subset(...)` in 2 locations and now fail when linting our own codebase 😄 If you handle that as a special case for now, I'll follow up later with a closer look at what other built-in functions we need to account for in OPA and here. Thanks!

anderseknert:
@mvanhorn issue created in OPA https://github.com/open-policy-agent/opa/issues/8692

mvanhorn:
Addressed in 7bff9df. Added `not expr.negated` to the rule so `not <builtin>(...)` falls through (the return value IS consumed, as the truthiness operand of `not`). Regression test `test_success_negated_builtin_return_value` covers it. Local verification: `./regal lint --format github bundle` clean across 339 files, rule selection 9/9, full Rego bundle 1020/1020.

Referenced OPA#8692 in the commit body so the broader follow-up stays traceable.

anderseknert (approved):
Thank you!

mvanhorn:
Thanks @anderseknert for merging the unassigned-return-value fix so it now covers namespaced builtins like json.

mvanhorn:
Thanks @anderseknert. Good to have `unassigned-return-value` catching namespaced builtins like `json.match_schema` now, not just the flat ones.

PR #2009, opened 2026-05-26
lsp: Introduce workspace abstraction

anderseknert:
Much like how an input manager recently got added to handle all things related to input, this PR introduces a new workspace package to handle, well, the workspace. Obviously there is still a *lot* of workspace related things going on outside of this handler, but it's a good start handling the most basic of things, like creating paths and URIs extending the workspace root. Also made a convenient implementation of workspace/applyEdit for us to extend further.

A similar, although nowhere near as extensive, abstraction has also be added for LSP clients.

As usual some other fixes here and there, but I have tried my best to limit the scope this time :)


internal/lsp/server.go
```diff
@@ -743,43 +722,36 @@ func (l *LanguageServer) processTemplateJob(_ context.Context, job fileJob) {
...
 
-	// combine content edits with any additional rename edits
-	edits := append(make([]any, 0, 1+len(additionalRenameEdits.Edit.DocumentChanges)), types.TextDocumentEdit{
-		TextDocument: types.OptionalVersionedTextDocumentIdentifier{URI: job.URI},
-		Edits:        ComputeEdits("", newContents),
-	})
-	edits = append(edits, additionalRenameEdits.Edit.DocumentChanges...)
+	// combine content changes with any additional rename changes
+	edit := types.NewTextDocumentEdit(job.URI, ComputeEdits("", newContents))
+	l.log.Message(string(edit.AppendJSON([]byte{})))
```
charlieegan3: Looks like a debug log left behind.

internal/lsp/types/types.go
```diff
@@ -298,3 +313,129 @@ type (
...
+		bs = r.Options.AppendJSON(append(bs, `,"options":`...))
+	}
+
+	return append(bs, '}')
+}
+
+func (ro RenameFileOptions) AppendJSON(bs []byte) []byte {
+	bs = append(bs, '{')
+	if ro.IgnoreIfExists {
+		if bs = append(bs, `"ignoreIfNotExists":true`...); ro.Overwrite {
```
charlieegan3:
```suggestion
		if bs = append(bs, `"ignoreIfExists":true`...); ro.Overwrite {
```

cmd/languageserver.go
```diff
@@ -47,11 +47,7 @@ func init() {
 				logLevel = log.LevelDebug
 			}
 
-			opts := &lsp.LanguageServerOptions{
-				Logger:       log.NewLogger(logLevel, os.Stderr),
-				FeatureFlags: lsp.DefaultServerFeatureFlags(),
```
charlieegan3, line 52: These are not needed for some reason?

internal/lsp/types/types.go
```diff
@@ -298,3 +313,129 @@ type (
...
+
+	if ro.Overwrite {
+		bs = append(bs, `"overwrite":true`...)
+	}
+
+	return append(bs, '}')
+}
+
+func (d DeleteFile) AppendJSON(bs []byte) []byte {
+	bs = append(append(append(bs, `{"kind":"delete","uri":"`...), d.URI...), '"')
```
charlieegan3: Do we know the URI is safe to insert in this way? e.g. \\" etc.

internal/lsp/workspace/workspace.go
```diff
@@ -0,0 +1,161 @@
...
+	return filepath.Join(slices.Insert(join, 0, w.path)...)
+}
+
+// RelativePath returns the path of the given file URI relative to the workspace root.
+func (w Workspace) RelativePath(fileURI string) string {
+	return uri.ToRelativePath(fileURI, w.uri)
+}
+
+// PathToURI converts a file path (absolute or relative to the workspace root) to a URI for this workspace's client.
+func (w Workspace) PathToURI(path string) string {
```
charlieegan3: We to have a util for these elsewhere in the server. Perhaps this could share that?

internal/lsp/client/client.go
```diff
@@ -0,0 +1,74 @@
+//nolint:recvcheck
```
charlieegan3, line 1: is this not just for UnmarshalJSON?

internal/lsp/workspace/workspace.go
```diff
@@ -0,0 +1,161 @@
...
+	"slices"
+	"strconv"
+	"time"
+
+	"github.com/open-policy-agent/regal/internal/lsp/client"
+	"github.com/open-policy-agent/regal/internal/lsp/uri"
+)
+
+type (
+	Workspace struct {
```
charlieegan3, line 20: This stutters in the name, I wonder if this is more descriptive as State, or Root?

internal/lsp/client/client.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/open-policy-agent/regal/internal/lsp/clients"
+	"github.com/open-policy-agent/regal/internal/lsp/types"
+	"github.com/open-policy-agent/regal/internal/lsp/uri"
+	"github.com/open-policy-agent/regal/internal/roast/transforms"
+	"github.com/open-policy-agent/regal/internal/util"
+	"github.com/open-policy-agent/regal/pkg/roast/encoding"
+)
+
+type Client struct {
```
charlieegan3, line 20: I know the package is `lsp/client`, but this feels strange to me that we have a client that is not actually a client to make calls to a server. Somewhat similar to workspace, I'd prefer this to be called something a bit closer to what it is, or in a package path that makes that clearer.

Two ideas:

* Is there a case also to be made that this should be a field on workspace? Then we only need to name one thing correctly?
* If we like having them broken down, perhaps workspace and client packages could be moved under internal/lsp/session/ or something so it's clear they are server session data.

internal/lsp/server.go
anderseknert: Oh, indeed, thanks! I'll fix that.

cmd/languageserver.go
anderseknert: Those are indeed not needed https://github.com/open-policy-agent/regal/blob/main/internal/lsp/server.go#L209

internal/lsp/client/client.go
anderseknert: Yes

internal/lsp/client/client.go
anderseknert: I think this is fundamentally a question of whether naming should follow the "domain model" or more technical things, like architecture. I would normally default to the latter, but in the context of a fairly rigid specification, I have no problems having a "client" mirror what the spec describes rather than properties like "something that makes calls to a server". In fact, I much prefer it :)

> Is there a case also to be made that this should be a field on workspace?

It is a field on the workspace. Or do you mean something else here? 🤔

internal/lsp/types/types.go
anderseknert: Oops! Copy-pasta from the delete options, I think. I'll fix, thanks!

internal/lsp/types/types.go
anderseknert: Nope, looks like I missed strconv.AppendQuote here and in a few more places. Will fix 👍

internal/lsp/workspace/workspace.go
anderseknert: I think that's unavoidable to some degree, and I do think Workspace is the more descriptive / accurate name here. See e.g. the `ring.Ring` mention here https://go.dev/doc/effective_go#package-names

I definitely try to avoid it when that's not the case though!

internal/lsp/workspace/workspace.go
anderseknert: Can you elaborate a little on this? 🙂

internal/lsp/input/input.go
```diff
@@ -52,24 +59,16 @@ func NewManager(store storage.Store, log *log.Logger) *Manager {
...
-func (m *Manager) LoadFromFS(ctx context.Context, rootPath string, rootFS fs.FS) error {
+func (m *Manager) LoadFromWorkspace(ctx context.Context, workspace workspace.Workspace) {
 	m.mut.Lock()
-	m.rootPath = rootPath
-	m.rootFS = rootFS
+	m.workspace = workspace
 	m.mut.Unlock()
 
-	return files.DefaultWalker(".").
+	_ = files.DefaultWalker(".").
```
SeanLedford: Noticed we're no longer surfacing an error here, curious if we're surfacing this elsewhere to be logged?

SeanLedford (commented):
Looks good on first pass! Only one nit that I've found so far. Will Let Charlie stamp this one whenever he's ready to.

internal/lsp/input/input.go
anderseknert: It's not a critical error, but you're right, we should still log it. Fixed.

internal/lsp/workspace/workspace.go
charlieegan3: Sure, just that we have https://github.com/open-policy-agent/regal/blob/main/internal/lsp/uri/uri.go#L62 and https://github.com/open-policy-agent/regal/blob/main/internal/lsp/uri/uri.go#L21

internal/lsp/client/client.go
charlieegan3: In that case, it might be best to put on that function.

internal/lsp/client/client.go
charlieegan3: Sorry, I just meant that it could be wrapped more in the workspace type rather than out on its own.

internal/lsp/workspace/workspace.go
anderseknert: Ah this remained from an earlier draft and wasn't used. Removed! Thanks :)

charlieegan3 (approved):

PR #2011, opened 2026-05-28
cmd: Remove git validation from fix command

charlieegan3:
The git repository checking that required the --force flag has been removed. This has been removed as it's understood that most Rego is either in a git repo, or recoverable in some other way if needed.

Other CLI tools commonly do not back up or check git for such operations, e.g. opa fmt.

Doing this we eliminate the go-git dependency.

The --force flag is marked as deprecated for backwards compatibility.


anderseknert (approved):
Awesome!

PR #2013, opened 2026-05-28
Extract file polling into waitForFile helper

charlieegan3:
Wrap the file exist polling logic with timeout into a reusable helper function to address break in select bug.


anderseknert (approved):
So much better!

PR #2015, opened 2026-05-28
build: Add capabilities verification to build check

charlieegan3:
Add check_capabilities job to pull_request workflow to verify build/capabilities.json is up-to-date.

Checks similar to what we do for readme.


srenatus (approved):

build/update-capabilities.sh
srenatus, line 1: I suppose we could try to put more of the logic into the rq script, but this is fine

PR #2016, opened 2026-06-02
Unify common aggregation and aggregated data

anderseknert:
This PR unifies aggregation of data and some aggregated reports in a centralized manner, allowing multiple aggregate rules to share data and reported aggregates between them.

Included here is also a new aggregate that collects a rule tree from each linted package, similarly to how the OPA compiler does it. Having a rule tree aggregated from all modules means that we now also easily can extract e.g. all the tests from the module, greatly simplifying how these are collected. Or implement the references feature from the language server spec. And so on! Querying the rule tree for a particular node is as easy as calling `object.get`, and for more advanced queries we can `walk`.

We'll also need to aggregate location data for more advanced things, but since we're preparing for a release, I'll leave it here until v0.41.0 is out!

### DIFF
--- a/bundle/regal/aggregators/aggregators.rego
+++ b/bundle/regal/aggregators/aggregators.rego
@@ -1,0 +1,63 @@
+package regal.aggregators
+
+import future.keywords.not
+
+import data.regal.ast
+import data.regal.util
+
+# METADATA
+# description: aggregates a rule tree for the given input module
+rule_tree := tree if {
+	pkg_name := replace(replace(ast.package_name_full, `["`, "."), `"]`, "")
+	pkg_path := split(pkg_name, ".")
+
+	pkg_obj := json.patch({}, [patch |
+		some i in numbers.range(1, count(pkg_path))
+
+		patch := {
+			"op": "add",
+			"path": $"/{concat("/", array.slice(pkg_path, 0, i))}",
+			"value": {},
+		}
+	])
+
+	tree := json.patch(pkg_obj, [patch |
+		some rule_path in _rule_paths
+
+		patch := {
+			"op": "add",
+			"path": $"/{concat("/", array.concat(pkg_path, rule_path))}",
+			"value": {},
+		}
+	])
+}
+
+# METADATA
+# description: |
+#   aggregates all input refs and their locations for the given input module§
+imports := [imp |
+	some _import in input.imports
+
+	_import.path.value[0].value == "data"
+	len := count(_import.path.value)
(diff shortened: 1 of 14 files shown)


bundle/regal/rules/imports/unresolved-reference/unresolved_reference.rego
```diff
@@ -53,101 +59,50 @@ _shadowed_imports contains var_name if {
...
-	vals := split(loc, ":")
-
-	row := to_number(vals[0])
-	col := to_number(vals[1])
-
-	from_col := substring(text, col - 1, -1)
-	ref_text := substring(from_col, 0, indexof(from_col, " "))
-
-	end_col := col + count(ref_text)
+	object.get(rule_tree, array.slice(ref_path, 0, i), false) == {} # regal ignore:superfluous-object-get
```
charlieegan3, line 105: I wonder if there is a better value for 'leaf' other than `{}`, it's a little obscure.

bundle/regal/aggregators/aggregators.rego
```diff
@@ -0,0 +1,63 @@
...
+	pkg_name := replace(replace(ast.package_name_full, `["`, "."), `"]`, "")
+	pkg_path := split(pkg_name, ".")
+
+	pkg_obj := json.patch({}, [patch |
+		some i in numbers.range(1, count(pkg_path))
+
+		patch := {
+			"op": "add",
+			"path": $"/{concat("/", array.slice(pkg_path, 0, i))}",
+			"value": {},
```
charlieegan3, line 20: See other comment about `{}` meaning leaf.

charlieegan3 (approved):

bundle/regal/rules/imports/unresolved-reference/unresolved_reference.rego
anderseknert: Indeed! However, when we build the tree we don't necessarily know whether a node is a branch or a leaf, and `json.patch` can only `add` keys to existing objects... so this is easier said than done, I'm afraid. 

In the next iteration, I'm planning to look into having a later step replace the leaf nodes (i.e. rules) with arrays containing the locations where they're defined. But that too is... complicated, as `object.union_n` will have array values overwritten from one object merged into another, while we would want the values to be appended (when one rule is defined across several modules), and Rego famously "doesn't need a reduce feature" ;) So yeah, coming back to this, but opted for the model to work first for what we have and need today.

PR #2018, opened 2026-06-02
debug: Determine abs debugging path for input.json

charlieegan3:
Addresses regression from https://github.com/open-policy-agent/opa/pull/8716/changes#diff-9bb0f472082191b5902d6089ff2d4240207cc7632bcbba6ea8f90099aa1d52b9R58

Still works when input.json is missing and thus "".

### DIFF
--- a/internal/lsp/server_command.go
+++ b/internal/lsp/server_command.go
@@ -133,6 +133,13 @@ func (l *LanguageServer) StartCommandWorker(ctx context.Context) {
 						break
 					}
 
+					// FindForPath returns a workspace-relative path (or ""); the OPA debugger
+					// resolves inputPath via os.Open against its own CWD, so pass an absolute path.
+					var inputPath string
+					if rel := l.input.FindForPath(args.Target); rel != "" {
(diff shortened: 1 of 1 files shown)


internal/lsp/server_command.go
```diff
@@ -133,6 +133,13 @@ func (l *LanguageServer) StartCommandWorker(ctx context.Context) {
 						break
 					}
 
+					// FindForPath returns a workspace-relative path (or ""); the OPA debugger
+					// resolves inputPath via os.Open against its own CWD, so pass an absolute path.
```
anderseknert, line 137: > the OPA debugger resolves inputPath via os.Open

We should fix that! But all in due course :) This is a nice workaround in the meantime.

anderseknert (approved):

internal/lsp/server_command.go
charlieegan3: Wrote up the issue here: https://github.com/open-policy-agent/regal/issues/2020, the real issue is that we have the client and regal both overlapping a bit.

PR #2019, opened 2026-06-02
Use common aggregate data in circular-import

anderseknert:
Just a tiny follow-up on the last change.

This does not block next release in any way.


SeanLedford (approved):
LGTM!

PR #2024, opened 2026-06-04
perf: remove a million allocations with this one simple trick

anderseknert:
The OPA maintainer's hate him for it!

(there are _many_ more locations where we can do this, but I'd rather not have to clutter the code with voodoo, so this covers only the most costly allocs, of which ast.found.expressions is the worst by far)

Also getting rid of some dead Go code.

### DIFF
--- a/bundle/regal/ast/search.rego
+++ b/bundle/regal/ast/search.rego
@@ -137,7 +137,7 @@ found.vars[rule_index].term contains var if {
 	fn_name := ref_static_to_string(call[0].value)
 	undeclared_start := count(all_functions[fn_name].decl.args) + 1
 
-	call[undeclared_start]
+	_ = call[undeclared_start]
 	fn_name != "print"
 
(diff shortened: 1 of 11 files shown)


SeanLedford (approved):
Looks good!

bundle/regal/ast/search.rego
```diff
@@ -236,7 +236,7 @@ found.expressions[rule_index] contains value if {
 
 	walk(rule[node], [_, value])
 
-	value.terms
+	_ = value.terms
```
charlieegan3, line 239: I wonder if one of these warrants a comment to avoid cleaning up in future by mistake?

charlieegan3 (commented):
Looks good to me, nice work on the dead code too.

anderseknert:
Ah, that reminds me! Finding dead code is another thing that'll be trivial once we have a robust system in place for tracking definitions and references 😃

PR #2026, opened 2026-06-07
feat: add performance rule repeated-computation

### LINKED ISSUE #1820: Rule: `repeated-computation`
Found an example of this in our own code, and there are probably a few more cases scattered around. Simplified, it would be neat if Regal identified duplicate computations in a given scope, and that something like this:

```rego
r := count(input.s) if {
	l := count(input.s)
	l > 10
}
```

...would better be expressed as this, where repeating same computation (`count(input.s)`) is avoided:

```rego
r := l if {
	l := count(input.s)
	l > 10
}
```

Even for a trivial case like the above, the performance difference is clear — and could be substantial with more advanced computations or repetitions involved.
```
┌───────────────────────────────────────────┬────────┐
│ samples                                   │ 186446 │
│ ns/op                                     │   6168 │
│ B/op                                      │   4887 │
│ allocs/op                                 │     77 │
└───────────────────────────────────────────┴────────┘
vs.
┌───────────────────────────────────────────┬────────┐
│ samples                                   │ 206259 │
│ ns/op                                     │   5640 │
│ B/op                                      │   4726 │
│ allocs/op                                 │     71 │
└───────────────────────────────────────────┴────────┘
```
"In a given scope" is of course important here, as the value of vars can be shadowed in e.g. a comprehension, meaning that `count(x)` isn't necessarily a repeated computation just because it looks like one. 

As an added bonus, assigning the result of at least more advanced, chained computations to variables will also help improve the readability of a policies.

mvanhorn:
## Summary

Adds the `performance/repeated-computation` rule requested by anderseknert in #1820. The rule flags repeated identical built-in calls (like `count(input.items)`) evaluated more than once in the same scope, suggesting the result be assigned to a variable instead. Ships with the standard rule bundle wiring: provided-config entry, rego unit tests, docs page, and the e2e most-violations fixture.

## Why this matters

#1820 is a maintainer-filed rule request: repeated computation of the same expression in one scope is wasted evaluation work that policy authors rarely notice, and Regal's performance category is the right home for it. The implementation follows the existing performance-rule patterns (`with-outer-binding` style AST walks), normalizes locations before comparing call sites so equivalent calls match regardless of position, and reports each repeated location with the standard ranged result shape.

## Testing

`go build ./...` clean; `regal test bundle` passes 1031/1031 including the 12 new unit tests covering positive matches (same scope, multiple rules), negatives (different args, different scopes, single use), and location text expectations. Docs page added at `docs/rules/performance/repeated-computation.md`; the e2e violations fixture includes the new rule.

Fixes #1820

### DIFF
--- a/.regal/config.yaml
+++ b/.regal/config.yaml
@@ -132,6 +132,10 @@ rules:
       ignore:
         files:
           - "*_test.rego"
+    repeated-computation:
+      ignore:
+        files:
+          - "*_test.rego"
   style:
     line-length:
       level: error
--- a/bundle/regal/config/provided/data.yaml
+++ b/bundle/regal/config/provided/data.yaml
@@ -161,6 +161,8 @@ rules:
       level: ignore
     non-loop-expression:
       level: error
+    repeated-computation:
+      level: error
     walk-no-path:
       level: error
     with-outside-test-context:
--- a/bundle/regal/rules/performance/repeated-computation/repeated_computation.rego
+++ b/bundle/regal/rules/performance/repeated-computation/repeated_computation.rego
@@ -1,0 +1,115 @@
+# METADATA
+# description: Repeated computation
+# related_resources:
+#   - description: documentation
+#     ref: https://www.openpolicyagent.org/projects/regal/rules/performance/repeated-computation
+package regal.rules.performance["repeated-computation"]
+
+import data.regal.ast
+import data.regal.config
+import data.regal.result
+import data.regal.util
+
+report contains violation if {
+	some rule_index
+	call := _calls[rule_index][_]
+
+	some prior in _calls[rule_index]
+	call.key == prior.key
+	_before(prior.location, call.location)
+
+	violation := result.fail(rego.metadata.chain(), result.ranged_location_between(call.term[0], regal.last(call.term)))
+}
+
+_calls[rule_index] contains call_info if {
+	some rule_index
+	call := ast.found.calls[rule_index][_]
+
+	name := ast.ref_to_string(call[0].value)
+	name in ast.builtin_names
+	not _excluded_builtin(name)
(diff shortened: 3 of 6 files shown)


anderseknert:
@mvanhorn sorry for not having looked into this yet. I was hoping you'd address the linter issues reported first :) Would you want to take a look at those? Thanks!

mvanhorn:
Cleared the linter issues in 585191f - switched to `some ... in` iteration and direct field access, used `json.patch` in the test, and set the rule to ignore `*_test.rego`. Should be green now; happy to look again if the matrix still flags anything.

anderseknert (approved):
Sorry for the *long* delay on this one @mvanhorn. I had some medical issues to tend to, but now back to better issues :)

anderseknert:
Seems like I can't squash this as we're using a merge queue. @mvanhorn could you please squash this into a single commit so that we can have it merged? Thanks!

mvanhorn:
Squashed to a single commit in 517144b, rebased on main. Same diff as before, 6 files. `regal test` on the rule is 9/9.

anderseknert:
@mvanhorn much appreciated. It looks like your sign-off got lost in the squash though. Hate to bother you about that but can't merge without one. Make sure that it matches your commit email too :) Thanks again.

mvanhorn:
Sorted - 1a953a5 carries the sign-off now, and it matches the commit author email. DCO is green.

Sorry for the round trip, the trailer did not survive the squash.

anderseknert:
Thanks, Matt!

mvanhorn:
Thanks for the merge, @anderseknert, and for proposing the rule in #1820. Flagging repeated built-in calls in the same scope turned out to be a nice small win.

PR #2029, opened 2026-06-08
lsp: first implementation of find references and rename

anderseknert:
So far only covers function arguments, but more will be added soon!


bundle/regal/ast/ast.rego
```diff
@@ -261,6 +261,19 @@ _format_term(term) := concat("", [".", term.value]) if {
 	[term.value],
 )
 
+# METADATA
+# description: |
+#   returns the static prefic of a terms array
```
SeanLedford, line 266: Minor spelling mistake 😱

SeanLedford (approved):

PR #2030, opened 2026-06-15
docs: Small updates to address some inconsistencies

charlieegan3:
Likely easiest to review commit by commit as all edits within a commit are on a theme.

### DIFF
--- a/bundle/regal/rules/performance/non-loop-expression/non_loop_expression.rego
+++ b/bundle/regal/rules/performance/non-loop-expression/non_loop_expression.rego
@@ -1,5 +1,5 @@
 # METADATA
-# description: Non-loop expression
+# description: Non loop expression in loop
 # related_resources:
 #   - description: documentation
 #     ref: https://www.openpolicyagent.org/projects/regal/rules/performance/non-loop-expression
--- a/bundle/regal/rules/performance/non-loop-expression/non_loop_expression.rego
+++ b/bundle/regal/rules/performance/non-loop-expression/non_loop_expression.rego
@@ -1,5 +1,5 @@
 # METADATA
-# description: Non loop expression in loop
+# description: Non-loop expression in loop
 # related_resources:
 #   - description: documentation
 #     ref: https://www.openpolicyagent.org/projects/regal/rules/performance/non-loop-expression
--- a/bundle/regal/rules/performance/with-outside-test-context/with_outside_test_context.rego
+++ b/bundle/regal/rules/performance/with-outside-test-context/with_outside_test_context.rego
@@ -1,5 +1,5 @@
 # METADATA
-# description: '`with` used outside test context'
+# description: '`with` used outside of test context'
 # related_resources:
 #   - description: documentation
 #     ref: https://www.openpolicyagent.org/projects/regal/rules/performance/with-outside-test-context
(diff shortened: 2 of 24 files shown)


bundle/regal/rules/performance/non-loop-expression/non_loop_expression.rego
```diff
@@ -1,5 +1,5 @@
 # METADATA
-# description: Non-loop expression
+# description: Non loop expression in loop
```
anderseknert: Out of curiosity, why no hyphen here? You obviously know English better than me, but some cursory searches seem to suggest that the hyphen is the correct form 🤓 

I think we "in iteration" might be better too to avoid repeating the same word twice (and that we technically have no "loops", but the rule was named before someone lectured me on that, lol). It's a good addition though.

bundle/regal/rules/performance/with-outside-test-context/with_outside_test_context.rego
```diff
@@ -1,5 +1,5 @@
 # METADATA
-# description: '`with` used outside test context'
+# description: '`with` used outside of test context'
```
anderseknert, line 2: Funny, as "of" here and a few other places used to be flagged by the language checker, and so I went with that. I too would use "of" here, but I wonder if it's some difference between written/formal/spoken language 🤔 Happy to use "of" here either way!

bundle/regal/rules/style/yoda-condition/yoda_condition.rego
```diff
@@ -1,5 +1,5 @@
 # METADATA
-# description: Yoda condition
+# description: Yoda condition, it is
```
anderseknert, line 2: 😆

docs/rules/custom/one-liner-rule.md
```diff
@@ -47,6 +47,9 @@ rules:
       #
       # one of "error", "warning", "ignore"
       level: error
+      # maximum line length for a rule to be suggested as a one-liner
+      # default: 120
+      max-line-length: 120
```
anderseknert, line 52: Excellent! How did you notice?

anderseknert (approved):
Very good! IIRC, descriptions were intentionally kept short at first so that the table listing the rule wouldn't look so bad 😄 Good that we've moved past that!

A couple of questions, but mostly things I'd like to learn about.

bundle/regal/rules/performance/with-outside-test-context/with_outside_test_context.rego
charlieegan3: Interesting! I just thought this one was clearer so went with that.

bundle/regal/rules/style/yoda-condition/yoda_condition.rego
charlieegan3: Consistency between docs and code, it is.

docs/rules/custom/one-liner-rule.md
charlieegan3: I asked AI to find all the cases where Rule docs didn't match the implementation. This is the only meaningful one it found.

bundle/regal/rules/performance/non-loop-expression/non_loop_expression.rego
charlieegan3: Oh no particular reason other than we used the 'Non loop expression in loop' form elsewhere and I preferred the in loop bit at the end. So this is just find and replace... I can add the hyphen back.

All these edits are just copying from the rule to docs or vice versa.

Fixed in d8e9d2427ad186c9e4d01319cb1e4aaff1170175

PR #2031, opened 2026-06-16
[lsp] Update find references and rename request to support "some" variables

SeanLedford:
This PR makes some updates to the find references request implemented in #2029:

1. Move cross-module logic for finding references from rename.rego to references.rego, and used the result there to power the rename request
2. Add `some` vars as an additional result set returned in references.rego
3. Add tests for the new find references rules

More variable types are planned to be supported here shortly, as well as sharing this logic with our semantic token rules!

image

### DIFF
--- a/bundle/regal/lsp/references/references.rego
+++ b/bundle/regal/lsp/references/references.rego
@@ -42,7 +42,8 @@ _arg_refs contains ref if {
 _some_var_refs contains ref if {
 	[var, rule_index] := find.some_var_at_position
 
-	walk(_module.rules[rule_index].body, [_, value])
+	some node in ["head", "body"]
+	walk(_module.rules[rule_index][node], [_, value])
 
 	value.type == "var"
 	value.value == var.value
--- a/bundle/regal/lsp/references/references.rego
+++ b/bundle/regal/lsp/references/references.rego
@@ -12,18 +12,50 @@ import data.regal.lsp.util.range
 
 # METADATA
 # entrypoint: true
-result.response contains reference if {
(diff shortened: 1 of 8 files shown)


bundle/regal/lsp/references/references.rego
```diff
@@ -12,18 +12,50 @@ import data.regal.lsp.util.range
...
+	ref := {
+		"uri": url,
+		"range": range.parse(value.location),
+	}
+}
+
+_some_var_refs contains ref if {
+	[var, rule_index] := find.some_var_at_position
+
+	walk(_module.rules[rule_index].body, [_, value])
```
anderseknert: I think you'll want to walk the rule head here too as vars declared in `some` in the body can be referenced both in the rule's ref, key and value.. like:

```rego
rule[x].y := x if some x in input.arr
```

bundle/regal/lsp/references/references_test.rego
```diff
@@ -0,0 +1,72 @@
+package regal.lsp.references_test
```
anderseknert, line 1: **Very** good! These are tests I should have written, but I mostly wanted to see what could be done quickly. Thanks for picking that up 👍

anderseknert (approved):
This is some excellent work, Sean! Just one thing to consider. Can be done later if you want, but should be fairly easy to do I think.

bundle/regal/lsp/references/references.rego
SeanLedford: Ah good catch, added!

PR #2035, opened 2026-06-23
[feature] Extending find references variable support

SeanLedford:
A follow-up PR to #2029 and #2031 to provide find references support for `every` expressions and comprehensions!

### DIFF
--- a/bundle/regal/lsp/references/references.rego
+++ b/bundle/regal/lsp/references/references.rego
@@ -16,6 +16,10 @@ result.response contains ref if some ref in _arg_refs
 
 result.response contains ref if some ref in _some_var_refs
 
+result.response contains ref if some ref in _every_var_refs
+
+result.response contains ref if some ref in _comprehension_var_refs
+
 _arg_refs contains ref if {
 	[arg, rule_index] := find.arg_at_position
 	sref := ast.static_prefix(_module.rules[rule_index].head.ref)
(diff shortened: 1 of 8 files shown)


bundle/regal/lsp/util/find/find.rego
```diff
@@ -73,3 +73,60 @@ some_var_at_position := [var, rule_index] if {
...
+	var_pos := range.parse(var.location)
+	input.params.position.line == var_pos.start.line
+	input.params.position.character >= var_pos.start.character
+	input.params.position.character <= var_pos.end.character
+}
+
+_comp_declared_vars(body) := [v |
+	some expr in body
+	some symbol in expr.terms.symbols
+	some v in array.slice(symbol.value, 1, 100)
```
charlieegan3, line 129: Should this be count(symbol.value) rather than 100?

bundle/regal/lsp/references/references.rego
```diff
@@ -54,6 +58,45 @@ _some_var_refs contains ref if {
...
+
+# METADATA
+# description: |
+#   References to the variable declared via `some x in y` inside a
+#   comprehension. Walking the entire comprehension covers both the
+#   declaration and every reference, scoped to that comprehension only.
+_comprehension_var_refs contains ref if {
+	[var, comp] := find.comprehension_var_at_position
+
+	walk(comp, [_, value])
```
charlieegan3, line 88: How does this handle `[x | some x in a; [x | some x in b]]` cases where a nested comp shadows a var from a parent?

bundle/regal/lsp/references/references.rego
```diff
@@ -54,6 +58,45 @@ _some_var_refs contains ref if {
...
+
+	value.type == "var"
+	value.value == var.value
+	not startswith(value.value, "$")
```
charlieegan3, line 92: is this needed for every as well as users can't use that for gtd/rename etc?

charlieegan3 (approved):

bundle/regal/lsp/util/find/find.rego
anderseknert: This is a common perf hack we do to mean “until the end”, saving us an extra builtin function call where we can assume there won’t be more than a 100 parts… like in refs. Regal can’t save the souls of those who do that!

charlieegan3: Ahh I see. I wondered if it might have been copied from somewhere.

I guess we could have a helper if it's popping up in lots of places.

anderseknert: Yes, had it not been for the fact that a helper (as in custom function) is even more expensive to call :lolsob:

I guess it could be inlined by something like partial eval / compiler optimizations, but that’s currently a very mixed and big bag of things  we don’t necessarily want all of. My personal preference would be having OPA do constant folding, so we could get rid of magic numbers in the source while compiling to what we got here. I added some code to identify “constants” in OPAs compiler some time back, so it’s doable… but we’d want to have a way to apply optimizations selectively, or so I reckon.

anderseknert (approved):
Great work, Sean! <3

PR #2051, opened 2026-07-14
dap: Implement Evaluate feature

### LINKED ISSUE #2047: dap: Debug adapter LSP integration
The Regal/OPA debugger is incredibly useful already, and what's missing now is mostly nice-to-have features. But we obviously should aim to have nice things! Skimming the DAP spec and there a few missing features that we've already solved, and solved well, in the language server. For example the REPL-like [Evaluate](https://microsoft.github.io/debug-adapter-protocol/specification.html#Requests_Evaluate) feature, and of course [Completions](https://microsoft.github.io/debug-adapter-protocol/specification.html#Requests_Completions), which we could use more or less as is, just with some added awareness for the scope of the current frame.

I don't know yet what the solution looks like exactly, but there seems to be some room in the spec to connect to a running server rather than forking a new process, so perhaps that's the way to go? But it could perhaps also be the case that the debug process runs separately as it does now, and we have that connect to the server.

anderseknert:
- Move most DAP logic from cmd to internal/dap for reuse
- Allow DAP to be launched inside of the language server as opposed to a separate process
- Implement the Evaluate feature to allow evaluating state in an editor's debug REPL, watchers, and more.

Resolves #2047

Screenshot 2026-07-13 at 10 02 22


anderseknert:
Some follow-up issues created in OPA from this:
- https://github.com/open-policy-agent/opa/issues/8892
- https://github.com/open-policy-agent/opa/issues/8893

internal/dap/server.go
```diff
@@ -0,0 +1,104 @@
...
+				protoManager,
+				debug.NewDebugger(debug.SetEventHandler(NewEventHandler(protoManager)), debug.SetLogger(s.log)),
+				s.log,
+			).WithEvaluateHandler(s.evaluateHandler)
+
+			if err := protoManager.Start(ctx, conn, state.HandleMessage); err != nil {
+				s.log.Local.Error("failed to handle connection: %v", err)
+			}
+		} else {
+			s.log.Local.Error("failed to accept connection: %v", err)
```
charlieegan3, line 66: I presume that this will loop max once per connection so not going to be too verbose even if the failure is permanent?

internal/funsafe/funsafe.go
```diff
@@ -29,8 +40,45 @@ type (
...
 	objectElemSlice []*objectElem
+	namedVar        struct {
+		name  string
+		value ast.Value
+	}
 )
 
+// ToDebugVar converts a [debug.Variable] to a [DebugVar]. See [DebugVar] for details on why this is needed.
+func ToDebugVar(v debug.Variable) DebugVar {
+	ef := (*eface)(unsafe.Pointer(&v))
```
charlieegan3, line 51: I presume you have plans to not need to do this in future? I forget the context about why we can't expose what's needed upstream.

internal/dap/evaluate/evaluator.go
```diff
@@ -0,0 +1,114 @@
...
+type debugEvaluator struct {
+	argsAssembler func(ast.Body) []func(*rego.Rego)
+}
+
+// NewHandler returns a new Handler that evaluates expressions in what appears to be the context of a debug session.
+// See for example the Delve implementation of this feature to understand why "appears to be" is key here:
+//
+// https://github.com/go-delve/delve/blob/93f14c7401724729e2127c9aab5b75c5d7dcc152/service/dap/server.go#L3359
+//
+// Long story short, we can't actually run the eval in a stopped state, since it is... stopped. What we can do is to
```
charlieegan3, line 27: Could we also add something to the OPA debugger that would make this possible without the reconstruction? (in future, not this PR)

internal/lsp/server.go
```diff
@@ -1258,6 +1265,21 @@ func (l *LanguageServer) initializeResultHandler(ctx context.Context, result any
...
+	} else if l.featureFlags.DebugProvider {
+		// TODO: make interface binding configurable, unify logging between LSP and DAP
+		server := dap.NewServer("127.0.0.1:0", dap.NoOpLogger())
+		evaler := evaluate.NewHandler(l.debugArgsAssembler)
+
+		l.workspace = l.Workspace().WithDAPServer(server.WithEvaluateHandler(evaler))
+
+		go func() {
+			ctx, cancel := context.WithCancel(ctx)
+			defer cancel()
```
charlieegan3: What's cancel used for here?

charlieegan3 (approved):

internal/dap/server.go
anderseknert: I certainly hope so if Accept fails 😅 If not, I’m sure we’ll hear about it, lol. I guess I could add something to make sure it’s only logged once but OTOH that would make it more difficult to troubleshoot

internal/funsafe/funsafe.go
anderseknert: The immediate need is to work around this https://github.com/open-policy-agent/opa/issues/8890

But if we can have the debugger do the eval, we’d need little of this https://github.com/open-policy-agent/opa/issues/8893

Implementing it here was pretty much a way to learn these things 🙂

internal/dap/evaluate/evaluator.go
anderseknert: I hope so! Check the link to Delve in that doc comment though. It seems like no matter where we put it, it’s not exactly trivial.

internal/lsp/server.go
anderseknert: Good question. I think this part originated from other code, and AFAICS it doesn’t add anything of value here. Will remove.

PR #2057, opened 2026-07-14
Various completion provider improvements

anderseknert:
Fixing some things I've noticed while working on policies recently.

- Have common rule name provider provide suggestions for default rules
- Have input.json completion provider provide suggestion for imports
- Have boolean value completion provider provide suggestions in many more locations
- Add inline docs to import provider
- Add label details to built-in function completion provider
- Move cheap check above expensive in package completion provider
- Use `_rules` in ast.rego to allow LSP callers with different input schema
- Some small doc fixes

### DIFF
--- a/bundle/regal/lsp/completion/providers/builtins/builtins.rego
+++ b/bundle/regal/lsp/completion/providers/builtins/builtins.rego
@@ -13,7 +13,7 @@ import data.regal.lsp.location
 # METADATA
 # description: without editRange item defaults (large amount of duplicated data)
 items contains item if {
-	not _default_edit_range_supported
+	not client.supports.edit_range_defaults
 
 	line := input.regal.file.lines[input.params.position.line]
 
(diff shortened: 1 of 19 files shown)


bundle/regal/lsp/completion/providers/builtins/builtins.rego
```diff
@@ -35,20 +35,25 @@ items contains item if {
...
 items contains item if {
-	_default_edit_range_supported
+	client.supports.edit_range_defaults
 
 	line := input.regal.file.lines[input.params.position.line]
-
+	location.in_rule_body(line)
 	not startswith(line, "default ")
 
+	location.in_rule_body(line)
```
charlieegan3, line 54: Why is this needed twice? (52, 55)

charlieegan3 (approved):
This looks ok to me, but it's hard to break apart when ~8 changes are all in one commit. Mostly seems to be refactoring. Main new thing seems to be the boolean logic change and tests, but that looks good.

bundle/regal/lsp/completion/providers/builtins/builtins.rego
anderseknert: Oops, fixed!

PR #2070, opened 2026-08-04
OPA v1.19.0

anderseknert:
- Replace deprecated `test.WithTempFS` with `test.TempDir`. And adapt to the fact this doesn't write files with no content. Perhaps it should... 🤔
- Adapt tests to change in OPA which has location stripped from schema refs found in annotations
- Try using `strings.split_n` in many places, but this often results in cache misses compared to `split` as we use it extensively
- Use `util.Atoi` now provided by OPA

Thanks @charlieegan3 for the pointer / work on Go 1.27 support!


charlieegan3:
Replaces https://github.com/open-policy-agent/regal/pull/2066

charlieegan3 (approved):

charlieegan3:
> Perhaps it should... 🤔

Yeah, I feel like there are cases, e.g. an empty regal config file or input.json that might be nice to have tests for.

PR #2075, opened 2026-08-18
Support the experimental `and`/`or` keywords

sspaink:
Enable the experimental keywords in the capabilities used for parsing, encode `ast.LogicalAnd`/`ast.LogicalOr` in both RoAST encoders, and stop `constant-condition` from reporting an expression that makes up an entire `and`/`or` operand, as its fix would leave the enclosing expression without one. 

Pins OPA to 437786c8a6 on main that includes the new keywords, not sure if that can be merged as is or preferred to wait for the release next week. Either way works for me, just want to make sure this was ready when needed. The OPA playground does need this Regal update before we can support the keywords there. So this is part of: https://github.com/open-policy-agent/opa/issues/8999


bundle/regal/ast/search.rego
```diff
@@ -239,6 +239,27 @@ found.expressions[rule_index] contains value if {
...
 
+# METADATA
+# description: |
+#   set containing the locations of all expressions that make up the entirety of an
+#   `and` or `or` operand in the rule at rule_index. These are found by
+#   found.expressions as well, but unlike the other expressions found there, they
+#   can't be removed without leaving the enclosing expression without an operand.
+#   Note that expressions in a brace enclosed operand body holding more than one
+#   expression — like `{ x := 1; x == input.x } and input.y` — are not included
+logical_operand_locations(rule_index) := {expr.location |
```
anderseknert: Looks good to me, but the way this is queried for pretty much every rule index makes me wonder if it wouldn't do better as an object generated with the rule index as key, and building a set for each?

```rego
logical_operand_locations[rule_index] contains exp.location if ...
```

Tends to make it easier to debug too as you can easily inspect what it contains. What do you think?

bundle/regal/rules/bugs/constant-condition/constant_condition_test.rego
```diff
@@ -183,3 +183,52 @@ test_success_adding_constant_to_set if {
 	r := rule.report with input as ast.policy(`rule contains "message"`)
 	r == set()
 }
+
+test_success_constant_and_or_operand if {
+	# operands of `and`/`or` are excluded, as the fix for this rule — removing the
+	# expression — would change the meaning of the enclosing expression
```
anderseknert: Would be good to address this later, so please open an issue for that if not fixed now 🙂

Besides that, do we detect when the whole expression is constant now already? That _can_ be removed. e.g. `1 or 2`. It's silly, obviously, but people do silly things when hacking and then forget about it, which is really what this rule tries to help with.

anderseknert (approved):
Code looks clean to me, and I'm looking forward to play with the new keywords soon. Great work!

Left a comment or two but nothing major. My main request would be this: when we introduce new linter rules, or support for new features, we try to always use those ourselves as part of that. That we've been dog-fooding Rego since day one has likely been our greatest source for finding issues in OPA as well as for new ideas about what to improve in both projects. Do you think you could find at least a handful of places in Regal's policies where you can rewrite existing code to use the `and`/`or` keywords? 🙂 That way we'll be invested in this from the start. I'm also keen to see how these keywords perform compared to the alternatives!  We'll most certainly add many more in due course, but it'd be good to have this in from day one.

OK to do in a follow-up PR if you prefer to get this merged first.

bundle/regal/rules/bugs/constant-condition/constant_condition_test.rego
sspaink: issue created: https://github.com/open-policy-agent/regal/issues/2076

No silliness allowed in Rego! It wasn't before but that is a good idea, latest commit now removes `1 or 2`

bundle/regal/ast/search.rego
sspaink: Sounds like a good idea to me! updated `logical_operand_locations` to be an object

sspaink:
Dog fooding the keywords is a good idea, I updated a few files to use them: https://github.com/open-policy-agent/regal/pull/2075/commits/ef1c8b28891f1c6306bce3857ade3371c24a4f0e

What do you think?

bundle/regal/rules/bugs/constant-condition/constant_condition_test.rego
anderseknert: Very nice! No we certainly don’t want to allow silly Rego 🤪✋👮🏻‍♂️ Never wrote any myself, but I’ve heard it exists.

This change is shaping up great! I checked out your branch last night to try it out and it worked well! Found a bug or two to report later, and I think we aren’t properly passing the new capabilities to the Evaluate / Debug features. Did you get those to work? Besides that 👏

anderseknert:
Love the dogfooding! Some more to do here, but merging this now so that we can follow up with more specific PRs targeting remaining issues.

PR #2078, opened 2026-08-24
internal/io: drop stale experimental-keywords capability assertion

sspaink:
OPA enabled `and`/`or` in default capabilities (open-policy-agent/opa#9054), so asserting they're absent now fails against OPA main and breaks OPA's nightly Regal compatibility job.

nightly job: https://github.com/open-policy-agent/opa/actions/runs/32705813539/job/97366570649


srenatus (approved):
Thank you!

PR #2079, opened 2026-08-25
Report constant `and`/`or` operands

### LINKED ISSUE #2076: Report constant `and`/`or` operands
`constant-condition` skips any expression that makes up the entirety of an `and`/`or` operand, so none of these are reported:

```rego
allow if {
    input.a and 1
    input.b or "str"
    input.c and 1 == 1
}
```

They're excluded because the fixer removes the offending expression, which would leave the enclosing expression without an operand. Fixing them means collapsing the whole expression down to its non-constant operand instead — or reporting them without offering a fix at all.

Fully-constant logical expressions like `1 or 2` *are* reported as of #2075, since those can be removed outright.

Ref: https://github.com/open-policy-agent/regal/pull/2075#discussion_r3810661900

sspaink:
fix: #2076

Report the constant operand of an `and`/`or` expression with a range covering the operator too, so that removing it collapses the expression to the operand kept, unless that operand can't stand on its own.

### DIFF
--- a/bundle/regal/rules/bugs/constant-condition/constant_condition.rego
+++ b/bundle/regal/rules/bugs/constant-condition/constant_condition.rego
@@ -7,6 +7,7 @@ package regal.rules.bugs["constant-condition"]
 
 import data.regal.ast
 import data.regal.result
+import data.regal.util
 
 # METADATA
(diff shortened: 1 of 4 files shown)


anderseknert (approved):
Thank you, and sorry for the delay. It's good to be back home at last ☺️

PR #2081, opened 2026-08-27
feat(rules): add bugs/unconditional-with-conditions

### LINKED ISSUE #2080: Rule proposal: catch-all definition without `default` conflicts at runtime
### The pattern

A catch-all definition intended as a fallback, written without `default`:

```rego
package policy
import rego.v1

gender_bias_detected(metrics) if metrics.fairness.details.gender_bias_detected == true
gender_bias_detected(metrics) if metrics.summary.stereotype_values.gender_bias_detected == true

gender_bias_detected(_) := false   # meant as a fallback; is a third definition
```

The author's intent is "false unless one of the above matches". What Rego does is define the function three times. Any input satisfying one of the first two yields `true` while the third yields `false`, so:

```
eval_conflict_error: functions must not produce multiple outputs for same inputs
```

The correct form is `default gender_bias_detected(_) := false`.

### Why it is worth linting

The failure is **invisible until the interesting input arrives**. For every input where the conditional bodies fail, the catch-all supplies `false` and everything looks fine. The conflict fires only when a body succeeds — which, for a detector, is precisely the case it exists to catch.

Concretely: this was live in [GOPAL](https://github.com/Principled-Evolution/gopal), a Rego policy library for AI compliance. The affected helper is imported by the EU AI Act fairness policy. For any AI system where bias *was* detected, the policy did not return a denial — it returned nothing at all, because evaluation aborted. Every test passed, because the tests mocked that helper and never exercised the real function.

Nothing in the toolchain flags it (Regal 0.42.0, which embeds OPA 1.18.2; `opa` CLI 1.19.1):

```
opa check           passes
opa check --strict  passes
regal lint          passes
regal lint --enable-all  passes
```

### It affects plain rules too

```rego
p if input.x
p := false          # same trap
```

`data.t.p` with `{"x": true}` → `eval_conflict_error`.

And the catch-all does not need a wildcard argument. A named-but-unused parameter behaves identically:

```rego
f(x) if x == 1
f(y) := false       # still a third definition, still conflicts
```

### The case a naive rule would get wrong

If the catch-all's value equals what the other definitions produce, there is **no** conflict and nothing to report:

```rego
f(x) := false if x == 1
f(_) := false       # both yield false — legal, evaluates to false
```

So the check cannot simply be "catch-all definition without `default`". It has to compare values. The tractable form:

> A definition of `f` whose arguments are all unconstrained and whose value is a constant `V`, coexisting with another definition of `f` that can produce a constant other than `V`.

The common case is fully static: a body-only definition (`f(x) if …`) has an implicit value of `true`, so a catch-all assigning any constant other than `true` is a guaranteed conflict whenever that body holds. A conservative rule covering only that case would have caught the GOPAL bug and would not fire on the example above.

### Relationship to #42

#42 is the same family and makes the same argument better than I can:

> The big problem with this is that you can test, publish and run a policy for days or longer, before the problem manifests at *runtime*. The compiler accepts this, but it would be extremely helpful if Regal didn't.

Exactly so. The mechanism differs, though. #42 is one definition whose body iterates and so yields multiple outputs for one input. This is several definitions where a catch-all disagrees with a conditional one. Same error code, same "clean until it matters" failure, different thing to detect — so this may be a sibling rule rather than part of that one. Happy for it to be folded in if you see them as one rule.

### Other rules and issues I checked first

None cover it, though several are adjacent:

| Rule or issue | Why it does not apply |
| --- | --- |
| `bugs/rule-assigns-default` (#1018) | Flags a rule assigning the *same* value as its declared `default`. Here there is no `default` at all |
| `bugs/argument-always-wildcard` | Requires the argument to be unused across *all* definitions. `f(x) if x == 1` uses it |
| `bugs/duplicate-rule` | Identical bodies, not conflicting values |
| `bugs/inconsistent-args` | Argument naming and ordering |
| #152, boolean rule without a default | That one is explicitly "not wrong per se" and style-category, because an undefined boolean is legal. This is the opposite case: a catch-all definition *instead of* a default, which is a runtime error rather than a stylistic gap. Worth distinguishing, since the fix for both is the `default` keyword |
| #211, prefer default function to conditionless `else` | Concerns `else`, and is a preference rather than a conflict |

### Category

`bugs` seems right: the code is legal, compiles, lints clean, and is wrong at runtime for a subset of inputs.

### Offer

Happy to implement it and send a PR if you think it belongs. I have the reproductions above as test cases, including the negative one. If you would rather it lived as a custom rule in the consuming project, that is a fair answer too — but the failure mode seemed general enough to be worth raising, since the whole point is that it passes every gate a careful author would think to run.

kmadan:
Closes #2080.

Thanks for the invitation to send this, @anderseknert — and for the
simplification. You were right that no comparison of the values is needed:

> The presence of a rule definition that's effectively constant should always
> have Regal report conditional definitions of the same rule. Even when the
> conditions aren't conflicting they are still pointless.

That made the rule simpler than what I had sketched in the issue.

## What it reports

A rule with one definition written without conditions alongside other
definitions that carry them:

```rego
allow := true if input.user.is_admin

allow := false   # always defined, so the condition above cannot decide anything
```

Either the values differ and evaluation fails with `eval_conflict_error`, or
they agree and the conditional definitions are dead code. Almost always the
intent was a fallback, which belongs in a `default`.

## Nothing catches this today

`opa check --strict` passes on all three forms. With `--enable-all`, the only
Regal rule that fires is `rule-assigns-default`, and only on the narrowest case —
where a `default` exists *and* the values match:

```rego
allow := true if input.ok
allow := false                  # reported by nothing

deny := true if input.bad
deny := true                    # reported by nothing

default f(_) := false
f(x) if x > 10
f(_) := false                   # rule-assigns-default fires here
```

I left the overlap on the third case rather than coupling the two rules — the
diagnoses differ ("this equals the default" vs "this is unconditional") and both
are true. Happy to suppress one if you would rather not have two violations on
that line.

## Detection notes

Mostly about avoiding false positives:

- The unconditional definition must assign a **constant**. `allow :=
  input.override` can itself be undefined, and then a conditional definition
  genuinely is what answers.
- **Multi-value rules are excluded** — their definitions union rather than
  conflict. Worth noting that testing `head.key` alone is wrong: a ref-head rule
  like `config.mode := "loose"` also carries a key (the last part of its ref) and
  is single-value. Presence of `head.value` is the real distinction, and my first
  version got this wrong.
- **Function definitions only catch everything when every argument is a distinct
  variable.** `f("a") := 1` has no body but applies to one argument; `f(x, x)` is
  a constraint.
- Refs with variable parts (`p[x] := v`) and `else` chains are left alone.
- `ast.static_ref` takes a term rather than a bare ref, and a rule head's `ref`
  is the array itself, so that check is spelled out locally. Let me know if you
  would prefer a helper added to `ast` instead.
- `body` is **absent** rather than empty on a definition written without `if`, so
  `count(rule.body)` is undefined rather than `0`. Fittingly for this rule, the
  first draft therefore reported nothing at all.

Known limitations are documented on the rule page: definitions are only compared
**within one file**, so a rule set split across files in the same package is not
detected.

## Verification

- **1095/1095** Rego tests, all 39 Go packages, and the e2e suite pass
- `regal lint bundle/` — **no violations** (355 files)
- **Zero false positives across 551 real Rego files**: Regal's own bundle (355)
  and the [GOPAL](https://github.com/Principled-Evolution/gopal) policy library
  (196)
- Confirmed to catch the real defect that prompted the issue — a third complete
  definition of `gender_bias_detected(_) := false` sitting alongside a `default`
  and two conditional definitions
- 22 tests: 10 for cases that must be reported, 12 for cases that must not
- `markdownlint` clean; `dprint fmt` applied to `data.yaml`

I could not run `build/do.rq pr` as a whole (no `rq` locally), so I ran its steps
individually: `regal test bundle`, `go test ./...`, `go test -tags e2e ./e2e`,
`regal lint bundle/`, `build/update-readme.sh check`, `markdownlint`, `dprint`.

Those figures are from the revision that opened this PR. For the current one I
re-ran `opa check` over the bundle, `regal test bundle` and `regal lint bundle/`,
the latter two against a copy with `main`'s `and`/`or` imports desugared, since
no released Regal binary parses them. The Go and e2e suites are covered by CI
rather than locally on this revision.


bundle/regal/rules/bugs/unconditional-rule-definition/unconditional_rule_definition.rego
```diff
@@ -0,0 +1,92 @@
...
+	violation := result.fail(rego.metadata.chain(), result.location(unconditional))
+}
+
+# Definitions of single-value rules, grouped by the name they define.
+#
+# Multi-value rules (`deny contains msg if ...`) are excluded, as their
+# definitions union rather than conflict — an unconditional `deny contains "x"`
+# says nothing about whether the other definitions matter. Refs with variable
+# parts (`p[x] := v`) generate a collection for the same reason.
+_single_value_definitions[name] contains rule if {
```
anderseknert: I think this could be simplified / made more efficient if you use the number of conditions as another level of keys for the object built here already, avoiding the quadratic iteration in `report` above. e.g.

```rego
_single_value_definitions[name][num_conditions] contains rule if {
```

Thinking some more, do we even need the number of conditions? Perhaps use just a boolean key for unconditional vs not would be enough:

```rego
_single_value_definitions[name][unconditional] contains rule if {
    # some rule ...
    name := ast.ref_to_string(rule.head.ref)
    unconditional := is_unconditional(rule) # must return true or false, not undefined
```
 A violation would then simply look like this:
```rego
# report contains ...
some name
_single_value_definitions[name][true] # has unconditional definition
_single_value_definitions[name][false] # also has conditional definition

# report first rule, any or what not
```

It's late here though, so don't take my word for it :) Worth playing some more with this though I think.

bundle/regal/rules/bugs/unconditional-rule-definition/unconditional_rule_definition.rego
```diff
@@ -0,0 +1,92 @@
...
+#
+# `body` is absent altogether on a definition written without `if`, rather than
+# present and empty, and an absent key makes `count` undefined rather than zero.
+# The default here is what keeps that from quietly failing every comparison —
+# the same "undefined is not false" trap this rule exists to catch.
+_condition_count(rule) := count(object.get(rule, "body", []))
+
+# `ast.static_ref` takes a term rather than a bare ref, and a rule head's `ref`
+# is the array itself, so the check is spelled out here.
+_static_ref(ref) if every part in array.slice(ref, 1, count(ref)) {
```
anderseknert: `ast.ref_static_to_string` is likely what you're looking for

bundle/regal/rules/bugs/unconditional-rule-definition/unconditional_rule_definition.rego
```diff
@@ -0,0 +1,92 @@
...
+	part.type == "string"
+}
+
+_always_defined(rule) if {
+	_condition_count(rule) == 0
+	not rule["else"]
+
+	# A non-constant value can itself be undefined, and then the conditional
+	# definitions are not pointless at all: in
```
anderseknert: Indeed. I'd try to bake this into the function that counts conditions, or `is_conditional` as mentioned above, if we go that way

bundle/regal/rules/bugs/unconditional-rule-definition/unconditional_rule_definition.rego
```diff
@@ -0,0 +1,92 @@
...
+
+	_matches_any_arguments(rule)
+}
+
+# Rules that are not functions take no arguments, so nothing further to check.
+_matches_any_arguments(rule) if not rule.head.args
+
+# A function definition only catches everything when every argument is a
+# distinct variable. `f("a") := 1` has no body, but applies only to the argument
+# "a", so it leaves the other definitions in the set doing real work. Repeating
```
anderseknert: Unlikely encounter, but valid (if non-sensical) Rego, and would be a violation against this rule:
```rego
f("a") := 1
f("a") := 2 if input.condition

# or

f("a") := 1
f("a") := input.condition
```
It's OK for a first version of a rule to not cover every possible case though.

bundle/regal/rules/bugs/unconditional-rule-definition/unconditional_rule_definition_test.rego
```diff
@@ -0,0 +1,228 @@
...
+	r := rule.report with input as ast.with_rego_v1(`
+	allow := true if input.ok
+
+	allow := false if input.blocked
+	`)
+
+	r == set()
+}
+
+# `else` is the supported way to write a fallback. It is one rule with an
```
anderseknert: Certainly supported, but it's one way of many. Like `default` as you previously demonstrated.

bundle/regal/rules/bugs/unconditional-rule-definition/unconditional_rule_definition_test.rego
```diff
@@ -0,0 +1,228 @@
...
+		x > 10
+	} else := 2
+	`)
+
+	r == set()
+}
+
+test_success_ref_with_variable_part if {
+	r := rule.report with input as ast.with_rego_v1(`
+	config[k] := "loose"
```
anderseknert: This is invalid, as `k` is unsafe in this definition. It still _parses_, which is why this test passes. Probably best to leave variable ref head rules out of this for now as that'll require much more complex heuristics, and I don't think it'll be worth it at this point.

bundle/regal/rules/bugs/unconditional-rule-definition/unconditional_rule_definition_test.rego
```diff
@@ -0,0 +1,228 @@
+package regal.rules.bugs["unconditional-rule-definition_test"]
+
+import data.regal.ast
+
+import data.regal.rules.bugs["unconditional-rule-definition"] as rule
+
+# The shape of the violation, asserted once in full. The remaining tests assert
+# which definition was flagged, which is the part that can actually regress.
+test_fail_catch_all_with_conflicting_value if {
+	r := rule.report with input as ast.with_rego_v1(`
```
anderseknert: As the docs for `ast.with_rego_v1` state: 

> deprecated: use ast.policy instead

You'll need to update the location below though as the import pushes other things down a couple of lines.

bundle/regal/rules/bugs/unconditional-with-conditions/unconditional_with_conditions_test.rego
```diff
@@ -0,0 +1,228 @@
...
+	`)
+
+	r == set()
+}
+
+test_success_different_rules_do_not_pair_up if {
+	r := rule.report with input as ast.with_rego_v1(`
+	allow := true if input.ok
+
+	deny := false
```
anderseknert, line 275: Perhaps for another rule, but we could also detect conflicting unconditional rules I suppose?
```rego
x := 1
x := 2
```
Obviously less likely to happen though

anderseknert (commented):
Thanks! Haven't reviewed the docs yet, but I figured I'd submit what I had in case you planned for some weekend hacking. Looks pretty good to me, but still some work to do. Will come back to the docs and any updates early next week.

kmadan:
Thanks for taking the time on this, especially heading into the weekend. All pushed and CI is green.

Done: switched to `ast.policy` (rows shift by two, as you called), and the pairing now goes through a set of names that carry a conditional definition, so no more walking both sides. Also wrote the `f("a")` gap and the variable-ref exclusion into the docs rather than leaving them implied.

Good catch on `config[k] := "loose"`. It turns out to be worse than an unlikely case: it can't compile at all, `opa check` gives `var k is unsafe`, so the test was asserting on a policy nobody can write. Removed.

On `ast.ref_static_to_string`, I may well be misreading it, but it looks like it truncates rather than tests, so `config[k]` would come back as `"config"` instead of being excluded. `ast.static_ref` seems to be the predicate I want, though it reads `ref.value` and a rule head's `ref` is the bare array. I've gone with `ref_static_to_string(ref) == ref_to_string(ref)` for now, since truncation is a no-op exactly when the ref is static. Very happy to wrap the term instead if that reads better to you.

I left the boolean key alone, but I'd value your call on it. My two predicates aren't quite complements: one is "has a body", the other is "no body, no `else`, constant value, catch-all args". So a `false` bucket would also pair `allow := input.override` with `allow := true`, which is a real bug but closer to your `x := 1 / x := 2` case than to this rule. Didn't want to widen the scope by accident, but if you'd rather have the simpler shape and the wider net, I'll happily do that.

Agreed on `x := 1 / x := 2` being worth catching separately.

anderseknert:
@kmadan reviews are commonly split up in multiple comments so that each of them can be evaluated, discussed and eventually marked as resolved individually. For next time, let's try to keep the discussion threaded by comment rather than trying to address everything in a single response.

I pulled down your branch to test it, and functionally I believe this rule does what it should now 👍 
I still think this rule could be simplified in the direction previously suggested, and more so now that I've had the time to see that it actually works :) The brand new `or` keyword did a lot for expressiveness here. Let me know what you think: https://gist.github.com/anderseknert/6eb921c9d3c8285b1ec4493c630a212a (and of course, feel free to use)

Before I go on to review the rule docs, I'd appreciate a few fixes around naming. The name `unconditional-rule-definition` doesn't really highlight the issue I think, as there is nothing wrong with having unconditional rule definitions, just mixing them with conditional definitions. Perhaps `unconditional-with-conditions` would better describe that?  Or even just `unconditional-conditions`. Let me know if you agree or not.

The tests look good and covers the scenarios I can think of 👍 Could we please just stick to "conditional" and "unconditional" in our naming (and comments) in these tests? While I understand what you mean with "catch all" I think that's a description of your original intent more than it describes how those rules behave, and it'd be good if we stick to the same language everywhere.

bundle/regal/rules/bugs/unconditional-rule-definition/unconditional_rule_definition.rego
kmadan: Adopted, with one change to the key. The group now records whether each definition carries conditions, rather than whether it is unconditional, so `report` is a lookup of `definitions[true]` followed by iteration over `definitions[false]`. Both the condition count and the separate set of names carrying a conditional definition are gone.

Keying on `is_unconditional` instead, as the gist does, places a definition that can itself be undefined in the `false` bucket, so `allow := true` beside `allow := input.override` is reported. That is a real conflict, but it extends the rule into unconditional-versus-unconditional territory, where `allow := true` beside `allow := false` is the plainest case and this structure cannot report it at all, since nothing lands in the other bucket. Keying on conditions keeps the scope that the name and the docs describe. I am happy to widen it deliberately if you would rather, in which case the summary line should stop mentioning conditions.

One detail worth keeping: the key function needs `default _has_conditions(_) := false`. Without it the function is undefined for a definition written without `if`, the enclosing rule body fails, and the definition drops out of the group and takes its bucket with it.

bundle/regal/rules/bugs/unconditional-rule-definition/unconditional_rule_definition.rego
kmadan: I used it in the previous round, and it turns out not to test staticness. `_format_term` has no format for a `ref`, `call` or `templatestring` part, so both `ref_static_to_string` and `ref_to_string` drop such a part, and both return `config` for `config[input.key]`. The comparison then holds, the head counts as static, and

```rego
config[input.key] := "loose"

config[input.key] := "strict" if input.strict
```

was reported. It does work for a `var` part, which is why the earlier fixture passed.

The rule now inlines the type test from `ast.static_ref`, as your gist does. Whether `ref_to_string` should drop those parts silently is a separate question, since other rules use it for identity. Say the word and I will open an issue.

bundle/regal/rules/bugs/unconditional-rule-definition/unconditional_rule_definition.rego
kmadan: Done. There is no condition count any more. `_has_conditions(rule) if rule.body`, with `default _has_conditions(_) := false`, is the key of the group, and the `object.get(rule, "body", [])` shim is gone with it. `body` is absent rather than empty on a definition written without `if`, and the default now covers that instead of the shim.

bundle/regal/rules/bugs/unconditional-rule-definition/unconditional_rule_definition.rego
kmadan: Still uncovered, and now recorded in the docs as a limitation. Reporting it needs the arguments of two definitions compared against each other, rather than each definition checked on its own, which is more than the rule does today.

bundle/regal/rules/bugs/unconditional-rule-definition/unconditional_rule_definition_test.rego
kmadan: Reworded. The comment now states that `else` gives one rule with an ordered chain rather than a set of definitions, and notes that it is one of several ways to write a fallback, with `default` as another.

bundle/regal/rules/bugs/unconditional-rule-definition/unconditional_rule_definition_test.rego
kmadan: Correct, and `opa check` confirms it: `rego_unsafe_var_error: var k is unsafe`. The fixture only ever parsed, which is why the test passed.

A variable in the ref cannot appear in a definition without a body for that reason, so the exclusion has to be covered by a computed key instead. The test now uses `config[input.key] := "loose"`, which compiles. That fixture is also what exposed the `ref_static_to_string` problem in the thread above. Variable ref heads stay out of the rule, and the docs record it.

bundle/regal/rules/bugs/unconditional-rule-definition/unconditional_rule_definition_test.rego
kmadan: Done in the previous round. Every fixture uses `ast.policy`, and the rows in the one full-shape location assertion moved up by two.

bundle/regal/rules/bugs/unconditional-with-conditions/unconditional_with_conditions_test.rego
kmadan: Agreed that it belongs elsewhere. This structure cannot report it: both definitions land in the same bucket and the other one stays empty, so there is nothing to pair them with. `duplicate-rule` does not catch it either, as it compares rule text. There is a test and a docs entry pinning it as out of scope, so the boundary is at least explicit. I can open an issue for it, or fold it in here if you would rather one rule covered both.

kmadan:
Threading noted, one comment per point from here. The eight review comments are answered individually above.

On the name, I agree, and for the reason you give: an unconditional definition on its own is fine, so the name should describe the combination. It is now `unconditional-with-conditions`. I passed over `unconditional-conditions` because it reads as a contradiction next to `constant-condition`. The rename covers the directory, both files, the docs page and its configuration sample, `data.yaml` and the e2e fixture. Nothing is released under the old name, so no configuration alias is needed.

The tests use "conditional" and "unconditional" throughout now, in names and in comments.

Your gist is adopted. The `or` keyword does read well here. Two things in it are worth flagging beyond the keying discussion in the thread above:

- `not rule.body or ast.is_constant(rule.body.terms.value)` reduces to `not rule.body`. `body` is an array, so `body.terms` is undefined and the second operand never binds; `rule.body[0].terms` is what reaches the expression. I left it as `not rule.body`, since `constant-condition` already reports a lone `true` in a body and folding that in would report the same line twice.
- Restricting function arguments to wildcards loses `f(x) := false` beside `f(x) := true if x > 10`. An argument the body does not constrain applies to every call whether it is named or a wildcard, so the check requires distinct variables, which wildcards satisfy as they are named `$0`, `$1` and so on. `ast.is_wildcard` covers the narrower reading if that was the intent.

The branch was force-pushed a few minutes ago, so a pull will need `--force` if you have it checked out. `b3e5822` replaces `53dd743`, whose commit message described the `ref_static_to_string` check as intentional, before the problem in that thread turned up.

bundle/regal/rules/bugs/unconditional-with-conditions/unconditional_with_conditions_test.rego
anderseknert: I did the other day: https://github.com/open-policy-agent/regal/issues/2084
As well as this one: https://github.com/open-policy-agent/regal/issues/2085

anderseknert (approved):
I'll probably do some light post-processing here, but let's get this in. Thanks for contributing to Regal 👑

PR #2089, opened 2026-09-02
bump gobwas/glob to v1.0.0

sspaink:
bumped in OPA: https://github.com/open-policy-agent/opa/pull/9116 so nightly job is failing without this bump in Regal: https://github.com/open-policy-agent/opa/actions/runs/33608605254/job/100178234442

dependabot attempted this here: https://github.com/open-policy-agent/regal/pull/2088, there is this comment in the dependabot config "[# update OPA manually to bump version in README too](https://github.com/open-policy-agent/regal/blob/main/.github/dependabot.yml#L24)" that makes me think dependabot wasn't suppose to open that PR? I think removing the v1 in the exclusion config would fix it.


anderseknert (approved):
Thanks a lot!

PR #2092, opened 2026-09-22
Rule: prefer-string-interpolation

anderseknert:
Ideally this rule would have been added following the OPA release introducing interpolation, but better late than never. As https://github.com/open-policy-agent/regal/issues/1877 describes, this should not be considered done until we've nailed the user experience for this completely. That shouldn't stop us from shipping something though, and while the obvious next step is a fixer for rewriting `sprintf` calls -> interpolation, that's likely non-trivial and should not hold back having basic linting flag this as something to look into, if only manually for now. We need more users to discover string interpolation!

See the docs added with this PR for all the details.


bundle/regal/rules/idiomatic/prefer-string-interpolation/prefer_string_interpolation.rego
```diff
@@ -0,0 +1,56 @@
+# METADATA
+# description: Prefer string interpolation where possible
```
charlieegan3, line 2: I would rather not have "where possible" here. To me, this is Regal's job (to determine "where possible" for the user)

Other than [idiomatic/use-strings-count](https://www.openpolicyagent.org/projects/regal/rules/idiomatic/use-strings-count), generally our use- rules do not have this suffix in the description.

bundle/regal/rules/idiomatic/prefer-string-interpolation/prefer_string_interpolation.rego
```diff
@@ -0,0 +1,56 @@
...
+# custom:
+#   severity: none
+notices contains result.notice(rego.metadata.chain()) if not capabilities.has_sprintf
+
+report contains violation if {
+	some rule_index, fun
+	ast.function_calls[rule_index][fun].name == "sprintf"
+
+	fun.args[0].type == "string"
+	fun.args[1].type == "array"
```
charlieegan3, line 32: What are we guarding against here? is it just avoiding firing the rule when the user's rego is incorrect and needs fixing first or something more serious? if the latter, might be nice to have a comment explaining why since it appears as a bit of a tautology.

bundle/regal/rules/idiomatic/prefer-string-interpolation/prefer_string_interpolation.rego
```diff
@@ -0,0 +1,56 @@
...
+
+	str_no_escape := replace(fun.args[0].value, "%%", "")
+
+	# for now, only if all patterns match either %s, &q, %d and %v (optionally,
+	# with # in front of it) do we recommend using interpolation instead
+	regex.match(`%#?[sqdv]{1}`, str_no_escape)
```
charlieegan3, line 38: If the sprintf has no formatting directives, then we could recommend it too? is `regex.match(`%#?[sqdv]{1}`, str_no_escape)` needed in addition to `not regex.match(`%#?[^sqdv]`, str_no_escape)`?

bundle/regal/rules/idiomatic/prefer-string-interpolation/prefer_string_interpolation.rego
```diff
@@ -0,0 +1,56 @@
...
+	not regex.match(`%#?[^sqdv]`, str_no_escape)
+
+	# simple variable reference args (e.g. `[x]`) are guaranteed to be defined by compiler,
+	# but whether to also flag calls like `sprintf("%s", [input.x])` (argument contains ref)
+	# must be up to the user to opt in to, as replacing such calls with interpolation changes
+	# the semantics of policy evaluation for potentially undefined references:
+	# - `sprintf` = evaluation fails on undefined ref
+	# - interpolation = writes `<undefined>` in place of ref
+	{
```
charlieegan3, line 47: This is nice.

docs/rules/idiomatic/prefer-string-interpolation.md
```diff
@@ -0,0 +1,132 @@
+# prefer-string-interpolation
+
+**Summary**: Prefer string interpolation where possible
```
charlieegan3, line 3:
```suggestion
**Summary**: Prefer string interpolation
```

See comment above.

docs/rules/idiomatic/prefer-string-interpolation.md
```diff
@@ -0,0 +1,132 @@
...
+**Category**: Idiomatic
+
+**Avoid**
+
+```rego
+package policy
+
+request_formatted := sprintf("%s: %s", [method, path]) if {
+    # extract method and path from the request
+}
```
charlieegan3, line 14: This might be the minimal example, but I feel like is more common in policies I see in the wild. 

```
request_formatted := result if {
    # extract method and path from the request
    result := sprintf("%s: %s", [method, path]) 
}
```

I don't really mind, the original example is a clearer like for like replacement, this example, I think, does a better job of showing the value of the string interp alternative.

docs/rules/idiomatic/prefer-string-interpolation.md
```diff
@@ -0,0 +1,132 @@
...
+
+request_formatted := $"{method}: ${path}" if {
+    # extract method and path from the request
+}
+```
+
+## Rationale
+
+The introduction of [string interpolation](https://www.openpolicyagent.org/docs/policy-language#string-interpolation)
+in OPA ([v1.12.0](https://github.com/open-policy-agent/opa/releases/tag/v1.12.0)) is without a doubt one of the most
```
charlieegan3, line 30: suggestion only, but I have tried to keep the language very plain in our docs
 (e.g. https://github.com/open-policy-agent/opa/pull/8786)

```suggestion
in OPA ([v1.12.0](https://github.com/open-policy-agent/opa/releases/tag/v1.12.0)) is one of the most
```

docs/rules/idiomatic/prefer-string-interpolation.md
```diff
@@ -0,0 +1,132 @@
...
+significant improvements to the Rego language since OPA 1.0, and interpolation should now be preferred over alternative
+methods (like `sprintf`) whenever possible.
+
+String interpolation improves readability by placing variables, references and even whole expressions directly in the
+string to be rendered. Coupled with sensible variable names, the result is code that is easier to both read and
+write, while losing none of its expressiveness. String interpolation comes with many additional benefits, but also a few
+caveats to be aware of.
+
+### Handling of Undefined
```
charlieegan3, line 39: Might it be better to ensure the docs at https://www.openpolicyagent.org/docs/policy-language#string-interpolation are complete and link to that instead?

This section has generally good content about the language feature while not being very regal specific.

charlieegan3 (approved):
Nice, this looks good to me. Some comments to consider, most importantly the placement of some of the docs content.

But ultimately nothing blocking from me.

bundle/regal/rules/idiomatic/prefer-string-interpolation/prefer_string_interpolation.rego
anderseknert: Tautology? 🤔 It's just us backing out of something like this, as we have no way of resolving it:
```rego
sprintf(input.format, input.verbs)
```

bundle/regal/rules/idiomatic/prefer-string-interpolation/prefer_string_interpolation.rego
anderseknert: I suppose so. But would either be correct then? I'd be inclined to report that as a separate issue, just like sprintf (and other built-ins) provided an empty array, as has been discussed some time in the past.

I mean, we definitely could report on a call like:
```rego
sprintf("", [])
```
But to say "you can use interpolation instead" is not the right message here 😄

docs/rules/idiomatic/prefer-string-interpolation.md
anderseknert: > is more common in policies I see in the wild

Indeed. That's what we are here to fix 😊

docs/rules/idiomatic/prefer-string-interpolation.md
anderseknert: I think it may be right to do that in the OPA docs, but Regal's docs were never "reference docs" and they shouldn't be. We're opinionated pretty much by design, and I think our docs should state clearly what we think is good and bad. I don't mind removing "without a doubt" if you think it sounds better without it, but "very plain language" sounds like all work and no fun 🥲

bundle/regal/rules/idiomatic/prefer-string-interpolation/prefer_string_interpolation.rego
anderseknert: > To me, this is Regal's job

That's a valid point. I think the distinction that I want to convey is that vs rules like e.g. that one unconditionally flagging/replacing `=` with `:=` in rule heads or whatever, while some rules are more nuanced, and the nuance is _important_ when you consider for example the undefined behavior change swapping sprintf for interpolation. 

On the other hand, `Prefer string interpolation` is the correct form when printed on the location _reported_  as then Regal has already determined that to be the case.

anderseknert: Removing "where possible" from the title, as well as in the title of use-strings-count. Users will have to check the docs for the nuances.

docs/rules/idiomatic/prefer-string-interpolation.md
anderseknert: It's specific to the linter rule in that it covers replacing `sprintf` with string interpolation, and particularly when input/data references are passed as args. Wouldn't mind moving this to the OPA docs, but that seems like it would be a rather opinionated section in that context?

anderseknert:
Follow-up PR covering `concat` will incorporate the changes suggested here.

PR #2093, opened 2026-09-22
Include `concat` in `prefer-string-interpolation`

anderseknert:
Also addresses some comments brought up in review of the first iteration.


charlieegan3 (approved):
Thanks for making those edits too, LGTM

PR #2094, opened 2026-09-23
lsp: Add support for coverage in eval inline

charlieegan3:
If the client inits with support for display of inline coverage, then the server will add coverage report to eval response. Note that this will cause the rego to be evaluated additional times to support early_exit and index_excluded annotations on not covered ranges. However, this extra runs will use the same ndbuiltin cache.

Server side of https://github.com/open-policy-agent/vscode-opa/pull/498


internal/lsp/types/types.go
```diff
@@ -34,6 +34,10 @@ type (
 		// EvalCodelensDisplayInline, if set, will show evaluation results natively
 		// in the calling editor, rather than in an output file.
 		EvalCodelensDisplayInline bool `json:"evalCodelensDisplayInline,omitempty"`
+		// EnableEvalInlineCoverage, if set, will cause a regal.eval command to also
+		// evaluate coverage and include the results in the regal/showEvalResult
+		// notification's "coverage" field.
+		EnableEvalInlineCoverage bool `json:"enableEvalInlineCoverage,omitempty"`
```
anderseknert, line 40: It's annoying that it's "Enable" for some options not others like `EvalCodelensDisplayInline` above. Not sure which naming scheme we want but it'd be good to be consistent. Oh well that'd require changing other options too..

internal/lsp/eval.go
```diff
@@ -102,9 +104,33 @@ func (l *LanguageServer) handleEvalCommand(ctx context.Context, args types.Comma
...
+		l.featureFlags.InlineEvaluationProvider &&
+		l.Workspace().Client().InitOptions.EvalCodelensDisplayInline
+
+	var inputOpts []rego.EvalOption
+	if inputValue != nil {
+		inputOpts = append(inputOpts, rego.EvalParsedInput(inputValue))
+	}
+
+	if evalWithCoverage {
+		if result, report, err = l.EvalInWorkspace(ctx, args.Query, cover.New(), inputOpts...); err != nil {
```
anderseknert: Not sure I get this. If we want coverage we first do one evaluation for that, then continue to do another one further down? 🤔

internal/lsp/eval.go
```diff
@@ -183,47 +213,88 @@ func (l *LanguageServer) getRuleHeadLocations(
...
+		return emptyEvalResult, nil, fmt.Errorf("failed preparing query %s: %w", resultQuery, err)
+	}
+
+	var (
+		resultSet rego.ResultSet
+		ndCache   builtins.NDBCache
+	)
+
+	if cov != nil {
+		ndCache = builtins.NDBCache{}
```
anderseknert, line 239: What's this needed for?

internal/lsp/eval.go
```diff
@@ -183,47 +213,88 @@ func (l *LanguageServer) getRuleHeadLocations(
...
+	if _, err = pq.Eval(ctx, append(cover.NoIndexingEvalOptions(indexExcluded, ndCache), opts...)...); err != nil {
+		return result, nil, fmt.Errorf("failed evaluating index-excluded coverage query: %w", err)
 	}
 
-	res, ok := result[0].Bindings["result"]
-	if !ok {
-		return emptyEvalResult, errors.New("expected result in bindings, didn't get it")
+	earlyExit := cover.New()
+
+	if _, err = pq.Eval(ctx, append(cover.NoEarlyExitEvalOptions(earlyExit, ndCache), opts...)...); err != nil {
```
anderseknert: How many evals are we doing for coverage? 😅 There's no other way?

anderseknert (approved):
Some questions, and I'm a bit concerned about doing evaluation repeatedly for a single request. But at least it's somewhere to start from. Looking forward to trying it out.

internal/lsp/eval.go
charlieegan3: This is needed so that any non deterministic builtins are not called again and use the same values between the multiple coverage runs.

internal/lsp/eval.go
charlieegan3: This is how it works in OPA's test runner, the repeated runs are just not wrapped and must be done manually by callers (which is something we might want to address, but needs a change in OPA to expose this).

charlieegan3: https://github.com/open-policy-agent/opa/pull/8937 originally implemented this differently, but in the end we opted to do it with separate runs to make the implementation more reliable and simple to understand.

internal/lsp/eval.go
charlieegan3: Yeah that was just meant to be in case the coverage run failed for coverage related reasons, but I can likely make that better.

charlieegan3: Have made a minor edit in bdc6364c68606cbb7cc400d9682d978c4f14f92d to handle these errors more neatly.

PR #2095, opened 2026-09-23
perf: relax GC pacing for one-shot commands

srenatus:
lint, test and fix allocate in bulk and then exit, so collecting at the default pace is largely wasted work.

    regal lint bundle   640ms -> 543ms, peak RSS 340MB -> 600MB

The language server keeps the default: a single-file lint gains 4.7%, and it would hold the larger heap for the life of the session. An explicit GOGC still wins.


anderseknert (approved):
Thanks! I've been tempted to turn GC off entirely for the lint command in the past, but back then we spent _seconds_ collecting garbage. Much good has happened since then ☺️ Back then I was concerned about whether it'd be an issue in CI runners and linting in other constrained environments. This seems like a reasonable compromise 👍

